PR #33, opened 2026-07-08
Change default port to 1777; drop legacy bookmarks localStorage migration

vasu2001:
## Port change
Changes the default server port from 8765 to 1777 (fallback range now 1777–1787). Updated in `fused_render/app.py`, `fused_render/cli.py`, `fused_render/_branch.py` (per-branch hashed port band shifts to 1788–2787), the vite dev proxy, and all docs/skills references (README, SPEC, ARCHITECTURE, EXPORT, skills, template "open in fused-render" hints).

## Remove legacy bookmarks localStorage migration
Every pre-D75 install has long since had its bookmarks migrated to `~/.fused-render/bookmarks.json` on first boot after that release. No interim version is being supported going forward, so the one-time `localStorage["fused.bookmarks"]` import/merge code is now dead weight — removed outright rather than left dormant:
- `frontend/src/lib/bookmarks.ts`: dropped `LS_KEY`, `MIGRATED_KEY`, `legacyMigrated`/`markLegacyMigrated`, `readLegacy`, `mergeLegacy`; `hydrateBookmarks()` is now a plain GET-into-cache.
- `fused_render/shell/bookmarks.py`: docstring/comments no longer describe the `exists` flag as gating an import (the flag itself is unchanged — it still distinguishes absent/corrupt from a valid empty file).
- Documented as `D104` in `DECISIONS.md`; `SPEC.md`/`ARCHITECTURE.md` updated to drop the stale description.

Tests: `tests/test_shell_bookmarks.py`, `tests/test_branch.py`, `tests/test_branch_runtime.py` all pass (67 total). 6 pre-existing failures elsewhere in the suite (`test_deploy.py`, `test_duckdb_reader.py`, `test_shell_mounts.py`, `test_templates_api.py`) reproduce identically on `origin/main` and are unrelated to this branch.


vasu2001:
there is a problem with this change, right now all bookmarks are stored in browser local storage that is separate based on the domain. If we change the port then everyone will loose their bookmarks

PR #40, opened 2026-07-08
Per-branch artifact & state isolation for parallel branch testing

iamsdas:
## Summary

Temporary, easy-to-remove scaffolding so **multiple git branches can be built, installed, and run in parallel on one machine** without colliding — while leaving `main` byte-for-byte unchanged.

**The whole feature is one idea:** everything keys off a single short **branch-ref**. Set the ref and every surface (port, state dirs, app identity) shifts to a namespace for that ref; leave it unset and you get today's exact behavior.

**Isolation is opt-in.** Being on a feature branch does *nothing* on its own — there is no automatic git detection. You supply the ref explicitly via `FUSED_RENDER_BRANCH=<ref>` — at **build time** to bake it into a DMG/wheel, or at **runtime** to isolate a source/editable run.

`main`/`master`/`head` and the empty string all resolve to `""` = **baseline** (no isolation).

---

## How the ref flows

```mermaid
flowchart TD
    ENV["FUSED_RENDER_BRANCH env var"] --> R
    BAKED["_baked_branch._BAKED_REF<br/>(baked into a build only when env is set)"] --> R
    R["fused_render/_branch.py<br/>resolve → sanitize (≤12 chars, main/master/head → '')"]
    R --> PORT["branch_port() → 8776–9775<br/>(baseline: 8765)"]
    R --> DIRS["state dirs nested under &lt;ref&gt;/<br/>(baseline: unnested)"]
    R --> APP["DMG/app: FusedRender-&lt;ref&gt;<br/>io.fused.render.&lt;ref&gt;"]
    R -.->|ref = ''| BASE["BASELINE = identical to main"]
```

Resolution order is **env → baked → ""**. No git.

---

## What a given ref changes

For ref `foo` (baseline shown for contrast):

| Surface | Baseline (`main` / no ref) | Ref `foo` | Set by |
|---|---|---|---|
| Server port | `8765` (scan 8765–8775) | `branch_port("foo")` e.g. `9734` (scan +10) | runtime, from ref |
| Shell home (`templates/`, `bookmarks.json`, `prefs.json`) | `~/.fused-render/` | `~/.fused-render/foo/` | runtime, via `home_dir()` |
| App Support dir | `…/fused-render/` | `…/fused-render/foo/` | runtime, from ref |
| macOS app name | `FusedRender.app` | `FusedRender-foo.app` | `setup_py2app.py` at build |
| Bundle identifier | `io.fused.render` | `io.fused.render.foo` | `setup_py2app.py` at build |

All shell state nests together: the per-branch subdir is applied once in `shell/storage.py`'s `home_dir()`, so templates, bookmarks, and prefs isolate as a unit (two branches never share a `bookmarks.json`).

**Not covered:** the Claude Code plugin/skill identity (`marketplace.json`, `plugin.json`, `SKILL.md`) is intentionally left alone — a git-URL plugin install of a branch still resolves to the baseline `fused-render` name. This change isolates the **app/runtime**, not the plugin registration.

---

## Usage

**Run a branch from a source checkout, isolated:**
```bash
FUSED_RENDER_BRANCH=foo fused-render serve
# binds foo's port, uses ~/.fused-render/foo/, logs "(branch foo)"
```

**Build a self-contained DMG or wheel for a branch:**
```bash
FUSED_RENDER_BRANCH=foo bash scripts/build_dmg.sh   # → FusedRender-foo.app / io.fused.render.foo
FUSED_RENDER_BRANCH=foo pip wheel .                  # → wheel with the ref baked in
```
The ref is **baked into the artifact**, so the installed app/wheel self-isolates (port, dirs, identity) with **no runtime configuration** — nothing to set, no flags, no config file. Just launch it. With the env var unset at build, you get a plain baseline `FusedRender.app` identical to `main`.

---

## Baseline invariant (`main` is untouched)

When the ref is `""` — on `main`, or with the env var unset in a packaged build — **every** name/path/port equals today's exact value:
- port stays `8765`/`8775`, dirs stay `~/.fused-render` / `…/fused-render` (no trailing-slash trap),
- app stays `FusedRender` / `io.fused.render`,
- a baseline build writes **no** `_baked_branch.py` at all.

This is enforced by tests, not just intent.

---

## Removing this later (it's temporary)

Delete `fused_render/_branch.py` and the `tests/test_branch*.py` + `tests/test_hatch_build.py`; revert the guarded edits in `shell/storage.py` (`home_dir()`), `app.py`, `cli.py`, `setup_py2app.py`, `build_dmg.sh`, and `hatch_build.py`; drop the `.gitignore` line. `server.py` needs no revert — it just uses `home_dir()`. No data migration — `main` users never wrote to `<ref>` subdirs.

> Based on `3c512eb`; verified to merge cleanly onto current `main` (which has since advanced with #32–#37) via `git merge-tree` + a semantic check that the registry reader still uses the nested `USER_TEMPLATES_DIR`/`USER_REGISTRY`.

---

## Test Plan

- [x] `tests/test_branch.py` — ref resolution (env → baked → baseline), sanitize (incl. `main`/`master`/`head` → baseline, truncation), deterministic `sha1`-derived port in `[8776,9775]`, cached no-arg path, that git is never consulted, and that a baked `main` resolves to baseline.
- [x] `tests/test_branch_runtime.py` — with ref `""`: `home_dir()` un-nested, `USER_TEMPLATES_DIR = home/templates`, `APP_SUPPORT_DIR` un-nested, ports `8765`/`8775`; with ref `foo`: `home_dir()` and App Support both nest under `foo/`, `USER_TEMPLATES_DIR = home/foo/templates`, `branch_port("foo")`, `MAX_PORT = DEFAULT_PORT + 10`, `cli.DEFAULT_PORT == app.DEFAULT_PORT`.
- [x] `tests/test_hatch_build.py` — bake writes + registers the artifact when a ref is set; a baseline build removes a stale `_baked_branch.py` and registers nothing; baseline is a no-op when none exists.
- [x] Full suite green: `python -m pytest --ignore=tests/test_server_export.py` → 69 passed, 2 skipped. (`test_server_export.py` fails at collection on a pre-existing, unrelated `httpx2`-missing env issue — identical on `main`.)
- [ ] Manual: build two branches' DMGs and confirm they install as distinct apps; run two branches' servers simultaneously and confirm distinct ports + isolated `~/.fused-render/<ref>/`.
- [ ] Not run: a full `bash scripts/build_dmg.sh` (needs py2app/dmgbuild) — the DMG app-name / bundle-id stamping verified by inspection + the `_branch` helper.


iamsdas:
Addressed the Bugbot findings in `6f72400` (on top of the opt-in refactor in `a66515b`):

- **Worktree git root not found** (`_git_ref`) — **obsolete**. The opt-in refactor removed git auto-detection entirely; `_git_ref` no longer exists. Isolation now engages only via `FUSED_RENDER_BRANCH` (build-time or runtime) or a baked ref, so there's no worktree `.git`-file walk to get wrong.
- **Stale baked ref after baseline** (medium) — **fixed**. A baseline build now *removes* any `_baked_branch.py` left by an earlier branch build, instead of returning early and leaving it on disk. Extracted into a pure `_write_baked_ref()` with unit tests (`tests/test_hatch_build.py`) covering write-on-ref, remove-on-baseline, and noop-when-absent.
- **Pip cache stale branch bake** (high) — **fixed**. `build_dmg.sh` reuses the build venv, so pip could treat an unchanged version as already-satisfied (or reuse a cached wheel) and ship a stale baked ref across ref switches. Added a forced `--force-reinstall --no-deps --no-cache-dir` reinstall of fused-render so the ref is re-baked every run.

PR #54, opened 2026-07-08
feat(bookmarks): persist to ~/.fused-render/bookmarks.json (D75)

vasu2001:
## What

Migrate the bookmark store off browser **localStorage** onto a **server-side file** at `~/.fused-render/bookmarks.json`. D21/SB-5 chose localStorage but flagged this move as a "trivial export" for later — this is that move. Bookmarks are now per-machine (portable) rather than per-browser-profile.

## Server — new `fused_render/shell/` package

The seam requested: shell-specific state backends kept **out of** `server.py`'s fs/HTML-render internals.

- **`shell/storage.py`** — `~/.fused-render` home dir (`home_dir()`, `FUSED_RENDER_HOME` env override for tests) + atomic `read_json`/`write_json` (temp file + `os.replace`). Foundation the next shell-state resource reuses.
- **`shell/bookmarks.py`** — `APIRouter`: `GET /api/bookmarks` → `{exists, bookmarks}`, `PUT` (whole-tree, atomic, **last-write-wins**, `X-Fused` D3 guard). Guard duplicated locally so `shell` never imports `server` (no cycle).
- `create_app` gains one `include_router` line.

## Frontend — `lib/bookmarks.ts`

- Reads stay **synchronous** off an in-memory cache (React renders can't await), hydrated once at boot by `hydrateBookmarks()`.
- Mutations become **async**: clone → `await` PUT → advance cache. **No optimistic update, no rollback** — localhost PUT is sub-ms, and advancing the cache only after a successful write means the UI never shows unpersisted state (surfaces a failed write instead of a fake success).
- **One-time import**: `exists` is false only until the file is first written; on `!exists` + non-empty legacy `localStorage["fused.bookmarks"]`, import once. Legacy key left dormant as a fallback, never re-imported (survives the user deleting all bookmarks).
- Event bus (`notifyBookmarksChanged`/`useBookmarksVersion`) unchanged. `armedBookmark` stays in sessionStorage (D38).

## Rejected (all owner-confirmed)

- **TanStack Query** — 3 tiny single-writer last-write-wins resources don't need caching/dedup/pagination, and its refetch-on-focus/reconnect defaults fight last-write-wins. Hand-rolled cache mirrors the existing `App.tsx:useStat` house pattern, zero dep.
- **Optimistic updates** — nothing to hide at localhost latency; await-then-reflect is simpler and more correct.
- **Auto-import + clear localStorage** / **fresh start** — kept dormant fallback / would drop existing bookmarks.

## Docs

D75 supersedes D21/SB-5; SPEC SB-5 + ARCHITECTURE §6.5 + dir-tree updated.

## Tests

`tests/test_shell_bookmarks.py` — absent→`exists:false`, PUT/GET roundtrip + on-disk file, empty-list-still-exists (import gate), corrupt→not-exists, no-header→403, last-write-wins. `FUSED_RENDER_HOME` redirected to tmp so no test touches the real home.

- `tsc --noEmit && vite build` — clean
- `pytest` — **86 passed, 2 skipped**

---

> [!NOTE]
> **Medium Risk**
> Changes bookmark persistence and multi-tab write semantics (last-write-wins); mitigated by local-only binding and serialized client writes, but failed PUTs or stale tabs can still surprise users until poll converges.
> 
> **Overview**
> Bookmarks move from **browser localStorage** to **`~/.fused-render/bookmarks.json`**, served by new **`GET`/`PUT /api/bookmarks`** (`{exists, bookmarks}` on read; whole-tree PUT with **`X-Fused`** guard). Server logic lives in **`fused_render/shell/`** (`storage.py` atomic JSON + `bookmarks.py` router); **`server.py`** only includes the router and points user templates at **`home_dir()/templates`** (D76).
> 
> The shell keeps **sync reads** off an in-memory cache filled by **`hydrateBookmarks()`**; all writes **`await` PUT** before updating the cache (no optimistic UI). A **serial `enqueue` chain** serializes hydration, mutations, and polling so boot races and overlapping writes cannot clobber the file. **One-time import** from legacy `localStorage` when `exists` is false; a **30s poll** refreshes from the server for cross-tab convergence (eventual, last-write-wins). **Breadcrumb** and **Sidebar** handlers are **`async`** around bookmark mutations.
> 
> Docs record **D75–D77**; **`tests/test_shell_bookmarks.py`** covers the API with **`FUSED_RENDER_HOME`** redirected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0ebc28becf9c1bf46d6d83d9e5959a061714b28. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vasu2001:
### Bugbot follow-up (commit a1d455e)

**① Mutations during hydration lose data (High) — fixed.**
All cache access now runs through one serial promise chain. `hydrateBookmarks()` is enqueued first at boot and every mutation after it, so a mutation can no longer read the still-empty cache before the initial GET returns, and two overlapping mutations can't both clone the same snapshot and clobber each other (the clone-read + commit are now one atomic step via a `mutate(transform)` helper). `hydrateBookmarks()` is also idempotent now.

**② Other tabs keep stale cache (Medium) — working as intended.**
This is the accepted design per DECISIONS **D75**: bookmarks are **last-write-wins with no cross-tab live sync** (the owner explicitly chose this over broadcasting changes). A per-tab in-memory cache is the deliberate trade-off; a multi-window user gets last-write-wins, same contract the endpoint advertises. Not fixing — would contradict the recorded decision.

vasu2001:
Update on **② Other tabs keep stale cache**: per owner, now addressed (not won't-fix) — a 30 s poll (`refreshBookmarks()`, commit b0ebc28, **D77**) re-reads the server through the same serial queue and re-renders on a real diff. Cross-tab convergence is eventual (≤30 s); simultaneous writes remain last-write-wins. No WebSocket sync.

PR #57, opened 2026-07-08
Search hidden entries on dot-prefixed queries

AkshilVT:
## What

Explorer search couldn't find hidden entries — the walk prunes dot-files/dirs unconditionally, so an explicit query like `.fused-render` from `~` returned "No matches" while the same entry is visible in the plain listing.

A dot-leading query (or one containing a `/.` segment) signals explicit intent: the walk is now fetched with a new `hidden=1` param that includes dot-files and descends dot-dirs. Everything else is unchanged — normal queries still prune hidden entries, `node_modules`/`__pycache__`/`venv` are always pruned, and the 20,000-entry cap + `truncated` flag hold in both modes.

Client side: the walk cache is keyed by the hidden flag (tagged alongside path/refresh in the walk state), so flipping between dot and non-dot queries refetches the right dataset; focus-prefetch stays non-hidden.

## Testing

- `tests/test_server_walk.py`: hidden=1 returns dot entries and descends dot-dirs, still prunes ignored dirs; default behavior unchanged (7 tests pass).
- tsc + vite build clean.
- Verified live: `.fused-render` query from `/view/Users/akshilthumar` returns the hidden dir.

---

> [!NOTE]
> **Low Risk**
> Read-only FS listing API plus explorer UI caching; default walk behavior is unchanged and ignored dirs stay pruned even with `hidden=1`.
> 
> **Overview**
> Explorer search could not find dot-files or contents of dot-directories because the recursive walk always pruned them. **Dot-prefixed queries** (leading `.` or a `/.` path segment) now request a separate walk with **`hidden=1`**, which includes those entries while still skipping `node_modules` / `__pycache__` / `venv` and honoring the existing entry cap.
> 
> On the client, **`walkDir`** passes `hidden=1` when needed, and the walk cache is tagged with **`forHidden`** alongside path and refresh so flipping between normal and dot queries refetches the correct dataset. Hidden intent is derived from the **immediate** search input (not the deferred value) so the right walk starts on the first keystroke; focus prefetch stays non-hidden.
> 
> Tests cover `hidden=1` behavior and unchanged default pruning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afba2fd94960558166e34af29b47d533b8331e03. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fused-render-shell",
-  "version": "0.0.0",
+  "version": "file:../../../private/tmp/fr-hidden-pr/frontend",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -19,7 +19,7 @@
         "vite": "^6.0.7"
       }
     },
-    "node_modules/@babel/code-frame": {
+    "../../../private/tmp/fr-hidden-pr/frontend/node_modules/@babel/code-frame": {
       "version": "7.29.7",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
       "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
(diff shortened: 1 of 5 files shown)


AkshilVT:
bugbot run

AkshilVT:
bugbot run

AkshilVT:
bugbot run

PR #62, opened 2026-07-08
feat(templates): stage core templates into ~/.fused-render/.core-templates

vasu2001:
## What

Built-in templates ship inside the package but were read directly from the read-only app bundle. This stages them into `~/.fused-render/.core-templates/` on startup and has the server + executor read every template, registry, and helper from that copy.

## Why

Moves core templates out of the immutable bundle into a writable, per-user location — the foundation for editing/refreshing them independently of the packaged app, while keeping each release's set pristine.

## How

- **New `fused_render/core_templates.py`** — `ensure_core_templates()` version-gates the copy with a `.version` marker holding `__version__`. Fresh install or upgrade → full wipe + re-copy (reset-on-release).
- **Atomic swap** — copy is built in a private `.staging.<pid>` sibling (marker included) and swapped in with `os.replace`. A request handler never sees a half-written tree; concurrent starts don't crash (last swap wins, loser discards its copy — no locking, single local user per D3); an interrupted copy leaves the old dir intact and retries next launch.
- **`server.py` / `executor.py`** — `TEMPLATES_DIR` / `_TEMPLATES_DIR` (+ in-process helper allowlist) now resolve to the staged dir.
- **Dev bypass** — `FUSED_RENDER_CORE_TEMPLATES` reads the named dir verbatim (abspath'd + stripped); `scripts/dev.sh` sets it to the in-repo templates so edits show up live without a version bump.
- **`tests/conftest.py`** — redirects `FUSED_RENDER_HOME` to a throwaway tmp dir (atexit cleanup) so import-time staging never touches the real home.

The user override channel at `~/.fused-render/templates/` is untouched and still shadows a core template of the same name.

## Adversarial review addressed

Ran a skeptical review; fixed: concurrent-start race (atomic swap), non-UTF-8 marker crash (`except (OSError, ValueError)`), bypass env validation (strip + abspath), conftest eager-mkdtemp leak. Left import-time crash-on-failure as intended (no bundle fallback — loud failure preferred).

## Test

`pytest` — template/executor suites green (66 passed). Pre-existing failures unrelated (React shell not built in a fresh checkout). Verified behaviorally: fresh stage leaves no orphan, non-UTF-8 marker recovers, whitespace/relative override handled, simulated lost swap race does not crash.

---

> [!NOTE]
> **Medium Risk**
> Changes where all built-in templates, registry, and in-process helpers resolve (import-time filesystem staging); failure to stage is designed to crash loudly rather than fall back to the bundle.
> 
> **Overview**
> Built-in templates are no longer read from the packaged app bundle. **`ensure_core_templates()`** copies them into `~/.fused-render/.core-templates/` on startup (version-gated via a `.version` marker so upgrades get a fresh copy), and **`server.py`** / **`executor.py`** resolve `TEMPLATES_DIR` and in-process helper allowlists from that staged tree.
> 
> Staging uses a **`.staging.<pid>`** sibling directory and **`os.replace`** so readers never see a partial copy; concurrent starts tolerate a swap race without locking. **`FUSED_RENDER_CORE_TEMPLATES`** skips staging for dev (abspath + strip); **`scripts/dev.sh`** points it at in-repo `fused_render/templates`. **`tests/conftest.py`** sets **`FUSED_RENDER_HOME`** to a temp dir before imports so tests do not touch the real home. User overrides under `~/.fused-render/templates/` are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31989d36b4b0c3218c1ec312c83e31dec9c3a4e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vasu2001:
Addressed both Bugbot findings.

**#2 — Swap failure leaves no templates (Medium): fixed** in the latest commit. `os.replace` `OSError` is now only swallowed as a lost race when `core_dir` exists (another instance already swapped this version in → a complete tree is live). A genuine failure (core_dir already wiped, replace failed for perms/lock) re-raises so it surfaces loudly (crash-on-failure by design) and leaves the complete `.staging.<pid>` copy for inspection instead of deleting it. Verified both branches: lost-race → no crash; genuine failure → raises.

**#1 — Home not created before staging (High): false positive, no change.** `shutil.copytree(src, dst)` calls `os.makedirs(dst)`, which creates the full parent chain — including `~/.fused-render`. Verified against a nonexistent home:

```
FUSED_RENDER_HOME=/tmp/.../does-not-exist-yet/.fused-render
home existed before: False
staged ok: True
```

The staging dir sits directly under `home_dir()`, so copytree creates the home dir en route. No prior `mkdir` needed.

PR #72, opened 2026-07-09
feat(windows): Explorer "Open with" opener and registration scripts

amanbagrecha:
## Setup (new user)

1. Install the wheel: `pip install fused_render-*.whl` (or `uv pip install fused_render-*.whl`).
2. Register the "Open with" integration once: `fused-render-open --register` (add `--port N` to pin a port).
3. To remove it later: `fused-render-open --unregister`.

## What this does

Adds Windows Explorer "Open with" support, the counterpart to the existing macOS Finder flow: right-click a file and open it in fused-render at its `/view/` URL. A new `fused-render-open` launcher (a `gui-scripts` entry point, so it ships in the wheel and shows as "fused-render" rather than "Python" in the Open With dialog) reuses a running server or starts one detached, then opens the browser. `--register`/`--unregister` write the per-user HKCU associations themselves (no admin, no external script needed): one ProgID per extension so Explorer's Type column names the format, `OpenWithProgids` entries for every extension in `templates/registry.json`, an all-files context-menu verb, and an app-icon. The `scripts/windows/*.ps1` files are thin dev wrappers around the same `--register`/`--unregister`. No changes to templates or frontend.


amanbagrecha:
Both Bugbot findings addressed in 2288f73:

- **Misses CLI server / duplicate spawn**: `find_running_server` now scans the whole default port range, not just the portfile hint, and identifies servers by probing `/api/config` — including a native-vs-mirrored check (a WSL fused-render mirrored onto localhost reports a POSIX home and is skipped rather than reused).
- **Silent failures under the windowless launcher**: `main()` now routes everything through one `except Exception` that logs and raises a message box, and a requested port occupied by a non-fused-render app fails with an explicit message instead of a bind-timeout.

Also in that commit: per-extension ProgIDs (`FusedRender.csv` → type name "CSV File (fused-render)") so Explorer's Type column keeps naming the format; the old shared `FusedRender.file` ProgID is migrated away on re-register.

amanbagrecha:
Latest two Bugbot findings addressed in 5672db5:

- **Timeout leaves detached server**: on a failed boot the opener now kills the spawned child and removes the pid/port files, so nothing is left squatting the port and a retry starts clean.
- **Startup race misclassifies fused port**: the portfile is now written at spawn time (intent) rather than after readiness, and a bound-but-silent port gets the same boot grace as our own spawn (`_settle`). A racing opener therefore waits for the peer's server instead of erroring with "in use by another application" or spawning a duplicate. The grace only applies to the requested/filed port, so a foreign or WSL-mirrored server elsewhere in the scan range is classified instantly and never stalls the scan.

Verified live: instant reuse of a running server, and a synthetic bound-but-silent socket times out through the grace path and fails with the explicit message.

amanbagrecha:
Fixed in 9bd8850: the `main()` failure handler now `raise SystemExit(1)` after the message box, so a failed `--register`/`--unregister` propagates a non-zero exit and the PowerShell wrappers' `exit $LASTEXITCODE` no longer reports failure as success. The success path still exits 0.

PR #91, opened 2026-07-10
feat(templates): parquet structure viewer (layout + metadata)

iamsdas:
## Summary

Adds a new **`structure`** template for `.parquet` files that surfaces the file's *internal structure* — the row-group / column-chunk level info you normally only get from a CLI — inspired by the [hyparquet demo](https://hyparam.github.io/demos/hyparquet/). It shows up as the **Structure** mode alongside `duckdb`/`table` in the parquet mode switcher.

- **Layout tab** — nested byte-range boxes: the `PAR1` header, each row group with its column chunks (each column → its `Data` page), and the `PAR1` footer, every box labelled `start / bytes / end`. A proportional colour track under each row group shows how its bytes split across columns.
- **Metadata tab** — a file summary, the per-column schema (physical + logical type, max def/rep levels), and one expandable panel per row group listing each column chunk's codec, encodings, value count, compressed/uncompressed size, offset, `min … max` and null count.
- **Grid tab** — a whole-file storage profile: one bar per row group, its length ∝ the row group's compressed size relative to the largest, split into per-column segments sized by bytes. A colour legend names each column with its file-wide size + share, each row carries its size / row-count, and segments have hover tooltips — so "which column/row-group eats the space?" is answerable at a glance.
- `reader.py` is built on **pyarrow** (already a dependency, same as the `table` reader) and parses **only the footer** — no data pages are read, so it stays cheap even on a multi-GB file.

## Visuals

image

Verified live against a 3-row-group parquet:

- **Layout** — `PAR1 header (0→4)`, `RowGroup 0` (100 rows, 1,434 B) with contiguous column chunks `id 4→603`, `city 603→740`, `score 740→1,380`, `active 1,380→1,438`, proportional track, then the footer ending exactly at EOF.
- **Metadata** — File (2.6, `parquet-cpp-arrow 24.0.0`, SNAPPY, 3.6 KB→5.5 KB), Schema (`id INT64`, `city BYTE_ARRAY/String`, …), and expandable RowGroup panels with codec/encodings/stats.

## Usage

Open any `.parquet` file in fused-render and pick the **Structure** mode (the striped-document icon). Toggle **Grid** / **Layout** / **Metadata** in the top-right; the choice is stored in the `view` param so it's deep-linkable (`?_mode=structure&view=metadata`). Reads are footer-only, so it opens instantly regardless of file size.

## Test Plan

- [x] `tests/test_structure_reader.py` — 7 new tests: file summary, schema physical/logical types, row-group + chunk byte-range self-consistency (`end == start + bytes`, first chunk starts at 4), column statistics (min/max/nulls), layout header/footer offsets (footer ends at EOF), row-group regions, and JSON-serializability.
- [x] `tests/test_templates.py` — updated the parquet routing assertion to include `structure` (now `duckdb, table, structure, h3, claude, annotate, history`).
- [x] Suite: **183 passed, 3 skipped** (7 TestClient files excluded — pre-existing missing-`httpx` env issue, unrelated).
- [x] Manual (cmux browser, a 250-row / 3-row-group parquet): Layout mode renders header/row-group/column/footer boxes with correct contiguous offsets and the proportional track; Metadata mode renders the file summary, schema table, and expandable per-row-group chunk detail.

---

### Naming & Metadata-tab cleanup

- Mode named **`structure`** (shown as "Structure" in the switcher) — the template holds both a **Layout** and a **Metadata** tab, so an umbrella name reads better than either tab's own name.
- The **Metadata** tab's File section no longer repeats size / rows / row-groups / columns / compression (the always-visible top summary bar already carries them); it now lists only path, format version, creator, compressed/uncompressed totals, and footer size.

---

### Update: Grid tab

Added a third **Grid** tab (Grid · Layout · Metadata; Grid is the default) — a whole-file storage profile modelled on the hyparquet demo's Grid but made more legible: one bar per row group scaled to the largest group, split into per-column segments, with a named colour legend (size + %), per-row-group size/row-count labels, hover tooltips, and a plain-language caption. Pure frontend — reuses the reader's existing `row_groups` data. Verified live on a 17-row-group file where the `text` column (64%) visibly dominates and the final short row group scales correctly.

---

### Update: Grid is the default; Layout strip removed

Grid is now the first tab and the default view (it's the clearest answer to "where do the bytes go?"). The per-row-group proportional strip was removed from the Layout tab, since the Grid view is now the single home for the "bytes by column" visual — Layout stays focused on exact byte offsets.


iamsdas:
Addressed both Cursor Bugbot findings in 628b445:
- **Racing loads** — added a `loadSeq` sequence guard so a slow read for an earlier `_file` that resolves late is discarded instead of overwriting the newer file's data/summary/view.
- **Stale tab switch** — `load()` now clears `data` and the summary up front, so switching Layout/Metadata during a load or after a failed read no longer repaints the previous file's content (`renderTab()` already early-returns when `data` is null).

iamsdas:
Addressed the latest Bugbot findings in d11c4fa:

- **Table removal breaks inventory test** — valid. Removing the `table` template had wider fallout than the test alone, all fixed:
  - `executor.py`: replaced the deleted `table/reader.py` with `structure/reader.py` in `INPROCESS_HELPERS`, so the pyarrow parquet reader runs **in-process** (no TCC prompt / subprocess isolation) exactly as `table`'s reader did.
  - `test_templates_api.py`: the core-template inventory assertion now targets `structure` (the template bound to `.parquet`).
  - `test_shell_prefs.py`: `.parquet`'s first mode is now `duckdb`.
- **Grid legend omits later columns** — valid (defensive). The Grid legend and file-wide totals now enumerate columns from `data.schema` (authoritative, stable order) instead of `row_groups[0]`, so every column is covered even if a row group holds fewer chunks.
- **Tab switch shows stale data** / **Racing loads overwrite data** — already fixed earlier in 628b445 (a `loadSeq` guard + clearing `data`/summary at the start of every load); those two were reported against the older commit 7a37d7a.

Note: `test_templates_api.py` and `test_shell_prefs.py` use `starlette.testclient`, which this local env can't import (missing httpx); I verified the assertion changes against the inventory/registry builders directly instead.

PR #96, opened 2026-07-10
feat(templates): comment deep-linking from history to annotate (D100/D101)

AkshilVT:
## What

History-view comment rows now deep-link into annotate mode, mirroring the claude `session_id` resume contract: the URL carries only `_mode=annotate&comment=<id>`.

## How

- **annotate template** — new boot-time `focusFromParam()`: resolves the id from the live URL `comments` param; when absent, a one-shot **full-state hydration** imports every non-tombstoned entry from the `<file>.json` comments log (server stamps stripped, live entries win by id), saves once, then focuses the pin/card (switching to the comment's own view first when it differs).
- **explicit delete tombstones** — the annotate delete button now also fires a new `annotate.py` `delete` action stamping `deleted_at` (seconds) on the log entry. Absence from a `record` array still never deletes (each URL carries only its own review subset — D98 records why absence-based deletion is impossible). Re-recording an id clears the tombstone.
- **history template** — comment rows with an id are clickable like session rows; tombstoned entries render dimmed + struck-through with a " · deleted" tooltip and are **inert** (deleted stays deleted — the resolver also refuses tombstoned ids on hand-crafted URLs). Continue banner restyled: resume arrow instead of a play triangle, human meta (mode + relative time) instead of the raw query string.
- **SPEC §17/§24 (HV-7, HV-8) + DECISIONS D97/D98** updated; supersedes D96's read-only-comments sub-call (owner reversed 2026-07-09 call on 2026-07-10).

## Containment

All logic stays in the two templates + `annotate.py`. No shell, server, or registry changes.

## Tested

- Playwright against a live server: deep link from timeline (live id → 4 cards hydrated + clicked one active), deleted row inert, hand-crafted tombstoned-id URL → no import/focus/save, cross-view focus, sidecar tombstone survives.
- `pytest`: 363 passed; 4 failures pre-exist on origin/main (env-dependent deploy/templates_api tests).

---

> [!NOTE]
> **Medium Risk**
> Changes sidecar comment semantics and deep-link hydration (merge live + log, permanent tombstones); scope is limited to templates and `annotate.py`, with new tests, but wrong hydration or tombstone handling could affect shared `.json` sidecars.
> 
> **Overview**
> **History → annotate deep links:** comment rows with an `id` now open the target with `_mode=annotate&comment=<id>` (same id-only resume pattern as claude `session_id`). Tombstoned comments (`deleted_at`) render struck-through, show a deleted tooltip, and are not clickable.
> 
> **Annotate boot focus:** on load, `focusFromParam()` reads `comment`; if the id is missing from the live URL `comments` JSON, it **hydrates** all non-tombstoned entries from the sidecar log (strips server stamps, live wins by id), saves once, then focuses—switching views first when needed. `switchView` now returns success so a refused switch can retry focus.
> 
> **Explicit deletes in the sidecar log:** deleting a comment removes it from the URL and passes `deleted_ids` on the **same** `record` call as the upsert so `annotate.py` can stamp permanent `deleted_at` in one atomic write; re-recording an id cannot clear a tombstone. Absence from `comments` alone still never deletes.
> 
> Docs (**SPEC** §17/§24, **DECISIONS** D100/D101) and **annotate** unit tests cover tombstone and merge behavior. Minor history UI: resume icon and friendlier “continue” banner meta.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9c9ffdb985160dafdf5cd24370919e815da367f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AkshilVT:
bugbot run

AkshilVT:
bugbot run

AkshilVT:
bugbot run

AkshilVT:
bugbot run

AkshilVT:
bugbot run

PR #110, opened 2026-07-13
feat(pdf_studio): flat PDF library instead of projects, default to View mode

amanbagrecha:
## What

Reworks the built-in **PDF Studio** template to drop the "project" concept in favor of a flat, persisted list of PDFs, and defaults the app to **View** mode.

### Before → After
- **Projects removed.** No more topbar project dropdown (New/Rename/Duplicate/Delete), no auto-seeded demo project.
- **Flat library.** Documents now live in a single `~/.fused-render/data/pdf_studio/library.json` (`{"paths": [...]}`) of absolute paths. Entries are *references* — files are never copied, and "Remove from list" keeps the file on disk.
- **Add via the file browser.** The sidebar **＋** and **File ▸ Add PDF… (Ctrl+O)** both open the existing browser modal, which lets you paste a full path or navigate the filesystem. "Import from URL" is kept as **Add from URL…** (downloads into `data/pdf_studio/downloads`, then adds to the list).
- **Defaults to View mode** on load; Edit is still one click away (or `?mode=edit`).

### Backend (`pdf.py`)
- New actions: `list_library`, `add_to_library`, `remove_from_library`.
- `import_url` no longer needs a project; `merge` / `rename_doc` drop the `project` arg.
- `_hist_dir` always uses the snapshots dir (undo/redo unchanged otherwise).
- Removed the demo generator, project/meta helpers, and now-unused `import time`.

## Why
The project layer added ceremony without much payoff for a local single-user PDF tool. A flat "add a PDF, it's remembered" model is simpler and matches how people actually reach for files.

## Notes for reviewers
- Net change is mostly deletions (~533 removed vs ~132 added).
- Working-copy / undo-redo / save / merge / split / compress / text-edit machinery is untouched — it was always keyed by file path, not project.
- Verified end-to-end: drove the `pdf.py` dispatcher directly (add/dedupe/list/open/rotate+save/merge/rename/remove) and exercised the same flow through a running server's engine via `fused.runPython`, plus the file browser's `listdir`.

---

> [!NOTE]
> **Medium Risk**
> Large API and persistence change for pdf_studio (library index vs projects) with behavior shifts (no on-disk delete, no demo project), though core edit/save/undo paths stay path-keyed.
> 
> **Overview**
> **PDF Studio** drops the project model for a single persisted library and simplifies the UI around “add a PDF, remember the path.”
> 
> The backend replaces project folders, `meta.json`, and demo seeding with `library.json` (`paths` of absolute references), `list_library` / `add_to_library` / `remove_from_library`, URL downloads under `data/pdf_studio/downloads`, and undo history always under `cache/snapshots` (no per-project `.history`). `merge`, `rename_doc`, and `import_url` no longer take a `project` argument; default action is `list_library`. Disk **delete** is gone — removal only unlinks from the library and clears work/undo for that path.
> 
> The template removes the project dropdown and standalone flows, drives the sidebar from `list_library`, renames **Add PDF** / **Add from URL**, uses **Remove from list** (file kept), auto-adds merged outputs to the library, and boots by adding deep-linked PDFs to the library. **View** mode is the default (Edit via toggle or `?mode=edit`).
> 
> Read-only tests now target `LIBRARY` instead of `PROJECTS` and replace `delete_doc` RO coverage with `remove_from_library` (no file delete).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 003700ba20849a11226e69ff557df26b05e63220. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/fused_render/templates/pdf_studio/pdf.py
+++ b/fused_render/templates/pdf_studio/pdf.py
@@ -10,28 +10,22 @@ ops, PyMuPDF handles text extraction/editing and rasterization. Every action
 returns JSON so an AI agent can drive the whole app headlessly exactly as the
 UI does.
 
-The source of truth is the .pdf files on disk, wherever they live — projects
-link imported files instead of copying them. Edits never touch the original
-directly: each open doc gets a working copy under .work/ that mutations (and
-undo/redo snapshots) apply to; an explicit save writes the working copy back
-over the original. Each call is a fresh process, so no in-memory state
-survives.
+The source of truth is the .pdf files on disk, wherever they live. A flat
+library (a single JSON file of absolute paths) just remembers which files the
+user added — it never copies them. Edits never touch the original directly:
+each open doc gets a working copy under .work/ that mutations (and undo/redo
+snapshots) apply to; an explicit save writes the working copy back over the
+original. Each call is a fresh process, so no in-memory state survives.
 
 Actions
   health                                   -> {ok, pymupdf, pikepdf}
-  list_projects                            -> {projects:[...], dir}
-  new_project(title)                       -> {slug}
-  rename_project(slug,title)               -> {slug}
-  duplicate_project(slug,title)            -> {slug}
-  delete_project(slug)                     -> {ok}
-  open_project(slug)                       -> {slug, title, docs:[...]}
+  list_library                             -> {docs:[...]}
+  add_to_library(src)                      -> {name, path}  (references src in place, no copy)
+  remove_from_library(doc)                 -> {ok}  (drops the reference; file on disk is kept)
+  import_url(url,name)                     -> {name, path}
   open_doc(doc)                            -> docinfo + {work, dirty, has_text, undo_depth, redo_depth}
-  which_project(doc)                       -> {slug}  ("" when in no project)
   listdir(path)                            -> {path, parent, dirs, files}
-  import_doc(project,src)                  -> {name, path}  (links src in place, no copy)
-  import_url(project,url,name)             -> {name, path}
-  delete_doc(doc,project)                  -> {ok}  (linked docs are only unlinked)
-  rename_doc(doc,name,project)             -> {name, path}
+  rename_doc(doc,name)                     -> {name, path}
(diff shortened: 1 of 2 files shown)


amanbagrecha:
Merged latest `main` (the read-only file-gating feature, PR #106) and reconciled it with this flat-library rewrite. Also addressed the two Bugbot findings:

**1. Null `paths` crashes library (fixed).** `_lib_load` now does `data.get("paths") or []`, so a `library.json` with `"paths": null` (or hand-edited garbage) degrades to an empty list instead of raising `TypeError` on startup — same spirit as the existing corrupt-JSON `try/except`.

**2. Undo history path changed (intentional, no action).** This is inherent to removing projects. Undo snapshots were only ever stored under `projects/<slug>/.history/` for files that physically lived inside a project folder; those files go away with the project system. Standalone and linked docs already used the path-hash `cache/snapshots/<sha1>` dir (keyed by the doc's real absolute path, unchanged), so their undo stacks are unaffected. Undo history is transient cache (cap 10), so orphaned per-project stacks are harmless.

**Read-only feature integrated into the flat design:**
- `_save` and `rename_doc` raise `PermissionError` on a chmod -w original (via `os.access(W_OK)`); `open_doc` returns `writable`/`readonly_message`/`readonly_tooltip`.
- UI: loads `ro-badge.js`, shows the RO badge, disables Save and hides Rename for a read-only original.
- `delete_doc` is gone — `remove_from_library` only drops the reference and never touches the file, so it needs no gate.
- `tests/test_pdf_studio_readonly.py` retargeted to the flat-library API (5 tests pass locally).

amanbagrecha:
Fixed the `removeDoc` stale-doc bug in 003700b: when removing the currently open PDF, it now clears the open-doc state (viewer + `state.doc` + params) via `clearOpenDoc()` **before** opening the next library entry. If that next open fails (missing/encrypted), the viewer falls back to a clean empty state instead of staying stuck on the just-removed file.

PR #125, opened 2026-07-14
feat(duckdb): show load indicator during remote sort/filter scans

iamsdas:
## Problem

Changing a filter or clicking a column to sort on an already-loaded **remote-mounted** parquet freezes the grid for several seconds with no indication anything is happening — then columns abruptly start shimmering in.

**Root cause:** for a same-file reload, `load()` takes the `keepGrid` fast path and deliberately keeps the old grid on screen (no `Loading…` wipe), since warm reloads usually land in under a repaint. But sorted/filtered pages then run a **serial `positions` await** — a multi-second remote row-positions scan — *before* anything repaints. The "break-the-hold" timer that swaps in skeletons is (intentionally) only armed *after* that await, so for the entire scan the screen is static with zero feedback.

## Fix

Add an always-present 2px indeterminate load bar between the filter bar and the grid:

- `setLoading(true)` at the very top of `load()`, before any `await` — motion appears the instant you click a header or Apply.
- `setLoading(false)` at the end of `renderAll()` — the moment rows or skeletons actually paint, handing the in-flight signal off to the existing per-column shimmer. Also covers the three paths that bypass `renderAll` (the `!file` guard, the error catch, and the failed-rows branch in `applyBatch`).

The existing `keepGrid` / `deferRender` / positions-timer logic is untouched — the bar just fills the previously-silent gap. It's pure CSS + a one-line class toggle, and the bar is always 2px tall so toggling it never shifts layout.

## Testing

Ran the worktree dev server against a mounted parquet: sorting/filtering now shows the animated bar across the top of the grid during the positions scan, then the columns shimmer in as before.

### DIFF
--- a/fused_render/templates/duckdb/template.html
+++ b/fused_render/templates/duckdb/template.html
@@ -93,6 +93,27 @@
     font-size: 10px; line-height: 15px; text-align: center;
   }
   .save-count[hidden] { display: none; }
+  /* Indeterminate load bar between the filter bar and the grid. A sort/filter
+     change on a remote file keeps the old grid up (no "Loading…" wipe) and then
+     awaits a multi-second row-positions scan before anything repaints — so
+     without this the UI looks frozen. The bar turns on the instant load()
+     starts and off as soon as rows or skeletons paint. Always 2px tall so
+     toggling it never shifts the layout. */
+  #loadbar { flex: 0 0 auto; height: 2px; overflow: hidden; background: transparent; }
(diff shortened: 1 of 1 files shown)


iamsdas:
Addressed the Bugbot finding in 9296c2e: `renderAll()` now only clears the load bar when called from the load/applyBatch paint path (via a `fromLoad` flag). Staging renders (add/remove row) no longer turn off the indicator during an in-flight reload.

PR #130, opened 2026-07-14
perf: direct-to-store bypass for cold ranged reads + row-group pruning

iamsdas:
## Problem

Cold first load of a sorted view on a mount-backed parquet is latency-bound: the rclone serve's VFS layer serializes concurrent uncached range reads of one file at ~0.25s/seek, so the `positions` phase of the duckdb grid's two-phase load (256 scattered range reads on a 700MB Overture file) took **~63s** (45.8s in this PR's same-day control run).

### Direct-to-store bypass (server/shell — templates untouched)

`/api/fs/raw` now answers a **cold ranged GET from a native client with a 307 redirect** to the store's own URL, so duckdb's httpfs fans its range reads out against the store in parallel with its own pooled connections:

- **Private buckets:** a presigned link minted by the running rcd (`operations/publiclink` — S3/GCS/B2/Azure; credentials never leave rclone).
- **Anonymous S3 remotes** (which can't presign — "unsupported signer type noAuth"): the plain public object URL, built from the remote's config.
- Anything else resolves to `None` and stays on the serve. Capability and links are memoized per remote/object.

Routing guards:
- **Cold only** — once the background prefetch (#121) has landed the file in the serve cache, reads route back to the serve's local-disk replays.
- **GET+Range only** — presigned links are GET-signed (a HEAD fails their signature), and a whole-file GET through the serve is effectively the prefetch itself.
- **Native clients only** — browsers always send `Sec-Fetch-Mode` and a cross-origin redirect to S3 would die on CORS, so its absence gates the redirect. Browser requests and HEADs keep today's serve proxy.

A server-side proxy variant was measured first and rejected: urllib pays a fresh TLS handshake per range read (~2x slower on the point-read phase) and streams every byte through the server twice.

### Direct source_url rewrite in /api/run (7c383e2)

The 307 alone left ~3x on the table: duckdb's httpfs re-follows the redirect **per range read** and opens a fresh TLS connection to the store each time, defeating its connection pooling (measured on a cold unsorted open: schema 8.5s vs 3.4s direct, a 9-column page batch 14.5s vs 3.8s direct). `/api/run` now resolves the direct store URL server-side and **rewrites a raw-proxy `source_url` before the reader sees it** — cold mount-backed files only; warm (prefetched) files keep the raw URL so the serve replays ranges from local disk. Pages keep sending the raw URL and stay mount-agnostic; the 307 remains as the fallback for native clients hitting `/api/fs/raw` directly. Cold first-paint on the 700MB file: **23.1s → 13.2s**, and a follow-up (a4919fd) sets `parquet_metadata_cache=true` on the shared http connection — the object cache's etag revalidation HEAD 403s against GET-signed presigned URLs, so every run had been re-fetching the ~700KB footer — taking it to **8.9s**.

### Row-group pruning (duckdb reader — backend-agnostic)

An unfiltered sorted `positions` scan now reads the parquet footer's per-row-group min/max/null-count stats (already parsed — no extra bytes), computes a **provable bound on the window's cutoff value**, and restricts the scan to the row groups that can hold the top `offset+limit` rows via `file_row_number BETWEEN` ranges. It bails to the full scan (never worse than today) whenever the guarantee doesn't hold: missing stats, windows deep enough to reach the NULL tail, nested columns, or nothing pruned.

## Measurements (cold, ~700MB Overture address files, distinct untouched file per scenario)

| Scenario | positions | full first paint |
|---|---|---|
| Today's serve path (control) | 45.8s | 53.6s |
| Bypass via 307 | **4.7s (10x)** | **25.1s (2x)** |

Unsorted cold open (schema → parallel page batches): 23.1s via the 307 → 13.2s with the source_url rewrite → **8.9s** with the parquet metadata cache.

- Pruning is honestly a **no-op on Overture's `number` column** — the top-100 values are 'SN…' strings present in all 256 row groups, so no pruning scheme (verified by simulating a dynamic probe too) can exclude anything there. Where stats have locality it works: `postcode:desc` went **29.3s → 16.4s (1.8x)** through the serve, and file-ordered synthetic data prunes 5 groups → 1.
- Remaining cold cost is genuinely cold column bytes at S3 bandwidth plus per-run footer refetches, which the background prefetch erases across the session.
- Pruned positions proven equal to full-scan ground truth across 40 sort/offset/column combinations (NULL tails, outliers, constant columns, scattered data).

## Tests

10 new tests: pruned-vs-full-scan equivalence, prune/bail behavior on correlated vs scattered vs NULL-tail columns, presigned-link preference + caching, anonymous-S3 public URL construction + quoting, none-capability memoization, 307/proxy routing gates (native vs browser vs HEAD), and the /api/run source_url rewrite (cold rewritten, warm/non-mount/foreign URLs untouched). Full suite green except 4 failures that reproduce on the untouched base commit (stale `5Gi` mounts assertions ×2, templates_api ×2).

### DIFF
--- a/fused_render/server.py
+++ b/fused_render/server.py
@@ -26,7 +26,7 @@ import time
 import traceback
 import urllib.error
 import urllib.request
-from urllib.parse import parse_qsl
+from urllib.parse import parse_qsl, urlsplit
 
 from fastapi import Body, FastAPI, Header, Request, WebSocket, WebSocketDisconnect
 from fastapi.responses import (
@@ -1593,6 +1593,33 @@ def create_app(start_dir: str) -> FastAPI:
         py = body.get("py")
         html = body.get("html")
         params = body.get("params") or {}
+
+        # Cold mount-backed reads: swap the raw-proxy source_url for the
+        # store's own URL before the reader sees it. The /api/fs/raw 307
+        # already sends cold ranged GETs to the store, but a redirect
+        # defeats httpfs connection pooling — duckdb re-follows it per
+        # range read and opens a fresh TLS connection to the store each
+        # time (measured ~3x on a cold open: schema 8.5s vs 3.4s, a
+        # 9-column page 14.5s vs 3.8s). Handing the reader the direct URL
+        # up front lets httpfs pool its store connections normally. Done
+        # here in the server, not in templates: pages keep sending the raw
+        # URL and stay mount-agnostic. Warm files (prefetch landed) keep
+        # the raw URL so the serve replays ranges from local disk; the
+        # explicit schedule() below matters because a direct-reading run
+        # never touches /api/fs/raw, which is otherwise the only place the
+        # prefetch learns a file is in use.
+        src = params.get("source_url")
+        if isinstance(src, str):
+            parts = urlsplit(src)
+            fpath = dict(parse_qsl(parts.query)).get("path")
+            if parts.path.endswith("/api/fs/raw") and fpath:
+                upstream = shell_mounts.serve_url_for(fpath)
+                if upstream is not None and not shell_prefetch.is_done(fpath):
+                    shell_prefetch.schedule(fpath, upstream)
+                    direct = await asyncio.to_thread(
+                        shell_mounts.upstream_url_for, fpath)
+                    if direct:
+                        params = dict(params, source_url=direct)
 
         if not py:
             return _error("request body must include 'py': a path to a Python file")
--- a/fused_render/server.py
+++ b/fused_render/server.py
@@ -33,6 +33,7 @@ from fastapi.responses import (
     FileResponse,
     HTMLResponse,
     JSONResponse,
+    RedirectResponse,
     Response,
     StreamingResponse,
 )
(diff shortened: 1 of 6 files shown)


iamsdas:
Follow-up commit 7c383e2: /api/run now resolves the direct store URL server-side for cold mount-backed files and rewrites the raw-proxy `source_url` before the reader sees it — the 307 defeated httpfs connection pooling (fresh S3 TLS handshake per range read, ~3x amplification). Cold first-paint on the 700MB overture file: 23.1s → 13.2s. The 307 stays as fallback for native clients hitting /api/fs/raw directly.

iamsdas:
Follow-up commit a4919fd (idea borrowed from reviewing how hyparquet stays fast — parse the footer once, never re-fetch it): the shared http connection now sets `parquet_metadata_cache=true`. The object cache alone revalidates metadata via etag/last-modified, and presigned GET-signed URLs 403 that HEAD — so every reader run re-fetched the ~700KB footer. Cold first-paint on the 700MB overture file: **23.1s → 13.2s (source_url rewrite) → 8.9s** (this).

PR #132, opened 2026-07-15
feat(windows): ship Explorer "Open with" registration in the wheel

amanbagrecha:
- Track `fused_render/winopen.py`, `assets/fused-render.ico`, `assets/file_icons/file.ico`, and the `scripts/windows/` helpers — previously hidden via `.git/info/exclude`, which hatchling ignores, so hand-built wheels shipped them while clean-checkout/CI wheels failed `--register` with `No module named fused_render.winopen`.
- Restore `[project.gui-scripts] fused-render-open = "fused_render.winopen:main"` in `pyproject.toml`, so every wheel install generates the windowless launcher exe.
- `extensions()`: register each registry key's final suffix only (Explorer keys on the last suffix) — `.csv.zst` → `.zst`, the `.*.json` glob → `.json`; adds the previously-unregistered `.zst`.
- `_view_url()`: mirror the frontend codec (`router.ts urlForFsPath`) — backslashes normalize only for drive-letter paths, so UNC paths (`\server\share\...`) encode as one segment instead of being mangled.
- Server discovery rewritten: spawns serialize under a named mutex (`Local\fused-render-open`); racing double-clicks wait and re-probe instead of double-spawning; discovery never waits, so a stale portfile can't hang the opener. Deletes the `_settle`/`_pid_alive` heuristics.
- `_spawn()`: drop redundant `close_fds`; skip `proc.kill()` on a child that already died during boot so the failure path always reaches pidfile cleanup and the log-path error.
- `executor.py`: `CREATE_NO_WINDOW` on the per-run subprocess (win32) — no console window flashing per `fused.runPython` call under a windowless server.
- Docs: README "Windows: Explorer Open with" section; DECISIONS.md D105.
- Tests: `tests/test_winopen.py` (8) — extension reduction, view-URL encoding, stale-portfile fallthrough, lock re-probe.

---

> [!NOTE]
> **Medium Risk**
> Adds Windows registry mutations and detached process spawning (HKCU-only, no elevation), plus user-facing file-open behavior that must work consistently across CI-built wheels.
> 
> **Overview**
> **Windows Explorer integration** is now part of the published wheel: `fused_render/winopen.py`, icon assets, and `scripts/windows/` helpers are tracked, with `[project.gui-scripts] fused-render-open` restored in `pyproject.toml` (decision **D105**). This fixes wheels built from clean CI checkouts missing `winopen` while ad-hoc local builds accidentally included it.
> 
> The **`fused-render-open`** entry point registers HKCU file associations for every previewable extension from `templates/registry.json` (`--register` / `--unregister`), or opens a file by reusing a live native server (via `/api/config` + drive-letter check), starting a **detached, windowless** `fused-render serve` under a spawn mutex when needed, then launching the matching `/view/...` URL in the browser. State lives under `%LOCALAPPDATA%\fused-render\` (pid/port files, `server.out.log`).
> 
> **`executor.py`** sets `CREATE_NO_WINDOW` on Windows subprocess runs so preview execution does not flash a console when the server was started from the Explorer opener.
> 
> Docs add a README Windows section; **`tests/test_winopen.py`** covers URL encoding, extension listing, server discovery, and lock/reprobe behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 329223e7306d5ade6a1bbec3910f457cf9ab2b58. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amanbagrecha:
Addressed both Bugbot findings in 7acc4da:

- `extensions()` now excludes any registry key containing `*` (e.g. `.*.json`, a template wildcard, not a real extension) so `--register` no longer writes bogus ProgID/OpenWithProgids entries for it.
- `test_view_url_encodes_each_segment` stubs `os.path.abspath` so it no longer depends on Windows-specific path-separator behavior.

Also trimmed `assets/file_icons/` down to just `file.ico` for now — the per-category icons aren't rendering correctly yet, so shipping the single generic fallback until that's sorted out (`_icon_for_ext` already falls back to the fused-render diamond for any variant whose `.ico` is missing, so this is a safe reduction, not a code change).

amanbagrecha:
Addressed the `close_fds` finding in bfac135.

On closer inspection the `ValueError` mechanism no longer applies to our supported Python range — the restriction on `close_fds=True` + redirected streams was lifted in Python 3.7, and this project requires >=3.10 (verified empirically: no error on 3.12). But the recommendation still holds on style grounds: `close_fds` already defaults to `True`, so passing it was redundant *and* inconsistent with every other detached spawn in the repo (docs/latex/usd install workers, the tile-server daemons, mounts), none of which set it. Removed it — behavior-preserving.

Also had two independent reviewers verify the whole PR end-to-end, including building a wheel from this branch and confirming `fused_render/winopen.py` + the `[gui_scripts] fused-render-open` entry point + assets + `registry.json` all ship in it (which is the packaging bug this PR fixes). Both came back clean.

amanbagrecha:
Addressed both new findings in 6f6d062 (plus doc tidy-ups in a2d57cb):

- **Stale portfile hang:** `find_running_server` now only grants the boot grace period when a new `_pid_alive` check confirms the recorded server is still alive (probed via `OpenProcess`/`WaitForSingleObject` — `os.kill(pid, 0)` terminates the process on Windows). A stale portfile whose port is now held by another app falls straight through to the range scan instead of blocking 15s. A genuinely-booting server is still awaited.
- **Race check grace:** the post-`pick_port` guard now uses `_settle` instead of `_fused_server`, so a peer double-click that bound the port but isn't HTTP-ready yet is awaited rather than double-spawned onto.

Verified the whole PR against every finding raised from the start (27-check harness: view-URL drive/UNC/POSIX/special, extensions final-suffix reduction, close_fds, `_pid_alive` live+reaped, stale-portfile-no-hang, booting-server-awaited, racing-guard-no-double-spawn) — all pass. Added 3 focused regression tests (`tests/test_winopen.py`, 8 total; the pid probe skips on non-Windows). Full suite shows no new failures vs `main`.

amanbagrecha:
Rewrote the discovery/race handling instead of patching it a fifth time (b6a7ff3):

The last four findings all targeted the same root cause — discovery tried to *infer* whether a racing double-click was mid-boot (pidfiles, bound-but-silent ports, grace periods), and every inference had a hole. Replaced inference with actual mutual exclusion: server startup now happens under a **named mutex** (`Local\fused-render-open`). The winning opener holds it until its server answers; racing peers wait on it and then re-probe, finding the ready server. `_settle` and `_pid_alive` are deleted outright; `find_running_server` never waits — it reports who answers *now*. A crashed winner surfaces as `WAIT_ABANDONED` (ownership passes to the next waiter), and a mutex failure/timeout degrades to proceeding unlocked rather than hanging. Also guarded `proc.kill()` on a child that already died during boot, so the failure path always reaches pidfile cleanup and the log-path error.

Separately (329223e): the executor's per-run subprocess now sets `CREATE_NO_WINDOW` on Windows — under a windowless server (spawned detached by the opener), every `fused.runPython` call was flashing a console window on screen.

Verified end-to-end on Windows: real spawn through the new path (boots on 1777, answers `/api/config`, second call reuses instead of respawning, clean teardown), real cross-thread mutual exclusion on the actual mutex, stale-portfile instant fallthrough, no-double-spawn re-probe, bounded requested-port behavior. 8/8 tests pass.

PR #155, opened 2026-07-16
perf(prefetch): size-gate small files, bound the maps, async I/O model

iamsdas:
Follow-up to #149 review findings (prefetch double-read + unbounded map growth), plus an async restructure of the worker.

## Problem
`shell/prefetch.py` streams a whole file through the rclone serve in the background so future scattered reads replay from the local VFS cache — a big win for large, latency-bound files (a cold sort on a 644MB parquet drops from ~30s to disk-speed). But #149 made whole-file GETs 307-redirect to the store, and costs surfaced for zarr/netcdf stores made of **thousands of tiny chunk objects**:

1. **Double-read waste** — the redirected client reads each chunk from the store directly, then the background worker re-streams the same chunk through the serve to warm a cache the client never reads back.
2. **Unbounded growth** — every distinct chunk path minted a permanent `_jobs`/`_touched` entry (no eviction), leaking memory for the life of the process.
3. **Thread + latency storm** — `schedule()` spawned one daemon thread per distinct path, and the size-gate HEAD sat behind a 5s delay *and* the single download semaphore, so thousands of chunks meant thousands of threads each sleeping 5s then serializing a pointless HEAD.

## Fix
- **`MIN_BYTES` floor (8Mi, `FUSED_RENDER_PREFETCH_MIN_BYTES`)**: skip files too small to benefit. **Size — not redirect-vs-proxy — is the correct discriminator**: cold reads of a big parquet always hit the redirect path first, so gating on the redirect branch would disable prefetch entirely.
- **`MAX_TRACKED` cap** (2048, `FUSED_RENDER_PREFETCH_MAX_TRACKED`): evict terminal jobs so the maps stay bounded; queued/running jobs are never dropped. Eviction is **LRU by last access** (`_touched`), not completion time — since `schedule()` refreshes `_touched` on every read, a `done` file still in use stays warm and keeps its `is_done` routing.
- **Async I/O model**: replace thread-per-path with a single dedicated `asyncio` loop; blocking `urllib` runs via `asyncio.to_thread` on its bounded shared executor. Thousands of chunk reads are now cheap coroutines on one loop thread instead of thousands of daemon threads. **Zero new dependencies** (stdlib `urllib` kept).
- **Phase split**: the size-gate HEAD (cheap) runs immediately and concurrently — off the download semaphore, no `START_DELAY`. A sub-floor chunk is dismissed in one round trip instead of sleeping 5s and queueing behind the download slot. Only the whole-file *stream* (expensive) stays serialized and delayed.

`schedule()` remains a plain sync call (submits via `run_coroutine_threadsafe`); public API and behavior unchanged.

## Tests
Added `test_size_gate_skips_small_files`, `test_evicts_oldest_terminal_jobs_over_cap`, `test_size_gate_decides_without_start_delay`, and `test_eviction_is_lru_by_access_not_completion`. Full suite: 762 passed, 3 skipped.

## Review
Cursor Bugbot flagged "Done eviction breaks active warm routing" — fixed in ae83ea0 by ordering eviction by `_touched` (LRU) rather than completion `at`.

## Not addressed here
Each distinct chunk path still incurs one HEAD on its first read before the floor skips it (now cheap and concurrent, no longer thread-backed or serialized). Eliminating the HEAD entirely would need a size source that avoids a round trip (e.g. piggybacking a listing the daemon already did), which reopens the stat-free-redirect tradeoff from #149 — deliberately out of scope.

### DIFF
--- a/fused_render/shell/prefetch.py
+++ b/fused_render/shell/prefetch.py
@@ -18,9 +18,21 @@ rclone's fingerprint check invalidates it if the object changes upstream.
 
 Triggered from the /api/fs/raw proxy (the one path every remote byte
 already flows through — templates stay mount-agnostic), so scheduling
-must be cheap, non-blocking, and never raise. Everything slow (the HEAD
-for the size gate, the download) happens on a daemon worker thread; a
-process-wide semaphore keeps at most one file prefetching at a time.
+must be cheap, non-blocking, and never raise. `schedule()` just records
+the job and submits a coroutine to a single dedicated prefetch event loop
+(one background thread for the whole process, not one thread per file —
+a zarr store's thousands of chunk reads would otherwise mint thousands of
+threads). The blocking urllib calls run via asyncio.to_thread on that
+loop's bounded shared executor.
+
+Two phases with very different cost profiles, kept separate:
+  1. Decide — a HEAD for the size gate. Cheap; runs immediately and
+     concurrently for every scheduled file, so a sub-MIN_BYTES chunk is
+     dismissed in one round trip without waiting behind anything.
+  2. Stream — the whole-file download. Expensive; serialized by an
+     asyncio semaphore (one file at a time: two big streams would evict
+     each other from the LRU serve cache) and delayed by START_DELAY_S so
+     the interactive read that triggered us wins its first cold seeks.
 
 The serve surfaces transient store errors as HTTP 500s mid-stream
 (observed on S3), so the worker fetches in ranged chunks and resumes with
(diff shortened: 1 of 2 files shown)


iamsdas:
Addressed the Bugbot finding ("Done eviction breaks active warm routing") in ae83ea0: `_evict_locked` now orders terminal jobs by `_touched` (last interactive access) instead of completion `at`. Since `schedule()` refreshes `_touched` on every read, a `done` file still being read stays warm and keeps its `is_done` routing; only genuinely stale entries are evicted. Added `test_eviction_is_lru_by_access_not_completion`. 762 passed, 3 skipped.

PR #174, opened 2026-07-17
Consolidate Zarr templates: delete legacy zarr, gate zarr_aoi on folder contents

iamsdas:
## Summary

- **Deletes the legacy `zarr` template.** `zarr_aoi` (the server-side AOI tile-streaming map viewer) is now the single Zarr previewer. `netcdf/` was verified self-contained, so nothing else depended on `zarr/`.
- **Recognizes unsuffixed Zarr stores.** A folder that *is* a Zarr store but isn't named `*.zarr` (e.g. `mur-sst/zarr-v1`) previously got no Zarr template — it only matched the generic `/` directory key. It now resolves `zarr_aoi` via a content-based `condition.py` gate that probes for group-root markers (`.zmetadata` / `zarr.json` / `.zgroup`) with bounded, short-circuiting `isfile` checks and **never lists/walks** the directory (the remote-timeout risk). `zarr_aoi` opens via `zarr.open_group()`, so the gate offers **group roots only**: v2 bare arrays (`.zarray`) are excluded by marker, and a v3 `zarr.json` root — shared by groups and bare arrays — is disambiguated by a bounded single-file read of `node_type` (offered only when `"group"`; fails closed otherwise). This avoids offering-then-erroring on a bare-array root.
- **Drops the dead sentinel-file keys** `.zgroup` / `.zattrs` / `.zmetadata` (they only pointed at the deleted template and are unreachable hidden dotfiles).
- **Gating stays implicit (unchanged mechanism).** A template folder shipping a `condition.py` is conditional on *every* binding that references it (CT-12, as on `main`) — no new server/API machinery is introduced. Because `zarr_aoi` ships a gate, it is a **conditional peer** on both its bindings: the built-in `_listing` is the immediate default and the map joins the switcher once the background gate confirms the store. On a `*.zarr` folder that verdict is effectively instant (the gate's zero-I/O name fast-path returns `True` with no filesystem calls).

## Visuals

Backend/registry change — no user-facing UI surface to screenshot. The delta is in how a directory resolves and defaults its templates:

```mermaid
flowchart TD
    D["Open a directory"] --> M{"basename matches<br/>a registry dir key?"}
    M -->|"ends in .zarr → .zarr/ key<br/>[zarr_aoi, _listing]"| Z["_listing = default<br/>zarr_aoi = gated peer"]
    M -->|"otherwise → / key<br/>[_listing, preview, zarr_aoi]"| U["_listing = default<br/>zarr_aoi = gated candidate"]
    Z --> G
    U --> G{"zarr_aoi condition.py<br/>(lazy, background)"}
    G -->|".zarr name (zero-I/O fast path)"| KEEP["offer zarr_aoi in switcher"]
    G -->|".zmetadata / .zgroup present<br/>(bounded isfile, no listing)"| KEEP
    G -->|"zarr.json present"| V3{"read node_type<br/>(one bounded file read)"}
    V3 -->|"group"| KEEP
    V3 -->|"array / missing / unreadable<br/>(fail closed)"| DROP
    G -->|"no group marker"| DROP["drop zarr_aoi → plain folder"]
```

`zarr_aoi` ships one `condition.py`, so it is gated on both bindings (implicit CT-12 gating). On `*.zarr` folders the gate's name fast-path confirms the store with no I/O, so the map appears in the switcher immediately; the listing is the default open view in both cases.

## Usage

Automatic — no new user-invocable command and no registry-authoring syntax change:

- Open a `*.zarr` folder → **listing** by default, with the `zarr_aoi` **map** joining as a peer (verdict is instant via the name fast-path).
- Open a Zarr store without the suffix (contains `.zmetadata`, `zarr.json`, or `.zgroup`) → `zarr_aoi` offered after the background gate.
- Open any plain folder → gate returns `False`, `zarr_aoi` dropped, normal `_listing` / `preview`.

## Test Plan

- [x] **`tests/test_templates.py` — 59 passed.** Covers: `.zarr/` and `/` both resolve `zarr_aoi`; the `condition.py` gate (name fast-path, group markers `.zmetadata` / `zarr.json` / `.zgroup`, v2 bare-array-not-offered, v3 `node_type` group-vs-array disambiguation incl. missing-`node_type` fail-closed, a `no-listing` monkeypatch asserting the gate never `listdir`/`scandir`/`glob`s, fail-closed on exceptions); legacy `zarr` template fully removed.
- [x] **SPEC** updated for the rename (`zarr` → `zarr_aoi`, §7.2 table, PT-8/PT-13, sentinel section) and to correct the `.zarr` default: the store now opens to the built-in listing with the map as a `condition.py`-gated peer (previously documented as map-by-default). No new decision record — the gating mechanism is unchanged from `main`.
- [x] **Live end-to-end verification (v2 + v3, local + remote-over-mount).** Brought the full server up via `scripts/dev.sh` (3.12 `[bundled]` venv + bun-built frontend), registered a v2-consolidated store as a real rclone/nfs mount, and drove `zarr_aoi` through the server: `/api/fs/stat` marks it `conditional`, `/api/fs/conditions` → `true`, `/api/fs/raw` returns `206 Partial Content` (Range honored), `/meta` → `zarr_format=2`, and `/probe` + `/tile` decode a real blosc chunk through `/api/fs/raw`. `/stats` shows `lists=0` (no listing storm) and the NFS mount stayed mounted (not wedged). v3 (zstd) and local v2 non-consolidated confirmed as controls.
- [ ] **Full suite not run in this worktree.** The untargeted suite still needs the same 3.12 `[bundled]` venv used above for *every* test plus specific fixtures; the changed surface here is only `templates/`, `registry.json`, and `SPEC.md`, exercised by the passing `test_templates.py` and the live e2e above. Run in a fully-configured checkout for the whole-suite number.


fused_render/templates/zarr_aoi/condition.py
```diff
@@ -0,0 +1,54 @@
...
+avoid; a targeted `isfile`/HEAD stays constant-time.
+
+Fails closed: any exception while probing returns False (the template is dropped
+quietly), and a path that isn't a directory (and isn't `.zarr`-suffixed) is
+False. Self-contained — the module is exec'd standalone (not imported as part of
+a package), so it imports only stdlib.
+"""
+
+# v2 group/array, v3, and consolidated metadata — cheapest/most-likely first.
+_STORE_MARKERS = (".zmetadata", "zarr.json", ".zgroup", ".zarray")
```
iamsdas: **Correctness — `.zarray` over-offers `zarr_aoi` for stores it cannot open.**

`zarr_aoi` opens the path with `zarr.open_group()` (`tile_server.py:418`) and has **no `open_array` fallback**. A directory whose root marker is `.zarray` (a v2 *bare array*, no `.zgroup`) is a valid Zarr store, so this gate returns `True` and offers `zarr_aoi` — but `open_group` then raises on it, so the user gets an error overlay instead of a clean "not previewable" fallback. (A v3 `zarr.json` with `node_type: array` has the same problem, but can only be told apart by reading the file, which would break the no-I/O-listing constraint.)

The legacy registry only ever keyed on `.zgroup`/`.zattrs`/`.zmetadata` — never `.zarray` — precisely because the previewer is group-oriented. Suggest dropping `.zarray` from `_STORE_MARKERS` (which also removes one probe from the common path, see the efficiency note below).

```suggestion
_STORE_MARKERS = (".zmetadata", "zarr.json", ".zgroup")
```

fused_render/templates/zarr_aoi/condition.py
```diff
@@ -0,0 +1,54 @@
...
+        if name.lower().endswith(".zarr"):
+            return True
+
+        # Otherwise it must be a directory carrying a store marker. Bail early
+        # (no marker probes) if it isn't even a directory.
+        if not os.path.isdir(path):
+            return False
+
+        # Bounded, short-circuiting probes — first hit wins, never a listing.
+        return any(os.path.isfile(os.path.join(path, m)) for m in _STORE_MARKERS)
```
iamsdas: **Efficiency — up to 5 stat/HEAD calls per *non-store* directory, on the browse hot path.**

Because `zarr_aoi` is now on the universal `/` key, this gate runs (in the background) for **every** directory the user opens. For the common case — a plain folder that is *not* a Zarr store — the miss path costs `isdir` (1 stat) + one `isfile` per marker (4 stats) = **5 remote HEADs per folder** on a slow mount. That is bounded and backgrounded (it never blocks `_listing`), so it is acceptable, but worth being explicit about: browsing a remote tree now issues extra per-folder round-trips it did not before. Trimming `_STORE_MARKERS` (e.g. dropping `.zarray`) directly reduces this worst case.

fused_render/templates/zarr_aoi/condition.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def main(path: str) -> bool:
+    import os
+
+    try:
+        # Zero-I/O name fast path: strip any trailing slash, then a case-
+        # insensitive `.zarr` suffix decides True with no filesystem calls.
+        name = os.path.basename((path or "").rstrip("/"))
+        if name.lower().endswith(".zarr"):
```
iamsdas, line 66: **Minor / latent — name fast-path returns `True` for a *file* or nonexistent `.zarr` path.**

`main("/x/foo.zarr")` returns `True` with no `isfile`/`isdir` check, even when the path is a regular file or does not exist. Harmless **today**: the gate only runs on entries the registry matched, and both zarr keys (`.zarr/`, `/`) are directory-only, so a `.zarr` *file* never reaches this condition. But it is a footgun — if `zarr_aoi` is ever added to a *file* key, this silently green-lights non-directories. Consider guarding the fast-path with `os.path.isdir(path)` (it stays zero-listing), or leave a comment noting the invariant it relies on.

fused_render/templates/zarr_aoi/condition.py
```diff
@@ -0,0 +1,54 @@
...
+
+        # Otherwise it must be a directory carrying a store marker. Bail early
+        # (no marker probes) if it isn't even a directory.
+        if not os.path.isdir(path):
+            return False
+
+        # Bounded, short-circuiting probes — first hit wins, never a listing.
+        return any(os.path.isfile(os.path.join(path, m)) for m in _STORE_MARKERS)
+    except Exception:  # noqa: BLE001 — any probe error: fail closed, quietly
+        return False
```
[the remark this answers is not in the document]
iamsdas: Fixed in 70a41fc..f81a4d9 (commit range on this branch). Confirmed the regression and fixed it properly with **key-scoped conditions (D116)**:

- Gating is now **opt-in per registry binding** via a trailing `?` on the mode token. `condition.py` presence alone no longer gates — a binding must ask for it.
- `.zarr/` = `["zarr_aoi", "_listing"]` (no `?`) → `zarr_aoi` is **unconditional and the default** again (map shows, `_listing` stays a peer), satisfying PT-13.
- `/` = `["_listing", "preview", "zarr_aoi?"]` → gated only on the universal key, so unsuffixed Zarr-store folders still resolve `zarr_aoi` while plain folders drop it.
- Existing conditional templates opt in explicitly: `.parquet` `h3`→`h3?`, `.toml` `canvas`→`canvas?` (their gating is preserved).

Behavioral confirmation: `.zarr/` → `zarr_aoi` default; `.parquet`/`.toml` → `h3`/`canvas` GATED; plain dir → `zarr_aoi` GATED. SPEC CT-12/PT-8 updated, D116 records the flip. Thanks for the catch.

fused_render/templates/zarr_aoi/condition.py
iamsdas: Fixed in f256e96: dropped `.zarray` from the store markers. `_STORE_MARKERS` is now group-roots only (`.zmetadata`, `zarr.json`, `.zgroup`), and a `.zarray`-only dir is asserted NOT offered (`test_bare_array_dir_not_offered`).

fused_render/templates/zarr_aoi/condition.py
iamsdas: Addressed in f256e96: dropping `.zarray` also trims the common non-store miss path to `isdir` + 3 `isfile`. The per-directory background cost is documented in the condition.py docstring; it stays bounded and never blocks `_listing`.

fused_render/templates/zarr_aoi/condition.py
iamsdas: Addressed in f256e96: kept the fast-path zero-I/O (adding an isdir stat would defeat it) and documented the invariant it relies on — the gate only ever runs on registry-matched entries, and both zarr keys (`.zarr/`, `/`) are directory-only, so a `.zarr`-named *file* never reaches this condition.

fused_render/templates/zarr_aoi/condition.py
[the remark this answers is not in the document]
iamsdas: Update — correcting the record on this thread: the **D116 `?` opt-in mechanism** referenced in my reply above was **reverted** (owner decision — this PR is intentionally a template + registry change only, with no server/dispatch changes).

The current design deliberately keeps `zarr_aoi` conditional on both bindings. On a `*.zarr` store the built-in `_listing` is the default open view, and the map (`zarr_aoi`) joins the mode switcher as a `condition.py`-gated peer once its background verdict lands — near-instant on `*.zarr` names via the gate's zero-I/O name fast path. `SPEC.md` §7.2 / PT-8 / PT-13 were updated to document listing-as-default.

So the behavior Bugbot flagged here is **intended**, not a regression — leaving this resolved on that basis.

fused_render/templates/zarr_aoi/condition.py
```diff
@@ -0,0 +1,65 @@
...
+        if name.lower().endswith(".zarr"):
+            return True
+
+        # Otherwise it must be a directory carrying a store marker. Bail early
+        # (no marker probes) if it isn't even a directory.
+        if not os.path.isdir(path):
+            return False
+
+        # Bounded, short-circuiting probes — first hit wins, never a listing.
+        return any(os.path.isfile(os.path.join(path, m)) for m in _STORE_MARKERS)
```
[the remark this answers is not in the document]
iamsdas: Fixed in `38c4970`. The `zarr.json` probe is now group-aware: on a `zarr.json` hit the gate does a **bounded single read** of that one file and offers `zarr_aoi` only when `node_type == "group"`. A v3 bare-array root (`node_type: "array"`), a missing/blank `node_type`, or any read/JSON error **fails closed** (not offered) — mirroring the v2 `.zarray` exclusion. `.zmetadata` / `.zgroup` stay pure `isfile` probes (inherently group-root markers). Reading one named file is constant-time and does not touch the no-listing invariant (the ban is on directory enumeration).

Tests (TDD, confirmed red before the fix): `test_v3_group_dir_offered`, `test_v3_bare_array_dir_not_offered`, `test_v3_zarr_json_without_node_type_not_offered`. Verified `node_type` values against zarr 3.2.1.

PR #217, opened 2026-07-21
fix(mounts+pyramid): keep the mount attached and rendering when opening/switching COGs

iamsdas:
## Summary

Opening — and especially **switching between** — `.tif` files inside an rclone-backed mount (e.g. the 44k-entry `source.coop/.../sentinel2-temporal-mosaics`) could **disconnect the mount** ("source.coop unresponsive") and/or **crash the pyramid render** with `worker exited with code -11`. This PR fixes both failure classes and the cold-render performance that sat on top of them:

- **No kernel FS probe ever touches a mount path.** A cold `os.stat`/`isfile`/`listdir` on an rclone-NFS mount forces rclone to enumerate the *entire* parent S3 prefix, trips the macOS NFS deadman, and drops the mount. Three open-flow handlers still did this — the `/api/fs/raw` warm-read fallthrough, `/api/session` GET/PUT, and `/api/recents` — now all route existence/shape through the rclone rc API (`rc_kind_for`/`_mount_probe`), never the kernel. Files-only gates reject a mount **directory** (a `"dir"` is not a file) while failing **open** on an indeterminate rc probe.
- **Cold remote-COG analyze fits the run budget.** Range reads use a 4 MiB block (was 64 KiB: ~195s → ~31s), the built-in *and* fused engines share a single 60s timeout, and the range cache is byte-bounded to a working-set-sized 256 MiB (a too-small cap thrashes: a 42 MiB COG went 23s → >180s timeout).
- **Fixed a native `fork()`-after-PROJ SIGSEGV.** The process spawning the reader worker has `libproj` resident with a live `proj.db` SQLite handle; `subprocess`'s `fork()+exec` ran PROJ's `pthread_atfork` child handler in the fork child, which segfaulted before `exec`. Workers now spawn via `posix_spawn` (no atfork handlers).

## Visuals

Where each fix sits in the open-a-mounted-COG flow — the removed kernel probes (✗) and the fork hazard are the delta:

```mermaid
flowchart TD
    Open["Open /view/&lt;mount&gt;/B0x.tif?_mode=pyramid"] --> Session["/api/session GET/PUT"]
    Open --> Recents["/api/recents/open"]
    Open --> Raw["/api/fs/raw warm-read fallthrough"]

    Session -->|"was: kernel os.path.isfile ✗"| GateS{"_is_file_mount_safe"}
    Recents -->|"was: kernel os.path.isfile ✗"| GateR{"_mount_exists"}
    Raw -->|"was: kernel os.stat ✗"| GateW{"_mount_probe"}

    GateS -->|rc_kind_for| RC[("rclone rcd operations/stat")]
    GateR -->|rc_kind_for| RC
    GateW -->|rc list| RC
    RC -->|"file to ok / dir,missing to 404 / indeterminate fail-open"| Analyze

    Analyze["/api/run to executor to _child.py to main()"] --> Spawn["spawn reader worker"]
    Spawn -->|"was: fork()+exec to PROJ atfork to SIGSEGV -11"| PosixSpawn["posix_spawn (no atfork handlers)"]
    PosixSpawn --> Worker["worker: 4 MiB blocks / 256 MiB cache / 60s budget"]
    Worker --> Render["pyramid renders / mount stays attached"]
```

## Usage

No API or config change — the behavior is transparent. To exercise it: open a large `.tif` inside a managed mount in the pyramid view and switch between several files, e.g.

```
http://127.0.0.1:1777/view/Users/<you>/.fused-render/mounts/source.coop/earthgenome/sentinel2-temporal-mosaics/01KAU_2023-01-01_2024-01-01/B02.tif?_mode=pyramid
```

The mount stays attached across switches and each COG renders its full overview pyramid on a cold cache.

## Test Plan

- [x] **Automated — new/updated:**
  - `test_fs_raw_warm_read_mount_safe.py` — warm-read fallthrough resolves via rc, never a kernel stat.
  - `test_server_session.py` — parametrized `_is_file_mount_safe` gate (file/indeterminate pass, dir/missing 404) with a loud kernel-`isfile` guard; PUT dir-rejection.
  - `test_shell_recents.py` — recents existence on `rc_kind_for` (four states), dir filtered, open-time dir rejection.
  - `test_pyramid_mount.py` — 4 MiB default block, byte-bounded LRU derivation + eviction.
  - `test_worker_forksafe.py` — worker spawns take the `posix_spawn` path (no fork); behavioral `pthread_atfork` guard (skips where the symbol isn't dynamically resolvable, e.g. glibc).
  - `test_shell_mounts.py` / `test_shell_prefetch.py` — stub `rc_list_dir` for the warm paths.
- [x] **Suite green** locally: pyramid + prefetch + fork-safe + session + recents + mount-safety → **122 passed, 3 skipped**. CI reddened only on the atfork probe's glibc symbol resolution — fixed in `877a244`.
- [x] **Manual (live mount, `source.coop`):** switched across B08 → TCI → B03 → B05 → B06 → B02 (40–232 MB) — mount never disconnected; B02 (158 MB), the reliable `-11` crasher, renders the full pyramid **with concurrent prefetch running** and **0 new PROJ-atfork crash reports** across 25+ analyzes.

### Known separate issues (pre-existing, out of scope here)
- A distinct worker-side `SIGABRT` (uncaught GDAL `std::length_error`) on some files under concurrent prefetch — surfaces as a clean "worker failed", does **not** wedge the mount.
- TCI.tif (232 MB) occasionally exceeds the 60s cold-read budget.

### DIFF
--- a/fused_render/engine.py
+++ b/fused_render/engine.py
@@ -80,16 +80,22 @@ def available() -> bool:
 
 def get_backend():
     # Lazy singleton: importing the backend pulls in the fused package tree,
-    # and constructing it is only needed once per server process. 30s matches
-    # the built-in executor's per-run timeout.
+    # and constructing it is only needed once per server process. 60s matches
+    # the built-in executor's per-run timeout (executor.DEFAULT_TIMEOUT) — a
+    # cold overview read of a large remote COG legitimately takes ~30-40s, so
+    # the two engines must agree or the cold pyramid analyze dies at 30s here.
     global _backend
     if _backend is None:
         from fused.agent_core.backends.local.python_compute import LocalPythonComputeBackend
 
+        from fused_render.executor import DEFAULT_TIMEOUT
+
         # cache_storage=None disables result caching explicitly (PY-9: fresh
         # execution every call). It is the upstream default today, but we may
         # track a nightly wheel — don't rely on a default staying put.
-        _backend = LocalPythonComputeBackend(timeout_seconds=30, cache_storage=None)
+        _backend = LocalPythonComputeBackend(
+            timeout_seconds=int(DEFAULT_TIMEOUT), cache_storage=None
+        )
     return _backend
 
 
--- a/fused_render/executor.py
+++ b/fused_render/executor.py
@@ -38,7 +38,11 @@ CHILD = os.path.join(os.path.dirname(__file__), "_child.py")
 # Built-in helpers run from the staged core-templates copy, not the bundle, so
 # the allowlist realpaths must point there too (see core_templates).
 _TEMPLATES_DIR = os.path.realpath(ensure_core_templates())
-DEFAULT_TIMEOUT = 30.0
+# 60s (not 30): a cold overview read of a large remote COG over the mount's HTTP
+# serve legitimately takes ~30-40s on first open (the pyramid analyze; the
+# template's own worker already allows 900s). 30s killed those mid-read. Still
+# well short of a runaway-guard — a genuinely hung script is caught, just later.
+DEFAULT_TIMEOUT = 60.0
 
 # Explicit allowlist of first-party helpers that run IN-PROCESS (D72): each is
 # bounded, self-contained, and never imports or executes user code — the
--- a/fused_render/executor.py
+++ b/fused_render/executor.py
@@ -182,6 +182,21 @@ def _run_python(path: str, params: dict, timeout: float) -> dict:
             capture_output=True,
             text=True,
             timeout=timeout,
+            # close_fds=False forces CPython to spawn via posix_spawn instead of
+            # fork()+exec (verified: the default close_fds=True takes the fork
+            # path on macOS/Linux). This is a native-crash fix, not an fd-policy
+            # choice. The server process has libproj resident with a live SQLite
+            # handle to proj.db in PROJ's SQLiteHandleCache (pyproj is pulled in
+            # transitively — e.g. via `fused`/geopandas). fork() runs every
+            # registered pthread_atfork *child* handler in the forked child
+            # before exec; PROJ's handler calls sqlite3_close/VFSClose on that
+            # inherited-but-now-invalid handle and segfaults (SIGSEGV) — so the
+            # child dies with code -11 before it can exec the worker, and the
+            # run surfaces as "worker exited with code -11 without producing a
+            # result". posix_spawn does NOT run atfork handlers, eliminating the
+            # crash path entirely. The worker is short-lived and inherits only
+            # the pipes it needs, so not closing inherited fds is harmless here.
+            close_fds=False,
             # a windowless server (Explorer-opener spawn) must not flash a
             # console window per run
             creationflags=subprocess.CREATE_NO_WINDOW if sys.platform == "win32" else 0,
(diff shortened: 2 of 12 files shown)


iamsdas:
All four Bugbot findings addressed in `7552932`, with regression tests in `8b6ea4f`.

| Finding | Resolution |
|---|---|
| `_is_file_mount_safe` accepts directories | Now `rc_kind_for(path) in ("file", "indeterminate")` — a confirmed `"dir"` 404s like `os.path.isfile` would; `"indeterminate"` still fails open. |
| recents `_file_path_from_url` accepts mount directories | `_mount_exists` switched from `rc_stat_for != "missing"` to `rc_kind_for in ("file", "indeterminate")` — files-only (D22), so a `"dir"` is filtered / not recorded. |
| Fused engine timeout left at 30s | `engine.py` now sources `timeout_seconds` from `executor.DEFAULT_TIMEOUT` (60s), so the two engines can't drift. |
| Block-cache memory bound inflated (64×4MiB = 256MiB) | `_HttpRangeFile` LRU is now byte-bounded (~32MiB): `max_blocks = max(4, 32MiB // block)`, scaling inversely with block size. |

New tests: parametrized `_is_file_mount_safe` gate (file/indeterminate pass, dir/missing 404) with a loud kernel-`isfile` guard; recents dir-filter + open-time dir-rejection; and the 4MiB default block, byte-bounded LRU derivation, and eviction honoring `max_blocks`. Full related suite: 496 passed, 3 skipped.

iamsdas:
## Follow-up: fixed a native SIGSEGV when opening large mount-backed COGs

While verifying cold pyramid renders across many files in a mount, switching to **larger** COGs (≥~150 MB) intermittently failed with `reader failed: worker exited with code -11` (SIGSEGV). Root-caused and fixed (`82d2e12`, `0d92b54`).

**Root cause — `fork()`-after-PROJ, not concurrency.** The server process has `libproj` resident with a live `proj.db` SQLite handle in PROJ's `SQLiteHandleCache` (`pyproj` pulled in transitively via `fused`/geopandas). When the executor spawns a worker via `subprocess.run(..., capture_output=True)`, CPython on macOS takes the `fork()+exec` path; the fork **child** runs PROJ's registered `pthread_atfork` child handler, which `sqlite3_close()`es the now-invalid inherited handle and segfaults **before exec** — so the worker never starts (`-11`). Timing-dependent (needs a live SQLite handle at fork), which is why it surfaced on large files under the shell's concurrent prefetch. 6/6 crash reports confirm the PROJ atfork frame.

**Fix (process internals only; templates stay mount-agnostic):** spawn workers via `posix_spawn` (runs no atfork handlers) instead of `fork()+exec`:
- `executor.py` `_child.py` spawn and `overview_pyramid.py` analyze/predict `subprocess.run` → `close_fds=False` (flips CPython to `posix_spawn`).
- `overview_pyramid.py` build/cogify → `os.posix_spawn(..., setsid=True)` with `/dev/null` file-actions (keeps job detachment).

Safe because Python fds are `O_CLOEXEC` (PEP 446), so they still close at exec.

**Verified:** B02 (158 MB) — the reliable crasher — now renders the full pyramid with concurrent prefetch running; **0 new PROJ-atfork crash reports** across 25+ analyzes; mount stays healthy. New `tests/test_worker_forksafe.py` (behavioral atfork guard) fails pre-fix / passes post-fix; `pytest` over pyramid + prefetch + fork-safe + mount-safety suites → **122 passed, 3 skipped**.

**Known separate issues (pre-existing, not addressed here):** a distinct worker-side `SIGABRT` (uncaught GDAL `std::length_error`) on some files under concurrent prefetch — surfaces as a clean "worker failed", does not wedge the mount; and TCI.tif (232 MB) occasionally hitting the 60s cold-read budget. Both are worth a follow-up.

PR #227, opened 2026-07-21
perf(fs): speed up conditions + folder-open navigation

iamsdas:
## Summary

- Opening a directory view fired `GET /api/fs/conditions`, which evaluated the `zarr_aoi` gate over the mount and took **~6.8s** — long enough that the mode switcher spun before the view was fully usable. On a non-zarr directory (the common case, e.g. an Ookla parquet partition on an anonymous-S3 mount) the gate hit its worst path: a redundant `is_dir` re-probe plus three serial marker `isfile` misses, ≈10 sequential S3 round trips.
- **No cache (#1):** re-navigating to the same directory now reuses a recent verdict via a short (60s) check-on-read TTL cache on the endpoint. Only success payloads are cached; `404`/error responses are always recomputed. Warm re-open drops from ~6.8s to instant.
- **Redundant probe removed (#2):** the endpoint already takes one `rc_kind_for` `is_dir` probe; that verdict is now threaded into the gate shim so the gate no longer re-probes the same path.
- **Marker probes consolidated (#3/#4):** for a direct-list-capable (anonymous S3/GCS) mount, one bounded unsigned listing of the directory's immediate children answers all three zarr marker `isfile` probes locally (the markers are always immediate children, so a *complete* page proves each present/absent). A truncated or failed listing transparently falls back to today's per-marker rc probes.
- Net cold cost on the reported ookla path: ≈**10 round trips → ≈3** (one `is_dir` probe + one listing), warm ≈**0**. All existing mount-safety invariants preserved: no kernel `os.*` ever touches a mount path, gates still fail closed, and the `404`-on-missing / `200`-on-indeterminate endpoint semantics are unchanged.

## Visuals

Gate evaluation for a mount-backed directory. The delta is the TTL cache + the `_GateSeed` (one is_dir probe and one bounded listing replacing the redundant re-probe and three serial marker probes).

**Before — ~6.8s, every call:**

```mermaid
flowchart TB
    A1[GET /api/fs/conditions] --> A2[rc is_dir probe]
    A2 --> A3["gate: os.path.isdir(path)<br/>REDUNDANT re-probe"]
    A3 --> A4["isfile .zmetadata (rc)"]
    A4 --> A5["isfile zarr.json (rc)"]
    A5 --> A6["isfile .zgroup (rc)"]
    A6 --> A7[verdict]
```

**After — ≈3 round trips cold, ≈0 warm:**

```mermaid
flowchart TB
    B1[GET /api/fs/conditions] --> B2{fresh in TTL cache?}
    B2 -->|hit| B9[cached verdict ~0ms]
    B2 -->|miss| B3[rc is_dir probe] --> B4[one bounded direct listing]
    B4 --> B5["_GateSeed: kinds + file_children"]
    B5 --> B6["gate: isdir -> seed.kinds (no probe)"]
    B6 --> B7["3x isfile -> seed.file_children (no probe)"]
    B7 --> B8[verdict] --> B10[store in cache]
    B4 -. truncated / not direct-capable .-> B11[fall back to per-marker rc probes]
    B11 --> B8
```

No UI change — the frontend already fetches this endpoint from a background effect and renders unconditional templates immediately; this PR only makes the verdict land faster.

## Usage

Not directly user-invocable — it's the existing `GET /api/fs/conditions?path=<dir>` endpoint, now faster. To observe:

```bash
# Cold (first open of a directory)
curl -s -o /dev/null -w '%{time_total}\n' \
  "http://127.0.0.1:1777/api/fs/conditions?path=<url-encoded-mount-dir>"
# Warm (within 60s) — served from the TTL cache, ~0s
curl -s -o /dev/null -w '%{time_total}\n' \
  "http://127.0.0.1:1777/api/fs/conditions?path=<same-dir>"
```

Tunable: `server._CONDITIONS_TTL_S` (default `60.0`) and `server._GATE_LIST_MAX_KEYS` (default `1000`).

## Test Plan

- [x] **Automated — new `tests/test_conditions_cache.py`** (4): cache hit within TTL, distinct paths cached separately, stale-TTL recompute, error/404 responses not cached.
- [x] **Automated — new `tests/test_condition_gate_seed.py`** (6): seed skips target re-probe; complete listing with no markers → False with no marker probes; complete listing with `.zgroup` → True; truncated listing falls back to probes; listing failure falls back to probes; v3 group via listing (`zarr.json` from listing, `node_type` read still runs). All under the `guard_kernel` fixture that fails loudly on any kernel `os.*` against a mount path.
- [x] **Regression** — existing `tests/test_condition_mount_shim.py`, `tests/test_zarr_aoi_mount.py`, `tests/test_templates.py` pass unchanged (all `seed` params default to `None`; behavior is byte-for-byte identical when unseeded). Added a shared cache-clearing fixture to the two endpoint test files so the module-level cache doesn't leak verdicts across tests.
- [x] **Full suite** — `python -m pytest -o addopts="" -q` passes (exit 0).
- [ ] **Manual** — open a mount-backed directory in the app; confirm the mode switcher resolves quickly, re-opening the same dir is instant, and a real `.zarr` store still offers the AOI streamer.

---

## Follow-up: the folder-open navigation pause

With `conditions` no longer the bottleneck, opening a folder still showed a **~1s blank screen** before the template list + a "loading" listing appeared. A network waterfall pinned it: **`/api/fs/stat` (~1.6s cold)** sat on the critical path — the frontend rendered nothing until it resolved — and `/api/fs/list` + `/api/fs/conditions` were **serialized behind it** rather than overlapping. `stat` is slow because a mount stat LISTs the parent prefix over rclone/S3 just to describe one child.

**Three fixes (backend + frontend):**

- **BE — `/api/fs/stat` TTL cache:** a 60s check-on-read cache (mirrors the conditions cache) so re-navigation / the stat that follows a listing is instant. Only **mount-backed success** payloads are cached (local stats are cheap and can change out-of-band); every mutation route (`write`/`mkdir`/`delete`/`rename`/`copy`) invalidates the touched paths + parents, so the editor's post-write stat always re-reads fresh.
- **FE #1 — don't gate first paint on `stat`:** navigation now paints a scaffold immediately (breadcrumb + folder name + a template-strip spinner + the real listing) instead of blanking. A directory hint rides in `history.state` so the destination knows to show a listing before `stat` resolves.
- **FE #2 — parallelize:** the scaffold's listing fires `/api/fs/list` in parallel with the in-flight `stat` (a 5s `prefetchListDir` dedupe means the remount into the resolved preview reuses that same request). `conditions` stays non-blocking as before.

**Net effect:** the blank screen is gone (instant scaffold); the listing renders as soon as `list` returns (~800ms) instead of after `stat`+`list` in series; warm re-navigation is instant (stat cache).

```mermaid
flowchart TB
    subgraph B["Before — serialized behind stat"]
      direction TB
      N1[click folder] --> N2["BLANK screen"]
      N2 --> N3["stat ~1.6s (blocks paint)"]
      N3 --> N4["then list ~0.8s → 'loading'"]
      N4 --> N5[then conditions]
      N5 --> N6[usable]
    end
```

```mermaid
flowchart TB
    subgraph A["After — scaffold + parallel"]
      direction TB
      M1[click folder] --> M2["scaffold paints NOW<br/>breadcrumb + listing skeleton"]
      M2 --> M3["stat, list, conditions in parallel"]
      M3 --> M4["list ~0.8s → rows render"]
      M3 --> M5["stat fills templates/metadata<br/>(cached & instant when warm)"]
      M3 --> M6[conditions gates AOI spinner]
    end
```

### Test Plan (follow-up)

- [x] **Automated — new `tests/test_stat_cache.py`** (7): cache hit within TTL, distinct paths, stale-TTL recompute, errors not cached, non-mount paths not cached, write invalidates, rename invalidates src+dst.
- [x] **Regression** — `tests/test_condition_mount_shim.py` fixture extended to clear `_STAT_CACHE` between cases; full suite **1336 passed, 4 skipped**.
- [x] **Frontend** — `tsc --noEmit` clean, `vite build` succeeds (no FE unit-test harness exists in this repo).
- [ ] **Manual** — open a mount-backed folder: confirm no blank screen (instant scaffold), listing appears without waiting on stat, re-opening a sibling is instant, and a wrong-hint edge (dir that's actually a file) resolves cleanly into the file preview.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -7,7 +7,7 @@
 // which is the React equivalent of the vanilla shell rebuilding the view DOM
 // on each route() call (fresh iframes, fresh fetches, dropped local state).
 import { useEffect, useState } from "react";
-import { IS_EMBED, fsPathFromLocation, urlForFsPath } from "./lib/router";
+import { IS_EMBED, fsPathFromLocation, urlForFsPath, navHintIsDir } from "./lib/router";
 import { useSessionRestore, useSessionTracking } from "./lib/session";
 import { useRecentsTracking } from "./lib/recents";
 import { statPath, getMounts, reconnectMount, type Config, type Mount, type StatResult } from "./lib/api";
@@ -134,12 +134,50 @@ function StatErrorView({
   );
 }
 
+// First paint while `stat` is still in flight (~1.6s on a cold remote mount),
+// so a navigation shows a populated scaffold instead of a blank screen. The
+// breadcrumb is already rendered by StatView; here the preview header shows the
+// folder/file name (from the URL) with a spinner where the template
+// ModeSwitcher will land once stat resolves. When the nav hint says this is a
+// directory, the real Listing mounts NOW — its /api/fs/list runs in parallel
+// with stat rather than serialized behind it, and the same fetch is reused
+// (api.prefetchListDir) when stat resolves and the preview remounts the
+// listing. Without a directory hint we can't safely show a listing (a file's
+// list would 404), so only the header + a neutral loading body paint.
+function LoadingScaffold({ fsPath, isDir }: { fsPath: string; isDir: boolean }) {
+  return (
+    <>
+      <div className="preview-header">
+        <h1 title={fsPath}>{basename(fsPath)}</h1>
+        <div className="preview-actions">
+          <span className="mode-icon-spinner" aria-label="Loading" />
+        </div>
+      </div>
+      <div className="preview-body">
+        {isDir ? (
+          <Listing fsPath={fsPath} />
+        ) : (
+          <div className="preview-resolving">
+            <span className="mode-icon-spinner" />
+            Loading…
+          </div>
+        )}
+      </div>
+    </>
+  );
+}
+
 // Stat-backed views (listing/preview): breadcrumb + content under one hook
 // component so useStat only runs when the pathname is a real fs path, not a
 // sentinel.
 function StatView({ fsPath, epoch, home }: { fsPath: string; epoch: number; home: string }) {
   // Bumped by StatErrorView to re-stat in place after reconnecting a mount.
   const [reloadKey, setReloadKey] = useState(0);
+  // Directory hint from the navigation that mounted this view (see router
+  // navHintIsDir). Captured ONCE at mount — StatView is keyed by epoch+fsPath
+  // so it remounts per navigation, and reading it live would be clobbered by
+  // the listing's own sort/search replaceState (which pass null state).
+  const [navIsDir] = useState<boolean | null>(() => navHintIsDir());
   const stat = useStat(fsPath, epoch, reloadKey);
   // null until the stat resolves — the session hooks opt out for anything that
   // is not a confirmed file, so a directory never gets a restore/track before
@@ -164,7 +202,11 @@ function StatView({ fsPath, epoch, home }: { fsPath: string; epoch: number; home
   // gate as session tracking.
   useRecentsTracking(fsPath, isDir, renderedTitle);
   let content = null;
-  if (stat.status === "error") {
+  if (stat.status === "loading") {
+    // Not a blank screen: paint the scaffold immediately (Fix #1). A directory
+    // nav also starts its listing fetch now, parallel with stat (Fix #2).
+    content = <LoadingScaffold fsPath={fsPath} isDir={navIsDir === true} />;
+  } else if (stat.status === "error") {
     content = (
       <StatErrorView
         fsPath={fsPath}
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -175,8 +175,10 @@ function StatView({ fsPath, epoch, home }: { fsPath: string; epoch: number; home
   const [reloadKey, setReloadKey] = useState(0);
   // Directory hint from the navigation that mounted this view (see router
   // navHintIsDir). Captured ONCE at mount — StatView is keyed by epoch+fsPath
-  // so it remounts per navigation, and reading it live would be clobbered by
-  // the listing's own sort/search replaceState (which pass null state).
+  // so it remounts per navigation. In-place param syncs go through
+  // router.replaceSearch, which preserves history.state, so the hint survives
+  // for Back/Forward; capturing once here is belt-and-braces (and correct even
+  // if some future caller forgets to preserve it).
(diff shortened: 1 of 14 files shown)


iamsdas:
Addressed both Bugbot findings + one self-review item in 4a5b174:

- **Indeterminate kind blocks gate retry** (medium): now only *definitive* `dir`/`file` kinds are seeded. An `indeterminate` probe leaves `seed=None` so the gate does its own probe (can recover; otherwise fails closed on its own budget) instead of being pinned False for the cache TTL.
- **Truncated listing ignores known markers** (low): `_isfile` now returns True for a marker *present* in the page regardless of truncation (presence is conclusive); only a truncated-*and-absent* marker falls through to the rc probe.
- **Self-review**: the bounded listing now runs only after `gated` is known and non-empty (no wasted round trip on dirs with no conditional template), and its `except` logs at debug instead of silently swallowing.

New regression tests in `tests/test_condition_gate_seed.py`; full suite green (1329 passed, 4 skipped).

iamsdas:
Addressed both new Bugbot findings on the follow-up commits (`360bf78`):

- **Stale stat can refill cache** (medium, `server.py`): added a monotonic `_STAT_CACHE_GEN`. `api_fs_stat` snapshots the generation *before* the slow `_fs_stat` and only writes to `_STAT_CACHE` if it's unchanged; `_invalidate_stat_cache` bumps it unconditionally. A mutation that completes while a stat is in flight is now observed and blocks the pre-mutation payload from refilling the cache. New deterministic test `test_midflight_invalidation_is_not_refilled` (invalidates *during* a stubbed `_fs_stat`, asserts no refill). Known micro-residual: the gen-check and dict-write aren't lock-wrapped, so a switch in that ~few-bytecode, I/O-free window is theoretically possible — self-heals within the 60s TTL, and the server's mtime compare on write catches real clobbers as a 409.
- **Nav hint cleared on replaceState** (medium, `router.ts`): added `replaceSearch(url)` which carries `history.state` through instead of nulling it; routed every in-place param sync (Listing sort/search/`setSort`, Preview `_mode`, session replay) through it. The `{ fsDir }` hint now survives Back/Forward. (The `/`→start_dir root redirect stays `replaceState(null)` — no hint applies there.)

Also fixed the wrong-hint flash risk: a `provisional` `<Listing>` in the loading scaffold suppresses the "Failed to list" error while stat resolves (committed render's error path unchanged).

Full suite: **1337 passed, 4 skipped**; frontend `tsc --noEmit` clean, `vite build` succeeds.

PR #233, opened 2026-07-22
fix(claude-template): survive reloads, keep full streamed turn

AkshilVT:
Fixes the three Claude Code integration bugs reported by Maxime + Akshil:

1. **Reply stops streaming mid-sentence / conversation disappears after an edit** — the chat template never opted out of LR-1 auto-reload, so Claude's own edit to the watched `_file` reloaded the chat iframe mid-stream: poll loop dead, run orphaned, history restore showing whatever the transcript had at that instant. The template now calls `fused.autoReload(false)` (same posture as `code`/`slides`); a sibling `_render` pane still live-refreshes.

2. **Frozen mid-word replies** — `_poll` swapped the final text to the `result` row, which holds only the **last** assistant message of a tool-using turn. The swap dropped all pre-tool text, and the shorter target left the typewriter's `shown > target` with nothing to redraw — the reply froze mid-word. Final text is now the joined stream deltas (`result` only as fallback when nothing streamed), consecutive assistant messages get a blank-line separator, and the typer clamps + redraws if the target ever shrinks.

3. **Prompt + mode switch loses the response** — a mode switch remounts the iframe, killing the poll loop while the detached subprocess keeps working; returning showed only the turn's user message, forever. The in-flight run id now rides the URL as the `run` param (set on start, cleared on done): boot re-attaches, trims any partial rows history already rendered for that turn, and resumes streaming. `poll` returns the run's first `message` so the user turn renders even when the frame died before any transcript rows existed. Stale `run` params (bookmarks, pruned tmp) are cleared silently.

## Testing

- Playwright edge suite, 13/13: mid-run mode switch → re-attach + full reply + edit on disk; return after run finished while away; reload <1s after send (no session id yet); follow-up resume turn; garbage `run` param (no error bubble); bogus model (error bubble); back → session list → history restore; double-Enter + mid-run send gated to one turn
- Headless poll-trace: no text shrink, both messages kept + separated; `weird dir (v2)/my page (final).html` path through start/poll/sessions/history
- `pytest -n auto`: 1239 passed (2 pre-existing local `test_deploy` failures unrelated)

---

> [!NOTE]
> **Medium Risk**
> Changes are confined to the Claude template and its poll/streaming contract; behavior is user-visible but does not touch auth, deploy, or core server APIs.
> 
> **Overview**
> The **Claude** chat template now calls **`fused.autoReload(false)`** so edits to the watched `_file` no longer reload the chat iframe mid-reply (sibling `_render` panes still live-refresh per LR-5).
> 
> **`agent.py` `poll`** keeps the **full turn** by using accumulated stream deltas instead of swapping to the `result` row (which only held the last assistant message after tool use), inserts blank lines between consecutive assistant messages, and returns the run’s first **`message`** for UI re-attach. Sidecar recording reads **`meta.json`** once and tolerates missing fields.
> 
> The **chat UI** stores the in-flight **`run`** param from start until done; **`resumeRun`** re-attaches after reload or mode switch (trims duplicate partial rows, handles finished-while-away and stale `run` ids). The typewriter **clamps** when streamed text shrinks. Boot enters chat when only **`run`** is set.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7388d93fd025525c85ec1f10be47eee7b10b896. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AkshilVT:
bugbot run

AkshilVT:
bugbot run

PR #302, opened 2026-07-28
markdown v2: one editable Live Preview surface, wikilinks, backlinks and a link graph

iamsdas:
## Summary

**`.md` becomes a place you write, not just read.** The `markdown` template is a single always-editable CodeMirror Live Preview surface: markup renders in place and reveals as source only on the caret's line, as Obsidian does. No source/reading switch and no toolbar — with one surface there is nothing to toggle to, and the shell's breadcrumb already names the file. Saving copies Obsidian's model: 2 s idle autosave plus blur/tab-switch, no save button and no dirty dot, `⌘S` only flushes early. Read-only is available (`?edit=0`) as CodeMirror facets, so nothing about the rendering changes.

**Wikilinks, backlinks and a link graph.** `[[Wiki Link|label]]`, `![[embeds]]` and `#tags` resolve against a sqlite-cached per-note index. A resizable side panel shows backlinks over a force-directed graph of the neighbourhood, and a folder-level `graph` mode shares the same canvas and the same index.

A link has **three** states, not two: resolved is a link, looked-for-and-missing is a dashed ghost that creates the note on click, and *never scanned* — a remote mount, a refused or failed walk — is inert. Asserting "this doesn't exist" where we never looked is a lie, and it would offer to create a note that may be sitting right next to this one. `[[#Heading]]` is none of the three: it points inside the open note, so it is a plain link that scrolls.

**The walk is bounded, and says when it stopped.** It runs on every `.md` open, so it is capped on notes, assets, and — the one that actually bounds cost — total directory entries *enumerated*. Vendored and build-output trees are skipped, which both stops drawing someone else's generated README as a node the author appears to have written and stops paying to walk it. Every cap that fires sets `truncated`, and all three surfaces render a notice from it: silently showing a subset is the one outcome forbidden.

**Remote mounts are safe by construction.** The view works over a mount; the graph refuses, in four layers — a `condition.py` gate that never enumerates, a vault-root climb that stops before a mount and probes rather than lists, a server-side refusal, and a client-side check. A recursive walk of a flat million-key S3 prefix is how this repo has previously killed a kernel NFS mount, and every layer fails closed: if mount detection is unavailable, that reads as "refuse".

**Obsidian's default neutral palette**, both themes, with the editing surface the same colour as the page — that unified surface is most of why Obsidian reads calm.

### Bugs found and fixed under review

Nine findings from `/code-review` (one raised independently by Bugbot), each with a regression test:

| Severity | Bug |
|---|---|
| **Data loss** | A ghost click wrote its target unconditionally, so clicking a ghost whose target *existed* replaced that note with a one-line stub |
| **High** | The entry budget was only tested inside the per-file loop, so a directory-only subtree (30,300 empty dirs) was walked in full and still reported `truncated=False` |
| Medium | Dragging a graph node also opened it — the browser dispatches `click` after any press+release on one element |
| Medium | `setData` re-seeded every node position, so the layout re-exploded on every 2 s autosave |
| Low ×3 | `[[#Heading]]` rendered as a dead create-me ghost; `reloadFromDisk` read `stat.writable` and never applied it; `resolve_link`'s documented fourth fallback was unreachable dead code |

The data-loss one is worth reading twice, because it was reachable in ordinary use rather than at an edge: `resolved` holds only the *last* scan's answers, so a `[[Link]]` you just typed is a ghost until the scan lands — and an **ambiguous** target is a ghost *by design*, which is exactly the case where a real file sits at the computed path. Both surfaces now pass `create: true` and let the server decide: absent → created, present → navigate to it, anything else reported rather than swallowed as "absent". `create` already existed in `/api/fs/write`; this plumbs it through `runtime.js` with its own `"exists"` error type, kept distinct from the optimistic lock's `"conflict"` because a caller may offer overwrite-anyway on one and must never offer it on the other. There is no stat-then-write left, so no TOCTOU window.

## Visuals

The delta: one decoration set over the source document, replacing the previous "render a second copy with `marked`" pipeline.

```mermaid
flowchart TD
    A[".md on disk"] --> B["EditorState"]
    B --> C["buildDecorations"]
    C --> D["decoration set"]
    D --> E["renders in place;<br/>caret's line shows source"]
    B --> F["idle autosave"]
    F --> A
    A --> G["graph.py: bounded walk,<br/>cached parses"]
    G --> H["backlinks + graph panel"]
```

`buildDecorations` is provided by a **`StateField`, not a `ViewPlugin`** — CodeMirror rejects a plugin-provided replacement that spans a line break, and the table widget replaces a multi-line `Table` node. As a plugin it threw during the view update the moment a table scrolled into view, abandoning the rest of the render and leaving blank regions on screen. The comment saying so is load-bearing.

## Usage

Open any `.md` file. It is editable immediately — type, and it saves itself.

| Action | How |
|---|---|
| Reveal a line's markup | Put the caret on it |
| Follow a wikilink / relative link | Click it |
| Jump to a heading in this note | Click a `[[#Heading]]` link |
| Create a missing note | Click its dashed ghost, in the text or in the graph |
| Backlinks + graph | The panel button, top right; drag its edge to resize |
| Reach the whole vault | Depth `all` in the panel |
| Move a node / pan / zoom | Drag it / drag the background / wheel |
| Read only | `?edit=0`, or the mode button |
| Force a save | `⌘S` (it had already saved) |
| Folder-level graph | Open a folder containing `index.md`, pick the `graph` view |

The folder graph is *offered* on `index.md` alone, never on `README.md`: a README says "this is a project", not "this is a vault", and probing it would put the mode on every checkout on the disk. That is discoverability only — an explicit `?_mode=graph` still works anywhere the mount rules allow.

## Test Plan

- [x] **Automated by execution, not source-reading.** `tests/test_markdown_live_preview.py` drives the real decoration builder against the real vendored grammar via `scripts/vendor-codemirror/live-preview-probe.mjs`. This caught two bugs no source assertion could: `---\ntitle: x\n---` parses as `HorizontalRule` + `SetextHeading2` (the grammar has no frontmatter rule), and the inner brackets of `[[Wiki]]` parse as a `Link` node — so a code-guard listing `Link` silently rendered *no* wikilinks at all.
- [x] **`tests/test_graph_canvas.py` (new)** loads the shipping `graph-canvas.js` under node with a stubbed canvas and dispatches real mouse events. Both canvas bugs above are pure interaction — a click that should not have happened, a layout that restarts under you — and no source assertion catches either.
- [x] `tests/test_markdown_template.py`, `tests/test_markdown_graph.py`, `tests/test_graph_condition.py` — template contract, resolver/index/graph payloads, and the mount gate. The mount tests assert the *absence of the syscall*, not the correctness of the answer: a probe that reaches a live NFS mount is the failure, so all three probe primitives are counted and directory enumeration is fatal.
- [x] Full suite: **2302 passed, 69 skipped, 1 failed**. The failure is `tests/test_calls.py::test_an_abandoned_merge_closes_the_day_s_files`, which reads `/proc/self/fd` and so fails on macOS on `main` too — pre-existing, untouched, missing a skip guard.
- [x] Manual, against a 205-note vault: a 239-line note with a table scrolls with no blank regions and no console errors; the mode toggle preserves caret and scroll; panel resize persists; selecting text in read-only mode does not reveal markers; dragging a node no longer opens it while a plain click still does; typing (so autosaving, so re-sending the graph) leaves every node position identical; and all three ghost-click outcomes — absent → created, present → opened with its bytes intact, a directory at the path → reported with nothing written.
- [x] Scale: 30,300 empty directories now stop at the entry budget and report it. A real repo is unaffected — 205 notes, 878 assets, `truncated=False`, 254 ms.
- [ ] **For the reviewer:** open a note on a remote mount and confirm the view works while the graph declines rather than walking the mount.

## Deviations

- **Depth `all` is a minimap, not a readable graph.** At 205 notes the fitted view is unlabelled dots: labels fade below 0.7 zoom by design, and fitting a vault that size puts it well under. `all` exists so the rest of the vault is *reachable* — before it the panel could only ever show a BFS ring around one note. Depth `1`–`3` is what you read; `all` is what you navigate. Clustering or level-of-detail labelling would be the real fix and is not attempted.
- **`.gitignore` is deliberately not consulted** by the walk; `SKIP_DIRS` is a fixed name list. Pattern matching with negations, `**` and nested ignore files is real per-file cost on a path that runs on every note open, and it silently does nothing outside a git repo — which is where a vault often lives (Obsidian likewise keeps its own list). `server._walk_bfs` can afford git's own answer via a `check-ignore` co-process; this cannot. The cost: a generated `.md` in an unusually-named output directory still appears as a node. The list stays conservative for the opposite reason — a skipped directory is invisible with *no* notice, so a folder of real notes silently missing is the worse failure.
- **The accent is not Obsidian's.** Obsidian's default violet `hsl(254, 80%, 68%)` read as off against these neutrals, so the accent stays fused-render's yellow; the neutral scale is Obsidian's verbatim. One variable to flip, which the canvas reads at draw time. Related: light `--warn` is darkened from Obsidian's l=46% to 36% (2.8:1 → 4.3:1 on `--bg-alt`), hue and saturation kept; and `oneDark` survives for syntax inside fenced code only, its surface and prose colours overridden — restyling the fence markers needs the vendored bundle rebuilt, whose blast radius covers `code` and `annotate`.
- **The layout is a hand-rolled O(n²) spring sim**, not a vendored force library. Fine at a folder's node count and deliberately not a general-purpose graph renderer; a vault an order of magnitude larger would need one.
- **Two save models for `.md` now coexist and are accepted, not reconciled:** `code` keeps its 250 ms + explicit Save (AS-1), `markdown` uses Obsidian's (MD-16). SPEC §32 records why.


iamsdas:
Both Bugbot findings were real Windows breakage, not style — fixed in 2643289.

**Ghost create breaks on Windows** (high) — confirmed. `resolvePath` prefixed a leading `/` unconditionally, so a drive path became `/C:/…`, which then failed `createGhost`'s scan-root boundary check and refused *every* create on Windows. Guarded in both surfaces that resolve a path, using the same `/^[A-Za-z]:$/` idiom the `docs`, `excel` and `slides` templates already carry — so this was a divergence from established prior art in this repo, not a new decision.

**Native separators in graph paths** (medium) — confirmed, and the `.replace("/", os.sep)` alongside each `os.path.join` made it deliberate rather than accidental: the payload was being converted *to* native separators for a consumer that splits on `/`. Client-facing `path` and `root` values now go through `_client_path`/`_client_join`, converted once at the boundary; the filesystem side above keeps native separators, which is correct there.

Tests drive the Windows string shapes directly instead of through a walk, so the case is exercised on every platform. Suite: 2258 passed, 69 skipped, plus the pre-existing macOS `/proc/self/fd` failure that also fails on `main`.

iamsdas:
Five commits since the review fixes — two are the side panel, three came out of a scalability review of the walk.

### Side panel

- **`962f1b8`** — the space split was inverted: `#links` was `flex: 1` and `#graph-sec` was capped at `height: 45%`, so 4 backlinks reserved ~374px (≈262px of it empty) while the graph was squeezed and its labels collided. Now `#links` sizes to content with a `max-height: 42%` cap and the graph takes the remainder. Measured on a 738px panel: links 374→112px, graph 332→594px. A 30-row list still caps and scrolls rather than starving the graph.
- **`6cea471`** — backlink rows read as a cramped table of cards. `.bl-path` was `flex: 1 1 auto; text-align: right`, which only works in a fixed-width column; in a resizable panel it put a differently-sized hole in every row, and pushed the path (the only thing telling two same-titled notes apart) as far as possible from its title. Rows are now borderless with the path inline, hover is a fill, and `:focus-visible` is an explicit accent outline — previously that ring was the UA default, so removing the border would have silently left keyboard users with no focus affordance. 29→24px per row; only a row that genuinely doesn't fit wraps, capped at two lines.

### The walk

- **`a0b9c68`** — `MAX_FILES`/`MAX_ASSETS` bounded what the walk *recorded*, not what it *enumerated*. `truncated` could only be set from the note branch, so once 5,000 assets were recorded the walk kept enumerating the whole remaining tree while recording nothing — and reported nothing, because by its own accounting nothing was dropped. Measured on 20,000 non-note files beside 1 note: `truncated=False`, all 20,002 entries read. There is now a `MAX_ENTRIES = 20000` budget counting every entry visited, and all three caps set `truncated`. A 60k-file tree goes 107.6ms → 35.4ms and now says so. `openfused` is unaffected: 205 notes, `truncated=False`, unchanged.

  This matters because the walk runs on **every** `.md` open, and warm opens are stat-only — so it was the steady-state cost of opening a note.

- **`48b27d7`** — `.gitignore` is not honored (verified: adding one changed the walk not at all), and `SKIP_DIRS` only covered dotdirs plus `node_modules`/`__pycache__`/`site-packages`/`venv`. So `build/`, `dist/`, `target/`, `_build/` were walked and any `.md` inside them was indexed as a real note and drawn as a graph node. Added the common output/vendor names; deliberately **not** `docs`, `site`, `public`, `static`, `content`, `notes`, `src`, `bin`, `env`. The asymmetry is the point: a skipped directory produces no notice at all, so wrongly skipping real notes fails silently while wrongly including a build tree is merely visible noise.

  Real gitignore parsing was considered and rejected — negations, `**` and nested ignore files are real per-file cost on this path, and it does nothing outside a git repo, where a vault often lives. Obsidian uses its own exclusion list rather than gitignore. The comment records this so it isn't "fixed" later as an oversight.

- **`37dae5a`** — test only. The ascent already refused to climb into a mount, and the note-under-a-mounted-ancestor case was covered; the note *itself* on a mount was not, and there the 8-level bound is no help — 8 `isdir`/`isfile` probes against a remote is 8 kernel GETATTRs on a live NFS mount. Asserts the absence of the syscall (all three probe primitives counted) rather than the correctness of the answer.

### Notes for the reviewer

Three notice strings changed wording, which is a behaviour change worth a look: "Only the first N notes were scanned" was false once the asset or entry cap could fire with every note indexed, so the strings now say a cap fired without claiming which.

Suite: **2286 passed, 69 skipped**. The one failure is still the pre-existing `tests/test_calls.py::test_an_abandoned_merge_closes_the_day_s_files`, which reads `/proc/self/fd` and fails on macOS on `main` too.

Two open judgement calls from earlier are still unresolved and are yours to rule on: whether the accent should switch from fused's yellow to Obsidian's violet, and whether the `graph-canvas.js` changes that went beyond the original brief (fit-to-canvas framing, disc seeding, speed ceiling — a file the folder graph also uses) are acceptable.

iamsdas:
Three more commits: an `all` depth for the note panel's graph, a narrower gate for the folder-level graph mode, and a count-label fix that the first of those exposed.

### `424cb61` — an `all` depth in the note panel

The panel could only ever show a BFS neighbourhood at depth 1-3, so a 205-note vault rendered as "13 of 205 notes" with no way to reach the rest. `all` skips the neighbourhood filter while keeping the focus note.

Expressed as a **negative** depth rather than reusing `0`, because `0` was not free in either direction: with a focus, `_neighbourhood` with `range(max(0, 0))` already keeps only `{focus}` — a reachable URL state — and the folder graph relies on `depth: "0"` with no `file` meaning "no focus, nothing to filter by". A negative value is a region of the input no existing caller could reach, so nothing else changes meaning. `_coerce_depth` still falls back on garbage. `_graph_payload`'s docstring now names all three regions; SPEC MD-20 records the sentinel; a test pins that the folder graph's `depth: "0"` path is unchanged.

**Honest limitation, measured rather than assumed.** At 335x649 the fitted whole-vault view drives zoom to ~0.2, and labels only draw above 0.7 — so `all` is 221 unlabelled dots. You can read the shape (one dominant hub, a long tail of orphans) and identify nothing. It is navigable — wheel in and labels return, the focus dot is unmistakable — so it works as a minimap you drill into, which is what a global graph is for, but it is not a readable whole-vault picture. Perf is a non-issue: 7ms warm server-side, 64KB payload, 156ms end-to-end.

Two pre-existing canvas issues it makes more visible, neither touched here: labels overlap heavily in the dense core at high zoom, and tags are drawn in the same accent colour as the focus note, so at `all` the focus is one of ~10 yellow dots distinguished only by radius.

### `0fe9154` — the folder-graph gate now means something

`condition.py` probed a 10-name list including `README.md`, which is present in essentially every code repository, so the folder-level `graph` mode was offered on ordinary repos where a link graph says nothing. The probe set is now exactly `index.md` / `Index.md` — both cases because only a case-insensitive filesystem answers one for the other, and the probe is free.

Everything about *how* the gate works is unchanged and still enforced by `tests/test_graph_condition.py`: mount refusal before any probe, a fixed set of `os.path.isfile` calls and never a listing (the gate runs per browsed directory, and a listing is proportional to directory size — this is how a kernel NFS mount has previously been wedged here), fail-closed on any exception. Tests cover both directions: `index.md` offered, `README.md` not — the latter is the point.

Narrowing costs discoverability, not capability: `?_mode=graph` still reaches the mode by hand, verified. Recorded in SPEC MD-21 and DECISIONS D156, which names the old list and why it was wrong.

### `ac6da64` — count notes against notes

`graph-count` printed `data.nodes.length` against `total_notes`, but `nodes` also carries tag and ghost nodes. It overcounted at every depth — "13 of 205" was really 11 notes plus two tags — and `all` turned it into **"221 of 205 notes"**. Now filtered by `kind === "note"`. The folder graph's own label was already correct (it compares nodes with nodes and prints the note total separately), so only the panel needed it.

### Reviewer notes

`/Users/iamsdas/Work/openfused` no longer offers the folder graph — it has `README.md` and no `index.md`. That is the intended effect of `0fe9154`, but it means folder-mode testing needs `?_mode=graph` unless an `index.md` is added.

Suite: **2291 passed, 69 skipped**, plus the pre-existing `tests/test_calls.py::test_an_abandoned_merge_closes_the_day_s_files` (`/proc/self/fd`, fails on macOS on `main` too).

Still open for a human call: the yellow-vs-violet accent, and the `graph-canvas.js` changes that went beyond the original brief (fit-to-canvas framing, disc seeding, speed ceiling) in a file the folder graph shares.

PR #305, opened 2026-07-28
Bundle the AI proxy and manage AI accounts from Preferences

vasu2001:
Stacked on #304 (`fused-ai-api`) — review that one first.

#304 shipped `fused.ai()` as a relay to an OpenAI-compatible proxy the user
installs and runs themselves. That works, but it makes the feature invisible
until someone has independently discovered CLIProxyAPI, installed it, run an
OAuth login per provider, and left it running. This branch removes that
prerequisite: the proxy ships inside the app, the app supervises it, and
connecting a Claude or ChatGPT account becomes a button in Preferences.

## Approach

CLIProxyAPI is a single static Go binary (MIT), which is the same shape as
rclone — already bundled for mounts. So the bundling and lifecycle work is a
copy of an established in-repo pattern rather than a new one: pinned version
downloaded and checksum-verified at build time, staged into the app payload,
resolved at runtime through an env override → bundled path → PATH fallback
chain, spawned lazily with a generated config on a random port, health-polled,
and torn down with the same only-kill-a-pid-we-can-prove-is-ours safety gates
`shell/mounts.py` uses for the rclone daemon.

Account management uses the proxy's Management API, whose contract is not
documented upstream (`docs/MANAGEMENT_API.md` 404s), so it was established by
probing a real v7.2.90 binary — see `docs/AI_PROXY_MANAGEMENT_API.md`. The
important result is that OAuth can be driven entirely over HTTP, so Preferences
can offer a real "Connect account" flow without shelling out to CLI login
subcommands.

Scope of the Preferences UI is deliberately Claude and ChatGPT only for now,
even though the proxy also speaks Gemini, Kimi, xAI and Antigravity.

A user who already runs their own proxy keeps working unchanged — the PATH
fallback and the existing `FUSED_RENDER_AI_BASE_URL` / `ai_base_url` overrides
still point at it, so nothing about #304's contract changes.

---

> [!NOTE]
> **High Risk**
> Introduces local credential storage, OAuth subscription login flows, and authenticated relay to a third-party proxy; misconfiguration or teardown bugs could leak or orphan secrets/processes, and the design doc flags provider ToS exposure.
> 
> **Overview**
> **Bundles CLIProxyAPI** (pinned in DMG/AppImage/Windows/dev scripts) and supervises it from `shell/ai_proxy.py` on the same model as rclone: lazy spawn on loopback with generated API/management secrets, health checks, config preservation for API keys, and ownership-checked teardown on app/server quit.
> 
> **`/api/ai` relay** now uses the supervised proxy (Bearer auth) when no explicit `ai_base_url` / `FUSED_RENDER_AI_BASE_URL` override exists; otherwise behavior stays as before, including dev checkouts without a bundled binary.
> 
> **New `ai_accounts` router and Preferences “AI accounts” tab** (`?tab=ai`): list proxy status, OAuth connect via fixed localhost callback ports + polling `get-auth-status`, disconnect, masked API key add/remove with Codex read-back verification, and **round-robin** vs **fill-first** routing (persists in prefs and restarts the proxy).
> 
> **SPEC RH-12 / PF-10–14** and design docs (`AI_PROXY_BUNDLING.md`, `AI_PROXY_MANAGEMENT_API.md`) document the contract; supervisor sets `FUSED_RENDER_AI_PROXY_BIN`; `dev.sh` enables `FUSED_RENDER_AI_PROXY_PERSIST` and optional one-time binary staging.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 327b82bf907572742883497498a52afb86a31f8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1213,6 +1213,28 @@ never imports server).
   exchanged it and a failed exchange is only visible afterwards. One login at a time (the
   callback ports are fixed, so this is structural, not a policy); an in-flight attempt is
   cancellable, and an attempt started elsewhere is reported rather than silently blocking.
+  A **settled** attempt is neither: it stays readable through the status route so its
+  outcome isn't lost, but it must not read as in-flight — doing so left every Connect
+  button disabled after the first login, making the second provider unconnectable.
+  **Both providers can be connected at once**; they are independent credentials.
+- **PF-13** A provider can equally be authenticated with an **API key**, pasted in the same
+  tab. Keys are config-level credentials, not files in the proxy's auth dir, so they are
+  listed and removed separately from OAuth accounts (a key has no session to expire and no
+  browser to reconnect through). A full key **never leaves the server**: responses carry
+  only a short masked hint, and every mutation is keyed on the server-assigned
+  `auth_index`, so a key never appears in a URL, a log, or shell history. Writes are
+  **read back and verified** rather than trusted — the proxy accepts a Codex key with no
+  `base-url` and then silently discards it, so a write that doesn't survive the re-read is
+  reported as a failure instead of a phantom success.
+- **PF-14** How multiple credentials for one provider are used is a **visible setting**,
+  written explicitly into the proxy's config rather than inherited by omission:
+  `round-robin` (the upstream default) pools OAuth logins and API keys as **peers** and
+  rotates across them with failover on rate limits; `fill-first` drains one before moving
+  to the next, which is what "prefer this credential, fall back to that one" actually
+  requires. Note there is no built-in OAuth-over-key precedence to inherit — the upstream
+  treats them as equals — hence exposing the choice rather than asserting one. The proxy
+  reads config only at spawn, so changing this **restarts** it (respawning lazily on next
+  use).
 
 ### 20.6 Template registry view
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -144,9 +144,13 @@ const { text, model, usage } = await fused.ai(prompt, {
   `fused.env === "local"` and degrading gracefully when `"hosted"`. Both runtimes expose
   it, so the check is a positive signal, not the absence of an API.
 - **RH-11** `fused.ai(prompt, opts?)` asks an AI model through the shell: the server's
-  `/api/ai` relays to a **local OpenAI-compatible proxy** (`/v1/chat/completions`) at a
-  base URL from the `ai_base_url` preference (`FUSED_RENDER_AI_BASE_URL` overrides;
-  default `http://127.0.0.1:8317`). Resolves with `{ text, model, usage }`; rejects with
+  `/api/ai` relays to a **local OpenAI-compatible proxy** (`/v1/chat/completions`).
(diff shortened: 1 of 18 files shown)


vasu2001:
## Bugbot review addressed — 327b82b

Both code findings were real; I reproduced each before fixing. The four doc findings were raised against an early commit (`ee29ea9`) whose text has since been superseded, but the wording was genuinely imprecise in ways that would mislead an implementer, so I've tightened all four rather than waving them off.

### Code — both confirmed and fixed

**Startup timeout orphans proxy process (High)** — real. Reproduced with a stub binary that starts and never serves: the process survived the failed spawn. Since no state file is written on that path, nothing afterwards knew its pid, `_confirmed_ours` could never prove it was ours, and each retry would spawn another beside it — worse under `FUSED_RENDER_AI_PROXY_PERSIST`, where `setsid` has already detached it. Now terminated and reaped via a new `_terminate_unhealthy_child`, which signals the `Popen` we just created directly rather than going through `_kill_current_ai_proxy` (that reads a state file which deliberately doesn't exist yet).

**Stale instance not killed before respawn (Medium)** — real. Reproduced by `SIGSTOP`ing a live proxy: the probe fails while the pid stays alive, and the respawn then overwrote the config and state describing it, leaving it running, unreachable and unreapable. Now killed while the state identifying it still exists. Worth noting ownership is still provable for a hung instance — `_confirmed_ours` has two tiers, and the cmdline tier (our exact `-config` path) answers when the probe tier can't. An unconfirmable stale pid (recycled onto an unrelated process) is logged and left alone rather than blocking the respawn, since refusing to start would be the worse failure.

Both are pinned by regression tests, and I mutation-tested them: removing the orphan fix fails 2, removing the hung-kill fails 1.

### Docs — all four corrected

- **Claude auth URL slug mismatch (High)** — the doc said `<provider>-auth-url`, which would build `claude-auth-url` and 404. The code was always correct (`_CALLBACK` maps `claude → anthropic`); the doc now names the slug swap and why it exists.
- **Accounts never start the proxy (High)** — the code was already correct: lazy spawn is on *first use*, which includes a management call, and the listing is deliberately non-spawning so opening the tab doesn't pay a ~15s startup. The doc said "first `fused.ai()` call", which implied otherwise. Fixed.
- **Health poll omits API key (High)** — the poll has always sent the generated bearer token; the doc just didn't say so. Now explicit, including why an unauthenticated poll would 401 until the deadline.
- **Re-login success detection fails (Medium)** — this one is already fixed in code (`get-auth-status`, not a listing count) but the finding is a good catch about *why*: credentials are `<provider>-<email>.json`, so a re-auth updates rather than adds, and a count-based check would report a successful re-login as a timeout failure. The doc now records that as a reason the superseded approach was wrong, not just weaker.

Full suite: 2138 passed. The 3 failures (`test_deploy.py` ×2, `test_calls.py`) are pre-existing and fail identically on the base branch.

PR #345, opened 2026-08-01
docs: support opening and saving .odt files

amanbagrecha:
## Summary
- The docs template (Word .docx editor) hardcoded `docx` at every layer even though pandoc, which it already bundles, round-trips ODT just as well.
- Detects the source format from the file's own extension for import and save-in-place, so `.odt` files open, edit, and save correctly without being forced through a docx conversion.
- Lists `.odt` alongside `.docx` in the template's file browser (Save a copy… / Open dialogs).
- Registers `.odt` with the `docs`/`reader` templates in `registry.json`, and adds it to the desktop file-association table (`winopen.py` icon/MIME maps, regenerated `scripts/file_associations.json`).

## Test plan
- [x] `pytest tests/test_file_associations.py tests/test_mime_package.py tests/test_docs_readonly.py` — pass
- [x] Manual round-trip: generated a `.odt` via pandoc, ran `docs.main(action="import", ...)` then `action="save"` then re-imported — content persisted correctly and format stayed `odt`
- [x] Confirmed `.docx` behavior unaffected (`_source_fmt` defaults to `docx`)
- [ ] `tests/test_template_listing_mount.py` has 6 pre-existing failures on Windows (`FileNotFoundError` from a Windows path resolving through a Linux-style fixture) unrelated to this change — confirmed they fail identically on `main` before this PR

---

> [!NOTE]
> **Medium Risk**
> Touches document import/save and format detection; behavior change is limited to new .odt paths while .docx stays on the existing pipeline, with a known loss of comments on .odt save.
> 
> **Overview**
> Extends the **docs** template so **`.odt`** files are first-class alongside **`.docx`**, using pandoc for import and in-place save based on the file extension (`_source_fmt` in `docs.py`) instead of always forcing **docx**.
> 
> The editor boot path accepts **`.docx` and `.odt`**; **Save a copy…** / **listdir** browsers list both. **`.odt` in-place saves** use content-only HTML (`docToHtml`) because pandoc’s ODT writer would inline comment markers as visible text—**`.docx` still uses `docToSaveHtml`** for native Word comments.
> 
> **`.odt`** is registered for the **docs** and **reader** templates (`registry.json`) and in desktop file associations (`winopen.py`, `scripts/file_associations.json`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e05892e241f6fd4c27c7196b383c634a35ed9e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/fused_render/templates/docs/docs.py
+++ b/fused_render/templates/docs/docs.py
@@ -3,12 +3,13 @@
 # ///
 """Backend for the docs preview template (fused-render).
 
-The document is one user file — a Microsoft Word .docx file — converted
-to/from the editor's HTML by pandoc. Everything the editor offers is limited
-to what pandoc's HTML<->docx round-trip genuinely preserves, so the .docx is
-the single source of truth: text, headings, lists, tables, images, math, and
-comments (written as native Word comments via pandoc's comment-start/
-comment-end spans). Version history lives in the file's JSON sidecar
+The document is one user file — a Microsoft Word .docx or OpenDocument .odt
+file — converted to/from the editor's HTML by pandoc. Everything the editor
+offers is limited to what pandoc's HTML<->docx/odt round-trip genuinely
+preserves, so the file on disk is the single source of truth: text, headings,
+lists, tables, images, math, and comments (written as native Word/ODT
+comments via pandoc's comment-start/comment-end spans). Version history lives
+in the file's JSON sidecar
 (<file>.json under the "docs" key). This script only holds what genuinely
 needs Python: pandoc conversion, PDF via the typst compiler, and browsing the
 filesystem for "Save a copy…". Params arrive as strings; annotate.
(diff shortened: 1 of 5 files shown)


amanbagrecha:
Fixed both bugbot findings in 0e05892e:
- The `WORD_EXT` → `DOC_EXT` rename left two call sites (`doExport`, and the Save-a-copy file-list click) still referencing the removed name — fixed both, and confirmed the previous state would have thrown for every download/save-a-copy, docx included.
- Save-a-copy's file-list name-stripping now uses `DOC_EXT` instead of a docx-only regex, so clicking a `.odt` entry no longer produces a double extension like `report.odt.docx`.

While auditing the rest of the change for anything else affected, found one more real issue: pandoc's `.odt` writer has no equivalent of the docx writer's `comment-start`/`comment-end` span convention. Verified directly with pandoc that round-tripping a comment-marked span through `-t odt` inlines the comment text into the visible paragraph instead of producing a real comment or dropping it. Save-in-place now only uses the comment-preserving HTML serialization for `.docx`; `.odt` falls back to the same content-only serialization already used for non-docx export downloads, so comments are dropped cleanly instead of leaking into the document body.

iamsdas (approved):

PR #350, opened 2026-08-03
claude template: "Open preview" button after successful replies

AkshilVT:
## What

After a chat turn finishes cleanly (`done && !error`), the claude template now offers an **Open preview →** button under the reply. Clicking it drops the top-level `_mode` param, landing on the file's default view so the user can immediately see what the reply changed — no hunting the mode switcher.

## Behavior

- Appears **only after successful replies** — never on error turns (start failure, claude exiting with an error, mid-stream crash/kill with partial text), and not when the annotate `claudeReturn` auto-return is about to navigate anyway.
- **Persists across round trips**: `session_id` survives the hop, and history restore re-adds the button when the transcript ends on an assistant reply (errored turns never reach the transcript as assistant rows). `resumeRun`'s clean-finish paths add it too, so a run that finished while you were in the preview still offers the hop back.
- Hidden when it would be dead: panes whose mode lives inside the `_layout` span (deleting top `_mode` would be a no-op) and cross-origin shells.
- One row per transcript — the hop always shows the file's current state, so older rows are removed rather than accumulating; removed on a new send.

## Implementation

- `returnToMode("")` now means "the default view": `_mode` is deleted rather than set (absent is how the shell spells default; `_mode=` would be a distinct broken mode name).
- The `_layout=(…)` span excision — previously duplicated in `clearTopClaudeComments` and `returnToMode` — is factored into `spliceLayoutSpan()`, also used by the new `canOpenPreview()` guard.

## Testing

- `pytest tests/test_claude_agent_sidecar.py tests/test_claude_permission_bridge.py tests/test_templates.py tests/test_templates_api.py tests/test_core_templates.py` — 328 passed.
- Playwright end-to-end (headless chromium against a live server + real `claude -p` runs):
  - no button on boot / home;
  - start-error turn (bogus model) → error row, no button;
  - clean haiku turn → button appears, single row, click removes `_mode`, keeps `session_id`, lands on the rendered default view;
  - two full claude → preview → claude round trips → button restored from history each time, exactly one;
  - mid-stream `SIGTERM` of the run's process group after text had streamed → error row, partial reply removed, no button; reload after the kill → transcript ends on the user row, still no button.

---

> [!NOTE]
> **Low Risk**
> Scoped to the claude template’s chat UI and shell URL editing; behavior is guarded and covered by existing shell-contract tests plus described e2e flows.
> 
> **Overview**
> After a **successful** Claude turn (not errors, not when annotate auto-return runs), the chat transcript shows a single **Open preview →** control under the latest reply. It calls `returnToMode("")`, which now **removes** top-level `_mode` so the shell shows the file’s default view instead of treating `_mode=` as a bogus mode.
> 
> **URL handling:** duplicated `_layout=(…)` parsing is centralized in `spliceLayoutSpan()` for comment clearing, return navigation, and `canOpenPreview()` (button hidden when top `_mode` isn’t meaningful). The hop is stripped on new sends and while a run streams; it’s restored on history load when the transcript ends on an assistant message, and on late `resumeRun` clean finishes.
> 
> **Tests:** annotate shell harnesses now load `spliceLayoutSpan` alongside existing claude URL helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11c952b5106ad8c220528cba0cee0072f3ae7d04. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Screenshots

**Chat after a successful reply — the "Open preview →" button (dark / light):**

Open preview button, dark theme

Open preview button, light theme

**After clicking — landed on the file's default view (`_mode` dropped, `session_id` kept):**

Default preview after the hop

### DIFF
--- a/fused_render/templates/claude/template.html
+++ b/fused_render/templates/claude/template.html
@@ -148,19 +148,20 @@
   .error { color: var(--error); }
   .error::before { content: "✗ "; }
 
-  /* the "Open preview" hop offered under a successful reply */
+  /* the "Open preview" hop offered under a successful reply — brand orange,
+     same accent as the banner and permission cards */
   .viewout { margin: -8px 0 14px 20px; }
   .viewout button {
     font: inherit;
     font-size: 12px;
     padding: 3px 10px;
-    border: 1px solid var(--border);
+    border: 1px solid var(--orange);
     border-radius: 5px;
     background: transparent;
-    color: var(--dim);
+    color: var(--orange);
     cursor: pointer;
   }
-  .viewout button:hover { color: var(--fg); border-color: var(--border-focus); }
+  .viewout button:hover { background: var(--orange); color: var(--on-orange); }
 
   /* ── permission cards (the approval bridge) ─────────────────── */
   /* Claude runs headless here, so a tool the session's rules don't already
--- a/fused_render/templates/claude/template.html
+++ b/fused_render/templates/claude/template.html
@@ -148,6 +148,20 @@
   .error { color: var(--error); }
   .error::before { content: "✗ "; }
 
+  /* the "Open preview" hop offered under a successful reply */
+  .viewout { margin: -8px 0 14px 20px; }
+  .viewout button {
+    font: inherit;
+    font-size: 12px;
+    padding: 3px 10px;
+    border: 1px solid var(--border);
+    border-radius: 5px;
+    background: transparent;
+    color: var(--dim);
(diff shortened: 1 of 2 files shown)


AkshilVT:
bugbot run

vasu2001 (approved):

PR #355, opened 2026-08-03
Inspect a PDF from inside PDF Studio: classification, OCR routing, security scan, Markdown

amanbagrecha:
Adds an **Inspect** drawer to the `pdf_studio` template: what kind of PDF is this, which pages need OCR, and is there anything in the file that acts on its own when you open it.

Prompted by [firecrawl/pdf-inspector](https://github.com/firecrawl/pdf-inspector). That library is a Rust crate with PyO3 bindings, and the built-in executor only sees the server's own environment — a template that needed it would be broken for everyone who hadn't installed it. So `pdf.py` implements the same detector against the already-bundled pymupdf/pikepdf, reusing pdf-inspector's thresholds and vocabulary so the verdicts line up. If `pdf_inspector` *is* importable it is reported alongside as a cross-check, and `health` exposes its version.

## What you get

Topbar **Inspect** button (also `Ctrl+I`, Document menu, three Tools cards) opens a drawer beside the page rather than over it:

- **Overview** — TextBased / Scanned / ImageBased / Mixed with confidence, which pages need OCR, and a one-click hand-off to the existing OCR action
- **Pages** — per-page content type, characters, image coverage, table count; click a row to jump there. Pages needing OCR also get a badge on their thumbnail
- **Fonts** — embedded / subset / ToUnicode per font, flagging Identity-H fonts with no Unicode map, where extracted text is gibberish and OCR is the only way to read the page
- **Security** — a ten-point checklist: JavaScript, `/Launch`, auto-run triggers, embedded files, forms/XFA, external links, multimedia, encryption, signatures, incremental revisions — with the offending code shown per finding
- **Markdown** — headings, lists, code blocks, tables, links, column-aware reading order; copy or save `.md`

The scan reads structure only; nothing in the PDF is executed.

New `pdf.py` actions: `inspect`, `to_markdown`, `save_markdown`.

## Things worth knowing

- `page.get_texttrace()` does the heavy lifting — one entry per text-showing operator, and each character arrives as `(unicode, glyph_id, origin, bbox)`. `glyph_id == 0` is notdef, `unicode` of 0/0xFFFD is undecodable, and `type == 3` is invisible text (an OCR layer). That gives the text/scan call *and* broken-encoding detection with no rendering.
- **Revision counting**: a linearized file carries a second `%%EOF` for its first-page xref, so it is discounted against `Pdf.is_linearized` — otherwise every web-optimized PDF reads as incrementally updated.
- **`/GoTo` is explicitly benign.** Treating unknown action types as suspicious made a 78-page report emit 298 findings from ordinary internal links.
- Cost is ~70 ms/page, dominated by `find_tables()`. Tables are counted on the first 40 pages; past 200 pages the scan samples evenly, keeping first and last, and the report says so.
- The report is dropped on every mutation, since the page numbers in it would be stale.
- The drawer is a flex sibling of `#canvasWrap`, not an absolute overlay like `#toolsHome`, so the page stays readable next to the report.

## Layout

The inspection code lives in its own module, so `pdf.py` grows by 65 lines rather than 837:

| File | Lines | Owns |
|---|---|---|
| `pdf.py` | 1535 (+65) | editing actions, library, working copies, dispatcher |
| `inspector.py` | 826 (new) | classification, security scan, fonts, Markdown — read-only |

`pdf.py` loads it **by path** via `importlib`, not by name. The built-in executor runs template helpers with `sys.path` untouched (it says so explicitly: *"never a sibling imported by name"*), and under the subprocess engine the script's own directory *is* on `sys.path`, where a module called `inspector` could shadow something else. The coupling is one-way and small: `pdf.py` resolves the working-copy path and the page selection, then calls `report(path, idx)` or `markdown(path, idx)`. Whole-directory staging (`copytree`) and hatchling's default package data both ship it, the same way the notebook template ships `kernel_body.py`.

## Review pass

Fable reviewed the diff before this PR opened and found 9 confirmed bugs plus 5 smaller issues. All are fixed here, each with a regression test:

| Fixed | Was |
|---|---|
| **XSS via PDF metadata** | `pdfDate()` returned its input unescaped when the `D:YYYYMMDD` regex failed, and both call sites interpolated it into `innerHTML`. A PDF with `/CreationDate ()` ran script the moment the Overview tab rendered. Now escaped; a browser test asserts four payloads across four metadata fields render as inert text |
| **Security scan crashed on malformed PDFs** | `/AA.items()`, iterating `/Annots`, `NameTree()` and `int(/SigFlags)` all raise on a wrong-typed object — so one malformed `/AA` cost the *entire* report, and those malformations correlate with the files worth scanning. Now type-gated |
| **Markdown auto-linker corrupted every real link** | The bare-URL pass ran after link building and re-wrapped its output as `[text]([url](url))` |
| **Link anchors were dropped or swallowed whole lines** | The old `md.replace(_inline(spans), …)` silently no-op'd on any transformed line (bullets, headings, lists) and otherwise made the whole line the anchor. Now matched against word boxes, so the anchor is the words the link rect actually covers |
| **Document-level JavaScript wasn't "automatic"** | Scripts in `/Names/JavaScript` run at open, but the Automatic-actions row said "Nothing fires on open" while the JavaScript row failed — contradictory on the row you'd actually read |
| **`/FileAttachment` payloads were invisible** | Only the `/EmbeddedFiles` tree was read, so an `.exe` on an annotation reported "No attachments" |
| **Stale report could render for the wrong document** | The scan doesn't supersede; an edit or document switch mid-scan landed a report describing pages that had moved. Now generation-guarded |
| **`E. coli` became `- coli`** | The bare-letter list pattern ate sentence-leading initials. Parenthesised form only |
| **`to_markdown` timed out silently on big documents** | `find_tables()` ran on every page with no cap against the 60 s limit. Now capped at 60 pages with an actionable error |

A second pass then caught a bug in one of those fixes: `_wrap_link` replaced the first *substring* match, so it hit "here" inside "Adherence", the anchor of an earlier link on the same line, and words inside another link's URL. It now matches whole words and skips existing links and code spans. Two more from that pass: a damaged object graph no longer costs the entire report (the scan degrades to an "unreadable" verdict — itself the finding — while pages, fonts and classification still render), and a failed scan no longer leaves the drawer stuck on "Scanning…".

## Cursor bugbot

All five findings fixed, each with a regression test:

| Fixed | Was |
|---|---|
| **Optional engine could abort the scan** (High) | `_rust_engine` caught only `ImportError`, so any failure inside `pdf_inspector` took the whole report down. The cross-check is now incapable of that: absent, it reports the install command; present but failing, it reports the error; either way the pymupdf verdict stands |
| **Malformed name tree broke the scan** | `pikepdf.NameTree` raises on a malformed `/Names/JavaScript`. The tree is walked by hand and type-gated, so a broken script tree costs its own findings, not the scan |
| **Linkifier rewrote code** | URLs inside code spans and indented code lines became nested links |
| **Stale report after clearing the document** | `clearOpenDoc` never dropped the report, so removing the last library entry left the previous document's inspection on screen |
| **Leaked document handle** (Low) | `_font_report` opened a second handle and never closed it; it reuses the open one |

Three more on the follow-up commit, also fixed:

| Fixed | Was |
|---|---|
| **Outline actions were never scanned** | A bookmark can carry `/A` actions, so a `/JavaScript` or `/Launch` sitting on one went unreported. The `/First`-`/Next` tree is now walked with a cycle guard (a malformed outline can point back at itself); those findings stay out of the "Automatic actions" row, since clicking a bookmark is a click |
| **Identity-V fonts not flagged** | Only `Identity-H` counted as undecodable. `Identity-V` has the same missing path back to Unicode and extracts as gibberish too — the rule now matches any Identity CID encoding with no ToUnicode map |
| **Stale markdown after a document switch** | `mdConvert` had no generation guard, so switching documents mid-convert wrote the previous document's markdown onto the new one and Copy/Save exported the wrong file |

Also: linearized files no longer misreport revisions as edits (wording is "cross-reference sections", and >64 MB says "not checked" instead of "0"); the undecodable-glyph threshold is no longer asymmetric with notdef; a crafted `?itab=` can't throw during boot; `Ctrl+I` closes the drawer as well as opening it; the sampled-scan hint no longer claims "the first 40 pages"; and three unused per-page fields are gone.

## Verification

Backend exercised over 9 documents (1–148 pages: IRS XFA form, DocuSign contract, a 148-page report, a wedding invitation, a scanned page) plus synthetic PDFs carrying JavaScript, a `/Launch` action pointing at `calc.exe`, and a `payload.exe` attachment — all three flagged, and the encrypted and image-only cases classified correctly.

50 backend regression checks (one per review finding — including a self-referential outline that must terminate, and `_font_report` driven over all four Identity/ToUnicode combinations) plus 19 browser assertions in a real Chromium: verdict and confidence, OCR thumbnail badges matching the reported pages, row-click page jump, a mutation invalidating the report, re-scan from the empty state, `Ctrl+I` toggling, `?inspect=`/`?itab=` surviving a reload, and the risky-content banner with its findings. A dedicated XSS test fires four payloads through four metadata fields and asserts zero injected nodes and zero window flags set.

All of it re-run after the module split, through the real server, with identical results.

`tests/test_templates.py`, `test_theme.py`, `test_pdf_studio_readonly.py`, `test_engine_requirements.py` pass (313 checks; the 2 failures in `test_templates.py` reproduce on `main`).

---

> [!NOTE]
> **Medium Risk**
> Large new code path parses untrusted PDFs and renders metadata in the UI; security scanning is structural-only but the feature surface and prior XSS/metadata issues warrant careful review despite fixes noted in the PR.
> 
> **Overview**
> Adds **read-only PDF inspection** to PDF Studio via a new **`inspector.py`** module (loaded from **`pdf.py`** with **`importlib`**, not a sibling import) and three backend actions: **`inspect`**, **`to_markdown`**, and **`save_markdown`**.
> 
> The **Inspect** drawer sits beside the viewer (top bar, **`Ctrl+I`**, Document menu, Tools cards) with tabs for **Overview** (text vs scanned classification, OCR hints, optional **`pdf_inspector`** cross-check), **Pages**, **Fonts**, **Security** (object-graph action walk, checklist, findings), and **Markdown** (layout-aware conversion with copy/save). Thumbnails get **OCR** badges from the verdict; cached reports are **invalidated** on edits and guarded against stale async results.
> 
> **`health`** now reports optional **`pdf_inspector`** without failing when it is missing or broken. **`#toolsHome`** z-index is raised so it stays above the inspector drawer.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 589fe9ed9c137b04bc24a88e1bc45a799beb369c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


amanbagrecha:
All three findings from the latest review are fixed in 476ea5a5, each with a regression test:

- **Outline actions skipped by scan** — the `/First`-`/Next` tree is now walked, with a cycle guard since a malformed outline can point back at itself. Verified against a bookmark carrying `/JavaScript` and a sibling carrying `/Launch` (both found, both marked risky), a self-referential outline (terminates, reported once), and an ordinary two-chapter TOC (still clean). These stay out of the *Automatic actions* row, since clicking a bookmark is a click.
- **Identity-V fonts not flagged** — the rule now matches any Identity CID encoding with no ToUnicode map. `_font_report` is driven over all four combinations (Identity-V/H without a map → flagged, WinAnsi → not, Identity-H *with* a map → not).
- **Stale markdown after document change** — `mdConvert` now takes the same generation + document guard as the scan. Verified in a browser by switching documents mid-convert: the old markdown is discarded rather than landing on the new document.

Full suite re-run: 50 backend checks, 19 browser assertions, the XSS test, the nine-document sweep, and `tests/test_templates.py`/`test_theme.py`/`test_pdf_studio_readonly.py`/`test_engine_requirements.py` (313 passing; the 2 failures in `test_templates.py` reproduce on `main`).

amanbagrecha:
Both findings are fixed in 7b9232af — but two reviews in a row surfaced the same two *shapes* of bug, so I replaced the mechanisms rather than adding another case to each.

**Action discovery was enumerating containers.** There were five hand-written walks, one per place an action was known to live, and each review found the next container the list didn't mention: annotations, then outline entries, now form fields whose scripts sit on a parent rather than the widget. That list was never going to be complete. Those walks are gone, replaced by one traversal of the object graph — an unknown container now costs a less specific label and nothing else. Recognition needs two rules because either alone is wrong: reached through `/A`, `/AA` or `/OpenAction` it's an action whatever it claims to be (catching unknown and obfuscated types), and reached from anywhere else it must name a real action type in `/S` (keeping `/S /Transparency` on a group dict and `/S /P` on a structure element out of the findings). "Fires without a click" is now recorded at discovery instead of inferred from a label's wording.

Writing it that way surfaced two bugs the container walks had been hiding:

- **`/OpenAction` can be a destination array** (`[page /Fit]` — "open at page 1") rather than an action. Walking it as one reported the page as a nameless action *and* marked that page visited, so its real annotations were never scanned at all. `main.pdf` in my test corpus is exactly such a file; the regression baseline caught it.
- **`_security_or_unreadable` caught `Exception`**, so a `NameError` in the module degraded into a plausible-looking "unreadable" report instead of failing loudly. It now catches what malformed *input* raises — including `pikepdf.DataDecodingError`, which doesn't derive from `PdfError` — and lets programming errors surface.

**Staleness guards were per-call-site and success-path only.** Hence `mdConvert` being missed, and now the error path. There's one `scoped()` helper deciding whether an answer still applies, for results and failures alike, and it's armed in `py()` — when the request is made, not when it lands. That last part fixed a real race the test exposed: switching documents takes a round trip, and until `state.doc` caught up an in-flight scan still looked current. The exclusion list there names the harmless *reads*, so a mutation added later invalidates by default instead of being silently missed.

19 new checks cover every container above, plus an unknown one, the `/S`-that-isn't-an-action cases, a self-referential graph, and the destination-array regression. Browser tests confirm a discarded failure stays silent while a live one is still reported. Rebased onto current main; full suite green (69 backend checks, 19 browser assertions, XSS, nine-document sweep, 313 repo tests).

amanbagrecha:
All four fixed in 4762aa09. One of them was another **two-parallel-lists drift**, so I removed the second list rather than editing it.

- **`/RichMediaExecute` omitted from the Multimedia row** — the root cause was that `ACTION_RISK` classified action kinds and each checklist row then *re-enumerated* the kinds it covered. Those two lists were always going to drift, and this is where it did: the finding appeared while the row that summarises it said "pass". Each action now declares its checklist row in the same table that gives it a level, and the rows select by that. Adding an action type can no longer leave it visible in the findings but absent from its row. A test asserts every row name originates in that one table.
- **Duplicate attachment findings** — a file present in both the `/EmbeddedFiles` tree and a `/FileAttachment` annotation was recorded twice and could raise two executable findings for one payload. Deduped by name.
- **Same-URL links dropped** — `_link_text` treated any `](uri)` already on the line as proof the link was wrapped, so a second annotation pointing at the same URI was skipped and its anchor left unlinked. The check is gone entirely: `_wrap_link` only ever touches text *outside* an existing link, which is already the correct behaviour to run per annotation. Verified with two links to one URI on a single line — both wrapped, no nesting.
- **Inspector covered the Tools overlay** — Tools is absolutely positioned over `#main` while the drawer is a flex child at a higher `z-index`, so the drawer painted over the card grid. Tools now sits above it; the test probes `elementFromPoint` over all 16 cards and the drawer's own centre.

Also fixed a latent fragility in my own browser tests: the readiness predicate dereferenced `#docName` before the iframe had navigated, which threw instead of waiting — it had been passing by luck.

Full suite: 80 backend checks across six files, 19 browser assertions, plus the XSS, stale-markdown, stale-failure and Tools-overlay browser tests. 313 repo tests pass (the 2 `test_templates.py` failures reproduce on `main`).

amanbagrecha:
Both fixed in a2940453. Each had a mechanism behind it rather than a missing case, so both fixes remove the way the bug was produced.

**An unrecognised action showed up in Findings while every row said "pass".** This is the same summary-vs-detail contradiction as the `/RichMediaExecute` drift last round, and the reason it recurred is that a checklist row only exists if some check *thought to ask* for that category — so any category no check asks about is invisible in the summary no matter how the table is organised. The checks now record which categories they consumed, and whatever is left over is summarised as **Other active content** (`fail` if any leftover is a danger, else `warn`). A finding the reader can see can no longer sit behind an all-pass checklist, including for action kinds added after that function was written — which is what the previous fix could not promise.

`/GoToDp` also got a real entry in the action table. It had been the one member of `ACTION_TYPES` with no risk classification, which is why it fell through to "unrecognised"; `ACTION_TYPES` is now exactly the table's keys, so the last list that had to agree with it is gone.

**A discarded scan left the drawer on "Scanning…".** `scoped()` decided whether an answer still *applied*, but nothing owned the busy state — each call site cleared it on its own success path, so the discard path cleared nothing. Two changes:

- Every drawer request goes through one helper that repaints on *any* outcome — answer, failure, or discard — so no outcome can leave the spinner up. The explicit `try/catch` that runInspect needed to clear "Scanning…" on failure is gone with it.
- Voiding now repaints immediately instead of when the void answer arrives. The first version of this fix only repainted on arrival, which is technically unstuck but means a minute of "Scanning…" for a 148-page result that is already going to be thrown away — the test caught that, since it asserts the drawer clears within 1.2 s of the Save while the scan takes far longer to return.

The status reset is guarded on the label still being the drawer's own, so a dropped scan landing after a Save can't overwrite "saved".

11 new backend checks: an unknown action kind reaching a row, `/GoToDp` staying benign, every risky kind in the table reaching a non-pass row, a row name the function has never heard of still being summarised, and the two cases that must *not* trigger the catch-all (document JavaScript, executable attachments). One new browser test drives the reported path through the UI — Ctrl+S during a scan of a 148-page document — and asserts the drawer clears at the discard, that the void answer stays discarded when it arrives ~30 s later, and that a rescan still works.

Full suite: 91 backend checks across seven files, 19 browser assertions in the e2e run, plus the XSS, stale-markdown, stale-failure, Tools-overlay and stuck-drawer browser tests. `test_templates.py`/`test_theme.py`/`test_pdf_studio_readonly.py`/`test_engine_requirements.py`: 313 passed, the same 2 `test_templates.py` failures that reproduce on `main`. Running the *entire* `tests/` tree hung twice in unrelated places (`test_claude_permission_bridge.py` in `mkstemp`, then `test_pyramid_mount.py` in a socketserver teardown); the first hangs identically from a clean `main` checkout, so I did not chase it here.

amanbagrecha:
Both fixed in a4a870a1 — and they led to a third that matters more than either.

**A structure element's `/A` is an attribute object, not an action.** The "reached through an action slot, it's an action whatever it says it is" rule was applying inside `/StructTreeRoot`, where `/A` means *attributes*. A 148-page tagged report in my test corpus was producing **34 nameless findings** from its tagged figures, and the file read as `notable`. It had read as `clean` until the previous commit's catch-all row stopped hiding `unknown`-row warnings — so that fix was correct and immediately exposed a wrong finding underneath it. Nothing in `/StructTreeRoot`, `/Resources` or `/Metadata` can fire, so the walk no longer enters them, and that file is clean again with a test pinning it.

Worth saying plainly: my earlier "nine-document corpus sweep is clean" was not true for that document once the catch-all landed. It is now, and the sweep is a test rather than a thing I ran by hand.

**The `/A` destination array** was the same bug the `/OpenAction` fix caught two commits ago, reached by the other path — you're right that guarding one slot and not the other was arbitrary. Both now share one rule: an action slot holds a dictionary that is not a page, and an array there is a `/Next` chain rather than a destination. The `/OpenAction`-specific check is gone in favour of that.

**The budget.** Two separate faults, as you describe: leaves (names, numbers, strings) were pushed only to be popped, and font/structure subtrees consumed the rest — then a truncated walk returned a partial list with nothing to say about it, so the checklist could report a clean bill of health for a file it never finished reading. Leaves are no longer pushed at all, and an unfinished walk now surfaces a **Scan coverage** row, placed first because it qualifies every row under it, with the risk no longer able to come out `clean`. The 148-page tagged report now completes within 20k of the 150k budget, where before it was spending the budget on the structure tree it had no business reading.

While testing that I found a locked PDF in my own corpus crashing the scan: `pikepdf.PasswordError` derives from `Exception`, not `PdfError`. That's the second time a hand-kept list of pikepdf exceptions missed one (`DataDecodingError` was the first), so the boundary now asks the module which exception classes it has — 11 of its 13 don't derive from `PdfError`. Programming errors still surface: a test injects a `NameError` and asserts it is not swallowed.

Also cleaned up what these changes made stale: `scoped()` had exactly one caller after the last commit, so it and the `DROPPED` sentinel fold into `insWork`, which now owns both staleness and busy state in one place; plus a dangling comment and a docstring that described the old hand-kept list.

13 new backend checks (tagged figures, the real corpus file, `/A` and `/OpenAction` destinations, a `/Next` array chain, truncation reported and surfaced, the budget headroom, the locked file, and the injected `NameError`). Full suite: 104 backend checks across eight files, 19 e2e assertions, plus the XSS, stale-markdown, stale-failure, Tools-overlay and stuck-drawer browser tests. Template/theme/readonly/engine tests: 313 passed, same 2 pre-existing `test_templates.py` failures.

amanbagrecha:
Fixed in 9fdf7b41, with a caveat on the stated cause.

**I could not reproduce it through ToUnicode.** I rebuilt a real Type0 font's `/ToUnicode` with four malformed CMaps — a bogus surrogate pair `<DBFFD800>`, reversed surrogates, a lone high surrogate, and a non-UTF-16 four-byte value — and MuPDF comes back with `0xFFFD` or the raw code every time. Across every PDF in my corpus the maximum `ucs` is `0xFFFD` and the minimum is `0`. So the specific path described doesn't produce an out-of-range value in pymupdf 1.28.

I applied it anyway, because this one isn't the same shape as the speculative branch I reverted last commit. There the message described a UI state that cannot occur. Here `ucs` is a C int filled in by a font parser reading attacker-controlled bytes, "MuPDF currently clamps" is a property of one version rather than a contract, and the existing test was already half-bounded — `ucs > 32` bounds the bottom for the `unique` set but nothing bounds the top before `chr()`. Completing that test is one expression, and the failure it prevents is a security scanner dying on the file it was pointed at.

The fix counts a value that is not a code point as **undecodable** rather than dropping it, which also fixes a real gap: MuPDF documents `-1` for a glyph with no unicode, and `-1` fell through both branches — reported as neither readable nor broken. The character count and the undecodable count now agree for such a font.

6 checks, driving the real `_page_scan` through a page whose font hands back `0x110000`, `0x7FFFFFFF`, `-1` and `INT_MIN` alongside valid ones: no crash, four counted undecodable, the three valid code points still counted, every character accounted for, and `0x10FFFF` (the top of the range) still treated as readable. Plus a real document re-classified unchanged.

Full suite: 110 backend checks across nine files, 19 e2e assertions, the five targeted browser tests, and 313 template/theme/readonly/engine tests with the same 2 pre-existing `test_templates.py` failures.

amanbagrecha:
Fixed in 8f0ef27d, and you were right that the previous fix was the wrong shape — pruning `/StructTreeRoot` treated a symptom, and `/SE` walks straight around it.

The spec already draws the line I was trying to draw by hand: **`/S` is a required key in an action dictionary.** So `/S` is what makes something an action, and where it was found only decides how much benefit of the doubt its value gets — any `/S` in an action slot (catching unknown and obfuscated types), a known one anywhere else. That is the whole rule now.

Everything special-cased before falls out of it:

- A structure element's attribute object has no `/S`, so it is not an action whether the walk reaches it through `/K`, through an outline's `/SE`, or not at all.
- A movie annotation's playback activation dictionary has no `/S` either — thanks for that one, I had not considered it and it would have been the next report.
- A destination's page has no `/S`, so the `slot()` "not a page" check is gone. Better than before: the page now goes on to be walked *as a page*, so its own annotations are scanned rather than merely not-misreported.
- `SKIP_KEYS` stays, but for the step budget only, and its comment says so. Correctness no longer depends on it, which is the point — a test asserts the tagged-bookmark case is clean with `/SE` present.
- The `(unnamed)` action kind is gone along with the rule that produced it.

What forcing is actually for still works: `/S /Frobnicate` on a link annotation is still found, still `warn`, still moves the verdict off `clean`.

One thing worth flagging, since it cuts against the direction of the last two rounds: an `/A` dictionary carrying `/JS` but no `/S` is now *not* a finding. That is deliberate — no viewer executes it, since `/S` is what dispatches an action — and the walk still descends into it, so a real action nested deeper is still found. A test covers exactly that shape.

14 new checks: the `/SE` hole, the movie activation dictionary, an unknown `/S` still caught, a non-action `/A` that is walked through to a script one level down, `/OpenAction` and `/A` destinations not hiding a page's annotations (with the finding correctly located on "page 1 annotation" rather than the document), the ordinary script/launch/URI cases, and the three corpus files pinned to their exact verdicts.

Full suite: 124 backend checks across ten files, 19 e2e assertions, the targeted browser tests, 313 repo tests with the same 2 pre-existing failures.

@amanbagrecha I have also kicked off an independent Fable review of the whole diff, focused on trying to break the four mechanisms these rounds produced rather than on line-level nits. I will post what it finds before this merges.

amanbagrecha:
Two rounds together: the `/OpenAction` finding (6e4e5581 6e4e5581 → fixed in 6e4e5581) and an independent Fable review of the whole diff (fixed in 06e5d029).

## The reported finding

Correct, and it was mine from the commit before: removing the "not a page" filter let a destination's page be walked *as a page* — right in itself — but it walked in carrying the trigger's `automatic` flag, so a click-only `/SubmitForm` on the opening page was reported as firing on open.

My first repro returned the *right* answer, because the `/Pages` walk happened to reach page 1 before the `/OpenAction` path. I had to force the losing order to see it — a page-open trigger on page 3 whose value is a destination pointing back at page 1, since page 3 is popped first. Arriving at a page already resets the label because it is a new context; it resets the flag now too. A document that opens at page 5 does not click the buttons on page 5.

## Fable review — two findings above the rest

**The scan was nondeterministic.** `pikepdf.Dictionary.keys()` returns a **set**, so the walk visited keys in hash order, randomised per process — and since `seen` keyed on the object alone, whichever path arrived first decided how a shared object was classified. Verified across fixed `PYTHONHASHSEED` values, with two live consequences:

- an obfuscated action (`/S /Frobnicate`) referenced from both an action slot *and* an ordinary key could **vanish entirely** — `risk: clean`, zero findings — defeating precisely the unknown-type detection the slot rule exists for;
- a script used as both `/OpenAction` and a click could be reported `automatic: false`, so the checklist printed "Automatic actions: pass — Nothing fires on open" for a document that runs JavaScript on open.

Every backend call is a fresh subprocess, so ↻ Rescan could disagree with the first scan of the same file. Dedupe is now per *reference* — `(object, automatic, recognised)` — because an object reachable as both an open action and a click genuinely is two triggers, and that is now what the report says. Pinned by a test that runs the scan under eight hash seeds and requires identical output.

**Executable attachments were judged by the wrong name.** `/EmbeddedFiles` keys are lookup names; a viewer shows and extracts the filespec's `/UF`. A `payload.exe` filed under `"readme.txt"` was reported as a harmless text file — a false negative in the one check most worth getting right — and the mirror case raised a false alarm on a harmless file keyed `"also.exe"`. The real name decides now, and since a mismatch is itself the trick, the finding names both: *payload.exe (filed as readme.txt)*.

## The rest of the review, all fixed

- `/AA` mouse events (`/E /X /D /U /Fo /Bl`) are no longer "fires without a click" — a mouse-up handler is a click. Page `/O`//`/C` and field `/K`//`/F`//`/V`//`/C` still are, and an unlisted event still counts as automatic, since over-reporting a trigger is the safer way to be wrong.
- A **file specification dictionary** (what Acrobat writes whenever there is a `/UF`) was rendered as a multi-line pikepdf repr in the Launch row and the finding's code block. One helper now reads `/UF or /F`, which also fixed a `/FS` given as a plain string being **dropped entirely** — that file reported "No attachments".
- `javascript:` and `file:` **URI actions** were reported as ordinary web links, so `javascript:app.alert(...)` came out `clean` with "No scripts embedded". They now land on the script and local-file rows.
- **The app's own edits looked like hidden history.** The scan reads the working copy, and pdf_studio saves text edits incrementally, so a document the user had just typed into reported "3 extra cross-reference sections — earlier content may still be recoverable". Revisions and web-optimization are now read from the file on disk, which is what those facts are about.
- An **inherited `/FT`** is now followed, so a signature field carried by a parent is no longer reported "Unsigned".
- `EXEC_EXTS`: Windows drops trailing dots and spaces when it executes, so `payload.exe ` and `payload.exe.` were slipping through; added `.msc .pif .cpl .chm .url .appref-ms` and the macro document types.
- One font object under two resource names was two rows and inflated the font count.
- `_wrap_link` emitted `[a](https://…/Foo_(bar))`, which breaks the link — bracketed URLs go in `<>` now.
- **Save Markdown re-ran the whole conversion**, spending a second 60 s budget and able to differ from the preview; it writes the text already on screen.
- Two status-pill bugs, both from `insBusy` being shared by two request kinds: a scan finishing while a slower convert ran had committed its report and then left the drawer showing "Scanning…" (repaint now happens on every settle, not only the last), and Convert → PDF to Word used the *identical* "converting…" label, so a discarded scan cleared the Word conversion's busy pill (drawer labels are distinct now, and only the drawer restores its own).

Two of Fable's lower-priority notes I have **not** taken, for the same reason I reverted a locked-file message earlier: `report()` has no degradation boundary of its own, and a user-password PDF gives an unhelpful error — but Fable confirmed neither is reachable through the UI, because `open_doc` fails first. I would rather not add a branch for a state the app cannot be in.

Fable also reported what it could **not** break, which is the more useful half: no XSS (it enumerated every interpolation in the added JS and confirmed each PDF-derived string goes through `esc()`), no way past the checklist-coverage rule, nothing that escapes the pikepdf degradation boundary, no cross-document report from any open/mutate/scan/convert interleaving, and `/Next` chains, name trees, `/AA` indirection, destination arrays, `SKIP_KEYS` and the page-sampling arithmetic all correct. It also re-scanned 21 real PDFs consistently.

29 new checks for this round. Full suite: **164 backend checks across twelve files**, 19 e2e assertions, the five targeted browser tests, and 313 repo tests with the same 2 pre-existing `test_templates.py` failures.

amanbagrecha:
Latest round, plus the CI failure and stale code.

## CI was red because the branch was 8 commits behind main

The `test-python` failures were not in the diff — they were `test_templates.py` / `test_history_template.py` asserting a registry order that main changed under me (#360 claude_split added a template, #362 moved versions before git). My branch still carried the old tests against the old registry. **Rebased onto current main**; the twelve commits replay cleanly (three files, no conflicts), and the full repo suite is now **495 passed** with only the 3 pre-existing Windows-only failures (`path.endswith("duckdb/template.html")` on a `\`-separated path — they pass on Linux, which is why main's own CI is green). Force-pushed the rebase; a fresh run is going now.

## Two review findings fixed (589fe9ed)

- **A broken optional engine could abort boot.** `_health` runs behind the `health` action that gates PDF Studio starting, and it caught only `ImportError` on the optional `pdf_inspector` probe. A prebuilt native wheel can raise more than that — `OSError` loading the `.pyd`, a package `__init__` that throws — and that would take down the whole health check, and separately an inspect run through `_rust_engine`. Both now treat a broken optional engine exactly like an absent one: the app runs without it, and health reports `pdf_inspector_error` without flipping `ok`.
- **A backslash in a link broke Markdown conversion.** `_wrap_link` handed the finished link to `re.sub` as a *replacement string*, where `\1` is a group reference and `\t` an escape — so a URL like `https://x/a\1b` raised `re.error` and aborted the entire conversion, and `\t` silently became a tab. A callable replacement takes the text literally. Verified end to end: a real PDF whose link URI contains `\1` now converts instead of throwing.

## Two more findings were already fixed

Both the string-`/FS` attachment and the `javascript:`/`file:` URI findings were reported on `6e4e5581` — the commit *before* `06e5d029`, where I fixed exactly those. Confirmed against HEAD: a `/FileAttachment` with a string `/FS` of `payload.exe` reports `risky`, and a `javascript:` link is a script finding. No change needed.

## Stale code

Swept all three files for unused symbols after the rebase: nothing of mine is dead (`_remote_dir` in `pdf.py` is pre-existing on main). The `scoped()`/`DROPPED` helper and the `needs_ocr` field were already removed in earlier rounds as they became stale.

## Coverage

9 new backend checks (backslash URLs in anchor and target, a real markdown conversion over a poisoned link, a broken-import `_rust_engine`, a broken-import `_health`, and an absent engine still reporting install instructions). Full suite after rebase: **173 backend checks across thirteen files**, 19 e2e assertions, the five targeted browser tests (drawer staleness, stuck-drawer, XSS, Tools overlay, markdown guard), and 495 repo tests with only the 3 Windows-path failures that pass on CI.

PR #373, opened 2026-08-04
Google Drive mounts: in-app sign-in, upload visibility, and a link-first Mounts page rewrite

iamsdas:
## Summary

Connecting cloud storage no longer sends the user to a terminal. Three things changed, and the Mounts page was rewritten around them.

- **In-app sign-in for Google Drive, Dropbox and Box.** `POST /api/mounts/remotes` hardcoded `"type": "s3"` and had no way to get an OAuth token, so the page's advice for any consent backend was "go run `rclone config`". The server now spawns `rclone authorize "<backend>"` — which owns its own browser and loopback callback — and turns the returned token into a remote **over the rc daemon, never argv** (a token on a command line is readable by any local user via `ps`). Drive collects the user's own OAuth client, because Google is retiring rclone's shared client ID; Dropbox and Box need none and are never shown the fields.
- **A revoked token says something that can actually fix it.** Credential-health detection only ran for `env_auth` remotes, so an expired Drive token fell through to a generic "reconnect it" — advice that cannot repair an OAuth failure. The gate covers `type == "drive"` and the remedy text is chosen per backend, since telling a Drive user to run `aws sso login` is the wrong-remedy bug this path exists to prevent.
- **Silent upload failures are visible.** `VFS_OPT` sets `CacheMode: "full"`, so a write succeeds the moment it hits the local cache and uploads later — a quota or permission rejection at the remote appeared nowhere, after the user had already seen the save succeed. Every mount now reports its pending/failed queue, read from rcd's structured `vfs/queue`.
- **Mounts page rewrite.** Two competing entry points (a Remote dropdown embedding uncreated remotes as `+` options, and an "Add storage" grid whose modals ended in prose telling you to scroll back up) collapse into one link-first flow: paste `s3://`, `gs://` or an S3/GCS console URL and Name/Remote/Path fill in; provider cards sit at the foot of the same section. Completing a setup scrolls the form in, focuses Path and flashes the details grid instead of explaining itself in four duplicated paragraphs. Pasting a link now prefers **your own credentialed remotes** over "public datasets" (the live UI was doing the opposite, then failing the mount), and a date-like tail (`…/telemetry/2026`) is qualified by its parent rather than proposing a mount named `2026`. `Mounts.tsx` (1700 lines) is split into `shell/mounts/{MountList,AddMount,setup,links}`.

Scope is full `drive` (`drive.file` only sees app-created files, which is useless behind a mount), with `skip_gdocs` so Docs/Sheets never appear as editable-looking files whose saves can't round-trip.

Rationale and rejected alternatives: **D219–D226** in `DECISIONS.md`.

## Visuals

The browser round-trip belongs entirely to the rclone child; the server spawns, watches, and converts the token.

```mermaid
sequenceDiagram
    participant UI as Mounts page
    participant API as /api/mounts
    participant CH as rclone authorize
    participant RCD as rcd daemon

    UI->>API: POST remotes/oauth
    API->>CH: spawn (own browser + callback)
    loop until settled
        UI->>API: GET oauth/status
    end
    CH-->>API: token on stdout
    API->>RCD: config/create (not argv)
    UI->>API: refresh remotes
```

Both entry points converge on one pre-filled form:

```mermaid
flowchart TD
    A[Paste a storage link] -->|recognized| B[Details auto-filled]
    P[Provider card] --> M[Setup modal]
    M -->|connected / Use| B
    B --> C[Confirm & correct]
    C --> D[Add & mount]
```

## Usage

**Sign in:** Mounts → *No link? Connect a provider first* → pick a provider → name the remote → approve in the browser → the form scrolls in with the remote preselected; type a folder path and **Add & mount**.

**Paste-a-link:** paste an `s3://`, `gs://` or console URL into "Paste a storage link", correct anything that looks wrong, **Add & mount**.

**API:**
```bash
curl -X POST localhost:PORT/api/mounts/remotes/oauth \
     -H 'X-Fused: 1' -d '{"name":"gdrive","provider":"drive",
                          "client_id":"…","client_secret":"…"}'
curl localhost:PORT/api/mounts/remotes/oauth/status   # poll until in_flight is false
curl -X POST localhost:PORT/api/mounts/remotes/oauth/cancel -H 'X-Fused: 1'
```

Upload state rides the existing `GET /api/mounts` payload as `uploads: {pending, failed, failed_names} | null`. `null` means *not read* and is deliberately distinct from zero — an unknown queue must never render as all-clear.

## Test Plan

- [x] **75 new backend tests** in `tests/test_shell_mounts.py`: argv shape, token parsing (marker frame + bare-JSON fallback), the `X-Fused` guard, 400-before-502 ordering, creation over rcd, cache invalidation, the no-token retryable path, concurrent-sign-in rejection, cancel, rollback (incl. its refusal to delete a remote the user asked to replace), backend-specific advice, and the upload queue's unknown-vs-zero distinction.
- [x] **Token-leak regression pinned** — failure detail is built from stderr, since stdout is where the token lives.
- [x] **72 frontend tests** across `shell/mounts/links.test.ts` and `platform/lib/{oauth,uploads,google-client}.test.ts`, written red→green: pasted-link remote priority, date-like name qualification, one-shot handoff semantics, GCS console URL shapes, and every stopping condition of the sign-in poll.
- [x] **Suites green**: backend `4515 passed, 107 skipped`; frontend `166 pass` across 14 files; `tsc --noEmit` and `check:boundaries` clean. One pre-existing unrelated failure — `test_supervisor_linux_paths.py::test_alert_is_a_no_op_when_tk_cannot_start`, which fails on a clean tree too (no `libtk8.6.so` on this machine).
- [x] **Visual verification via browser automation**: the rewritten page, all six setup modals, the live setup→form handoff, and the light theme; the submit button now exposes an accessible name (`Add & mount`, was `""`).
- [x] **Verified against real rclone v1.74.4** rather than assumed.
- [x] **Review findings addressed** (see below).
- [ ] **Manual smoke against a real Google account is outstanding** — CI has no browser and no credentials to complete a consent screen. A reviewer should confirm the consent round-trip, then that the mount reads and writes.
- [ ] **Manual: failed-upload display.** Easiest reproduction is mounting a remote you lack write permission on, saving a file, and confirming the card reports the failure rather than staying silent.

## Review findings

Six BugBot findings; three were already fixed by later commits on the branch (re-auth under an existing name via the Replace opt-in, and both cancel-races in `oauthCancelOutcome`). The rest:

- **Cancel resolved before the server let go** — `terminate()` only starts the teardown, so restoring the sign-in button handed the user a retry that 409'd on the sign-in they had just cancelled. An accepted cancel now returns a third outcome, `standdown`, and a second poll (`oauthStandDownTick`) waits for `in_flight` to clear. It reports a remote created inside that gap, and reads the server's own "was canceled" as silence rather than an error banner (**D225**).
- **A 409'd start left nothing to cancel** — the refusal cleared `connecting`, which took the Cancel button with it, so the message told the user to cancel something no longer on screen; the only exit was the 300s timeout. The form now asks the status endpoint whether one is really in flight and offers "Cancel that sign-in" (**D225**).
- **`replacing` was read from the client's flag** — the Replace checkbox stays ticked while the name is edited to a free one, so a brand-new remote was marked as replacing and a failed `config/create` skipped its rollback, leaving exactly the half-created remote that rollback prevents. Derived server-side from whether the remote exists (**D226**).
- **GCS object links kept `;tab=live_object`** — the console matrix param was stripped from the bucket segment only, but the object view hangs it on the key, so the mount pointed at a path that does not exist. Stripped from the tail too, anchored on `;tab=<word>` because `;` is legal in an object name.

Merged `origin/main`, which reorganized the frontend into `apps/`, `shell/` and `platform/` behind `@platform` aliases; this branch's files moved with it and its decisions were renumbered off main's D205–D218.


iamsdas:
## Review round 1 — seven findings fixed

Two reviewers (general correctness + a silent-failure pass) went over the branch. Seven findings were confirmed against the code and fixed in commits `f04b658a`…`1b4d7477`. Suite: **3960 passed** (+41 from the first round), 73 frontend tests, 0 fail.

### The one that mattered

**An unreadable upload queue rendered as a clean card** — the exact bug D207 exists to prevent. `UploadQueue` collapsed `!uploads` (*unknown*) and `pending === 0` (*verified empty*) into the same silent branch, directly beneath a comment saying those must never be conflated. A healthy read-write mount whose `vfs/queue` read failed showed all-clear while files sat stranded — strictly worse than before the feature, since the UI now implies the queue is watched.

It was not fixable in the component: the wire contract was ambiguous, with `uploads: null` meaning both *not applicable* (read-only/unhealthy mount) and *read failed*. Fixed on both sides — `_mount_upload_status` now returns `{unknown: true, reason}` carrying **no counts**, so it cannot be misread as zero, and the TS type is a discriminated union, so a caller that skips the unknown case does not compile. The rc error is now logged (it was previously discarded entirely), `ValueError`/`JSONDecodeError` are caught alongside `RuntimeError`, and the `probe()` worker is wrapped so a thread death reports unknown rather than nothing.

### Also fixed

| # | Finding | Fix |
|---|---|---|
| 2 | OAuth token retained in a module global for the process lifetime | `job.out.clear()` immediately after the parse, both branches |
| 3 | Mounts page polled forever on a permanently-failing upload (`pending` counts failed items, which rclone re-queues indefinitely) | Gate on `pending > failed` — work that can actually drain |
| 4 | No server-side guard against `config/create` overwriting an existing remote | 409 **before spawning the child**, so it's refused before the browser opens; `replace: true` is the explicit opt-in |
| 5 | Sign-in poll retried forever on fetch failure with no feedback | Bounded by consecutive failures and a wall clock |
| 6 | Cancel cleared the UI before awaiting the server; `canceled` field discarded | Awaits first; `canceled: false` reconciled as "it already succeeded" |
| 7 | Partially-created remote could linger after a timeout | Mirrors `create_detected_remote`'s rollback, including its amended message when the delete itself fails |

### Two deliberate judgement calls

- **No credential probe after remote creation.** `create_detected_remote` probes because a dotfile credential is of unknown age; this token is seconds old from a consent just completed, and an `lsd` against a large Drive would add up to 30s to the spinner the user is watching. Recorded in the docstring with a trigger for revisiting.
- **Rollback is skipped when `replace: true`.** Deleting a remote we did not create would destroy the user's previous working config on top of an already-failed sign-in. The error message tells them to check it instead; a test pins this.

### Verified sound (not changed)

The child-process lifecycle was reviewed in detail and held: lock discipline, the cancel/finish/restart race interleavings, SIGTERM→SIGKILL on both the timeout and cancel paths, and that the token reaches neither argv, a log line, nor any user-visible error string. The `GET /api/mounts` join budget is genuinely unchanged, since the upload probe is mutually exclusive with the credential probe.

### Still outstanding

**The manual smoke against a real Google account remains the gap.** No automated test here exercises a real consent screen — CI has no browser and no credentials — so the OAuth round-trip and a real read-write mount still need a human to confirm.

PR #394, opened 2026-08-05
Preview pane: URL-synced pane mode (_panelMode) and top-bar switcher hide

vasu2001:
## What

Two explorer preview-pane changes:

1. **`_panelMode` URL param syncs the pane's mode.** The right preview pane's mode menu now writes its choice to the URL as `_panelMode`. The pane component remounts per selection (keyed on the previewed path), and each mount re-seeds from the param — so the chosen mode persists when switching rows instead of resetting to the default. The param also rides directory navigation alongside the sticky `preview` param (both `navigate()` and the breadcrumb's `_mode`-preserving hop), and is removed when the pane is toggled off. A selection that doesn't offer the mode falls back to its default while the param stays put, so the next selection that does offer it picks it up again.

2. **Top-bar mode switcher hides while the pane is open.** With `?preview=true`, the pane header carries its own mode switcher, so the top-bar one (portaled into the breadcrumb bar) was duplicate chrome. `TemplatePreview` now tracks the URL (`useUrlVersion`, fired by the pane toggle's `replaceSearch`) and skips rendering the top-bar switcher when the pane is open. Deploy / "Open as app" buttons stay; file previews are unaffected (`preview` is never carried onto file URLs).

## Files

- `ListingPreviewPane.tsx` — mode override seeded from / written to `_panelMode`
- `listing/pane.ts` — pane toggle-off deletes `_panelMode`
- `platform/lib/router.ts` — `navigate()` carries `_panelMode` with `preview` for directory targets
- `Breadcrumb.tsx` — same carry in the `_mode`-preserving breadcrumb navigation
- `Preview.tsx` — hide top-bar `ModeSwitcher` while `preview=true`

---

> [!NOTE]
> **Low Risk**
> Changes are limited to explorer URL params and UI duplication; the `sel` seeding tweak can alter selection on load when URLs omit `sel`, but does not touch auth or server APIs.
> 
> **Overview**
> The listing **right preview pane** now persists its mode in **`_panelMode`** on the query string: the pane menu reads and writes that param (via `replaceSearch`), so changing rows no longer resets mode. **`_panelMode`** is carried with the sticky **`preview`** param on **directory** navigation (`navigate()` and breadcrumb hops that preserve `_mode`), and is **removed when the pane is closed**. Rows that do not support the saved mode fall back locally while the param stays for the next compatible selection.
> 
> **`Preview`** hides the **top-bar** `ModeSwitcher` when `?preview=true` (pane header already has one), using **`useUrlVersion`** so toggling the pane updates chrome without a full navigation.
> 
> **`useListingSelection`** treats the URL as **authoritative** when seeding `?sel` after load: missing `sel` clears a recalled selection; bookmarks and back/forward win over stale recall.
> 
> The pane keeps showing its **loading skeleton** until gates and folder app probe settle, even when `_panelMode` is already in the URL, to avoid a mode flash then jump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3aa4f6aa64ab3b85a44632da28fed0a30866fdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/src/apps/explorer/Breadcrumb.tsx
+++ b/frontend/src/apps/explorer/Breadcrumb.tsx
@@ -208,7 +208,11 @@ function navigatePreservingMode(target: string): void {
   if (mode) {
     const params = new URLSearchParams({ _mode: mode });
     const preview = url.get("preview");
-    if (preview !== null) params.set("preview", preview);
+    if (preview !== null) {
+      params.set("preview", preview);
+      const panelMode = url.get("_panelMode");
+      if (panelMode !== null) params.set("_panelMode", panelMode);
+    }
     navigateUrl(urlForFsPath(target, "?" + params.toString()));
   } else {
     navigate(target, { isDir: true }); // breadcrumb targets are always dirs
--- a/frontend/src/apps/explorer/ListingPreviewPane.tsx
+++ b/frontend/src/apps/explorer/ListingPreviewPane.tsx
@@ -233,16 +233,22 @@ export default function ListingPreviewPane({
     if (allowed(e)) modes.push({ mode: e.mode, icon: templateModeIcon(e) });
   }
 
-  // While the default is still undecided, hold the skeleton — the pane must
+  // While the mode list is still undecided, hold the skeleton — the pane must
   // never settle on an interim mode and then jump. Undecided means: the
   // folder's app probe is in flight (`_app` would lead the list), or any
   // gated template's verdict is unresolved (a higher-ranked conditional mode
   // may still enter the list — and for a self target, an all-conditional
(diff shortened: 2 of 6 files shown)


vasu2001:
Bugbot finding addressed in d3aa4f6a: the skeleton-hold gate no longer skips when `modeOverride` is non-null — it holds whenever the app probe or condition verdicts are pending, so a URL-seeded `_panelMode` can't settle on an interim mode and jump. Safe because both signals resolve once per mount (pane keyed per path) and switcher clicks can only happen post-settle.

PR #409, opened 2026-08-07
Declare dependencies per folder in pyproject.toml, not per file in PEP 723 headers

iamsdas:
## Summary

- **Dependencies are declared once per folder, not once per file.** A `.py` now gets its environment from its project root's `pyproject.toml` — resolved via `app_git.app_dir_for`, the template roots, else the topmost ancestor declaring a manifest — and every script under that root shares one venv, however deep it sits. Per-file `# /// script` headers are no longer read.
- **Breaking, with no migration path.** An existing app with headers stops working until its folder gains a `pyproject.toml`. A file that still carries a header is *reported*, not silently run: the error names the manifest to create and quotes the packages the dead block declared, because running it would otherwise fail later on an import of something the file does declare. Recorded as deliberate in D229 — a release note is required.
- **One install per page instead of N.** Five scripts in one folder now resolve to one key, so the loader shows a single "Preparing *my-app*" row with one cancel button. `installEnv` dedups by key into a module-level promise registry, and the already-installed guard is page-scoped rather than per-call, so a stuck environment fails once instead of once per script.
- **Reproducibility, and environments that survive an upgrade.** `uv sync` writes a `uv.lock` the user commits; the four templates with real dependencies ship a pre-generated one so a released build never resolves against PyPI on first render. Venvs live in `~/.fused-render/venvs/<hash of path>` with a sibling uv cache on the same filesystem, so the folder itself gains only source files and hardlink dedup actually works.
- **Staleness is a digest of `pyproject.toml`**, recorded in a `.fused-source.json` sidecar written before the ready marker. A manifest edit is picked up automatically on the next `/api/run` — nobody ever has to run `uv sync` by hand, which would create an in-folder `.venv` and diverge from the store.

## Visuals

What `/api/run` does with a `.py` path. The folder decides; the file no longer has a say.

```mermaid
flowchart TD
    A[".py path"] --> B{"project root<br/>declares deps?"}
    B -->|no| C{"stray header?"}
    C -->|yes| D["error: name the fix"]
    C -->|no| E["app interpreter"]
    B -->|yes| F{"venv fresh?<br/>digest vs sidecar"}
    F -->|yes| G["project venv"]
    F -->|no| H["needs_install"]
    H --> I["one progress row<br/>per project"]
```

`uv.lock` is an output, never an input to that freshness check — a lock edit alone changes nothing, while a manifest edit always triggers a resync, and the sync runs bare (no `--frozen`) so uv reconciles the lock itself.

## Usage

Declare the folder's dependencies once, then delete the headers:

```toml
# my-app/pyproject.toml
[project]
name = "my-app"
version = "0"
requires-python = ">=3.12"
dependencies = ["altair", "pandas"]
```

Every `.py` under `my-app/` — including `my-app/readers/tiff.py` — runs in that one environment, with its working directory still its own folder. A `pyproject.toml` in a subfolder is ignored and says so rather than sitting inert. Adding a dependency to the manifest is picked up on the next render; no command to run. Folders with no manifest keep running on the app's own interpreter exactly as before.

## Test Plan

- [x] `tests/test_projectenv.py` (new, 509 lines) — root resolution across app dirs, template roots, nested manifests and the home-dir ceiling; venv keying; digest staleness and the stat-fingerprint memo; `gc()`.
- [x] `tests/test_template_locks.py` (new) — every shipped template with a manifest has a lock, and the lock is packaged (the `.gitignore` negation is load-bearing: hatchling excludes gitignored files).
- [x] `tests/test_env_install.py`, `test_engine.py`, `test_engine_requirements.py`, `test_server_env_install.py`, `test_inspector.py`, `test_bundle_contents.py` — reworked onto project keys; D212's marker-stamp/generation machinery re-pinned unchanged; server-side single-flight covered.
- [x] Full suite: **4507 passed, 107 skipped, 6 failed** — the 6 are `tests/test_browse_mount.py`, pre-existing on `main` and untouched here.
- [x] End-to-end against real `uv sync`: install → folder gains only `pyproject.toml` + `uv.lock` → manifest edit reads stale → resync re-locks automatically → no in-folder `.venv` → renaming the folder yields a new key → `gc()` reclaims the orphan.
- [ ] Reviewer check: open a template with dependencies (geotiff or pano) on a clean `~/.fused-render` and confirm one progress row, one cancel, and an instant second render.
- [ ] Reviewer check: point an old header-bearing app at this build and confirm the error names its `pyproject.toml` rather than failing on an import.

**Depends on** openfused `pyproject-project-venv-support` (`c9953ffb`): `resolve_workflow_venv_from_path` gains `subdir`/`uv_path` and stops hardcoding `.venv/bin/python`, which is what lets a project dir without a `scripts/` subfolder resolve at all, and what makes it work on Windows.


iamsdas:
Both Bugbot findings addressed in 702765d6.

### 1. Test needs backend mock — *valid, fixed*

Confirmed rather than assumed. With `fused` made unimportable (an import blocker mimicking what the 3.10–3.13 matrix actually installs, `.[dev]` only), the test failed exactly as described:

```
File "fused_render/engine.py", line 773, in _execute
    backend = get_backend()
File "fused_render/engine.py", line 730, in get_backend
    from fused.agent_core.backends.local.python_compute import LocalPythonComputeBackend
ImportError: blocked: fused
```

The refusal this test replaced returned *before* the backend was reached, so this was new exposure introduced by the test itself, not by the behaviour change.

`@requires_fused` would have been wrong twice over: it's defined below this test in the file, and it would skip the case on every matrix entry — turning the one test covering the newly-changed behaviour into a test that runs only in the single `fused-engine` job.

Stubbed with the file's existing `_FakeBackend` instead, which is a better test anyway. The assertion is about what `run_python` *decides*, not about executing Python, and the fake carries both halves of the backend contract — so it can now assert that the file took the **interpreter** path rather than the venv path, and that PY-17 supplied an interpreter:

```python
assert backend.calls[0]["via"] == "_execute_sync"
assert backend.calls[0].get("interpreter")
```

Neither was visible from the previous `result == 7`. Verified under the blocker: `test_engine.py` → 54 passed, 18 skipped, with both header tests among the passes.

### 2. Migration error names inert path — *already resolved*

Moot as of D233 (75799e41): `ScriptHeaderNoLongerRead` no longer exists. A leftover `# /// script` block is now an ordinary comment — not read, not detected, not refused — so there is no message left to name a path.

Worth noting the finding was **correct for the code it reviewed**, and it's close to why the refusal was dropped: it only fired when the folder had no manifest, so the likelier half-migrated case (manifest written, one stale block in a sibling file) passed straight through it. It never delivered the guarantee it was justified by.

PR #431, opened 2026-08-10
Fix Windows "server did not become ready": decouple readiness + builtin mount-readiness from live mount I/O

amanbagrecha:
Fixes the Windows **"Python server did not become ready"** dialog, and the related cold-start slowness, both rooted in the same mount↔request-path coupling.

## 1. Readiness probe decoupled from `/api/config`

The desktop supervisor's readiness probe polled **`/api/config`**, whose `learn_mount_ready()`/`sessions_mount_ready()` did a **live rcd + WinFsp check on every request**. During cold startup (automount respawning rcd + remounting the builtin `learn`/`sessions` mounts), every `/api/config` response exceeded the probe's 0.5 s timeout for the whole 20 s budget → the supervisor killed the child and retried 3× → the dialog.

**Fix:** a dedicated, dependency-free **`GET /api/desktop/ready`** that echoes only the launch token — no mounts, no rcd, no dir picker — and `desktop_probe` now polls it. Readiness = "the HTTP server is up and is the instance we launched", never "every optional subsystem is warm".

## 2. `/api/config` no longer blocks on mount attach

Even with the probe moved, `/api/config` itself took **62 s** on a cold start (the browser window couldn't paint). Root cause: `builtin_mount_ready` did `mp in mounted_paths()` + **`_ismount(mp)`**, and on Windows `_ismount()` on the WinFsp reparse point blocks for ~the rc timeout while the mountpoint is mid-attach — ×2 builtins. macOS (nfsmount) and Linux (FUSE) attach fast enough to hide it.

**Fix:** `builtin_mount_ready` is now an **I/O-free read of an attach-owned `_builtin_ready` flag** instead of a live probe. `run_automount` is the sole writer: it resets every builtin to **False** before its force-detach+remount pass, then flips one **True** only once its own `attach_mount` succeeds; a manual `reconnect_mount` also sets it on a successful builtin repair. So **True always means a mount this run attached** — never one that merely survived a previous run, which matters because the frontend sticky-caches the first True it sees. The health monitor (`poll_once`) never touches the flag: it must not infer readiness from an observed "mounted" (a mount lingering from a failed detach reads mounted while stale; a transient post-attach snapshot reads not-mounted), so readiness stays deterministic and attach-owned.

## Why Windows-only
`/api/config` is identical cross-platform; the difference is purely mount-backend latency — WinFsp attach/`ismount` is slow, nfsmount/FUSE are fast. Both fixes make the paths robust on every platform.

## Tests / e2e
`test_desktop_probe` (incl. `test_readiness_is_independent_of_mount_health`) and the learn/sessions suites are green (49 targeted tests). New lifecycle coverage in `test_learn_mount`: `builtin_mount_ready` reads the flag not a live probe; `run_automount` marks a builtin ready only after a successful attach and leaves it not-ready on failure; `poll_once` never sets or clears the flag from observation; `reconnect_mount` marks it ready only on success.

Built the installer from this branch and installed it end-to-end on real Windows: **build ~19.6 min, install ~4.6 min, supervisor readiness ~1.4 s** (server log: boot → `GET /api/desktop/ready -> 200`), supervisor + child both alive, **no dialog**. Cold `/api/config` worst-case dropped from **62 437 ms → 63 ms**. (3 pre-existing `test_shell_mounts` failures on Windows also fail on `main`.)

> [!NOTE]
> The browser window can still paint slowly on a *first* cold start — py-spy shows that's the first-run Fused engine bootstrapping the showcase UDFs' project venvs via `uv` subprocesses (plus the WinFsp mount attach), a separate first-run cost. It is not the readiness/mount-probe path fixed here (the supervisor probe runs before the browser opens and stays ~1.4 s).

---

> [!NOTE]
> **Medium Risk**
> Touches desktop startup gating and when Learn/Sessions sidebar links appear; logic is well-tested but wrong flag timing could hide links or show them before attach completes.
> 
> **Overview**
> Fixes Windows **"server did not become ready"** and multi‑second `/api/config` stalls on cold start by separating **supervisor readiness** from **sidebar mount readiness**.
> 
> **Readiness probe:** `desktop_probe` now polls **`GET /api/desktop/ready`**, which only echoes the desktop instance id and token (no mounts, rcd, or dir picker). The supervisor can confirm the launched HTTP server within its timeout instead of failing while builtins are still attaching.
> 
> **Builtin “ready” for `/api/config`:** `learn_mount_ready` / `sessions_mount_ready` no longer call **`mounted_paths()`** or **`_ismount()`** (WinFsp could block ~60s mid-attach). They read an in-memory **`_builtin_ready`** flag set only when **`run_automount`** or a successful **`reconnect_mount`** actually attaches that builtin this run—cleared at automount start so stale mounts from a prior run don’t satisfy the UI’s sticky cache.
> 
> Tests cover `/api/desktop/ready` decoupling from mount health and the attach-owned readiness lifecycle (`poll_once` never infers readiness from observation).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 549d8bf64396bbc396843d76d9ce3fd51ff54aee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/fused_render/desktop_probe.py
+++ b/fused_render/desktop_probe.py
@@ -4,10 +4,19 @@ A desktop launch must confirm that the server answering on its port is *the
 one it started* — not an unrelated process that happened to hold the port
 (a stale dev server, another app). Each launch publishes a per-launch 256-bit
 token plus the desktop instance id into the child/in-process server's
-environment; the server echoes them from `/api/config` (see
-`fused_render.paths.desktop_instance` and the `/api/config` handler). This
-module polls that endpoint with the token header and reports ready only when
-the echoed id + token match.
+environment; the server echoes them from `/api/desktop/ready` (see
+`fused_render.paths.desktop_instance` and the `/api/desktop/ready` handler).
+This module polls that endpoint with the token header and reports ready only
+when the echoed id + token match.
+
+`/api/desktop/ready` is a dedicated, dependency-free endpoint on purpose: the
+probe must reflect only "the HTTP server is up and is ours", never the health
+of an optional subsystem. It used to poll `/api/config`, whose
+`learn_mount_ready`/`sessions_mount_ready` fields do a live rcd/WinFsp check on
+every request; a slow cold-start mount attach then pushed each response past
+the 0.5s per-request timeout below for the whole readiness window, so the
+Windows supervisor declared "Python server did not become ready" and
+killed-and-retried a server that was in fact serving fine.
 
 Shared by both backends: the Windows supervisor (out-of-process child server)
 and the macOS app (in-process server thread). urllib + json only, so it stays
@@ -28,12 +37,12 @@ DESKTOP_INSTANCE_ID = "desktop-v1"
 
 
 def matching_server(port: int, token: str, instance_id: str = DESKTOP_INSTANCE_ID) -> bool:
-    """True iff 127.0.0.1:<port>/api/config answers 200 and echoes our
+    """True iff 127.0.0.1:<port>/api/desktop/ready answers 200 and echoes our
     desktop instance id AND token. Any failure (refused, timeout, non-200,
     bad JSON, mismatched id/token) is a plain False — the caller keeps
     polling until its own deadline."""
     req = urllib.request.Request(
-        f"http://127.0.0.1:{port}/api/config",
+        f"http://127.0.0.1:{port}/api/desktop/ready",
         headers={"X-Fused-Desktop-Token": token},
     )
     try:
--- a/fused_render/desktop_probe.py
+++ b/fused_render/desktop_probe.py
@@ -9,10 +9,7 @@ environment; the server echoes them from `/api/desktop/ready` (see
 This module polls that endpoint with the token header and reports ready only
 when the echoed id + token match.
 
-`/api/desktop/ready` is dedicated and dependency-free on purpose: it reflects
-only "the HTTP server is up and is ours". The probe used to poll `/api/config`,
-whose live mount-readiness check pushed cold-start responses past the timeout
-below and made the Windows supervisor kill a healthy server.
+`/api/desktop/ready` is dependency-free on purpose: it once polled `/api/config`, whose live mount check timed out on Windows cold start and made the supervisor kill a healthy server.
 
 Shared by both backends: the Windows supervisor (out-of-process child server)
 and the macOS app (in-process server thread). urllib + json only, so it stays
--- a/fused_render/desktop_probe.py
+++ b/fused_render/desktop_probe.py
@@ -9,16 +9,10 @@ environment; the server echoes them from `/api/desktop/ready` (see
 This module polls that endpoint with the token header and reports ready only
 when the echoed id + token match.
 
-`/api/desktop/ready` is a dedicated, dependency-free endpoint on purpose: the
-probe must reflect only "the HTTP server is up and is ours", never the health
-of an optional subsystem. It used to poll `/api/config`, whose
-`learn_mount_ready`/`sessions_mount_ready` fields then did a live rcd/WinFsp
-check on every request (since made cached — see `builtin_mount_ready`); a slow
-cold-start mount attach pushed each response past the 0.5s per-request timeout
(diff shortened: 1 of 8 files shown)


amanbagrecha:
Validated e2e on the installer built from this branch: install 157s, warm startup→ready **4.4s**, no dialog. **(1)** Readiness probe moved off `/api/config` to a dependency-free `/api/desktop/ready` — cold start: 31ms vs `/api/config` 62s. **(2)** Residual `/api/config` 62s root-caused + fixed: `builtin_mount_ready` did a live `_ismount()` on the WinFsp reparse point, which blocks ~the rc-timeout ×2 builtins during cold-start attach (Windows-only; mac nfsmount / linux FUSE attach fast). It now reads the health monitor's cached state (same I/O-free source `/api/mounts/health` uses). Re-validated cold start vs the real home: `/api/config` worst **63ms** (was 62437ms); `learn_mount_ready` flips true ~20s later via the background poll. Tests: desktop_probe 14/14, learn/sessions green (3 pre-existing `test_shell_mounts` failures also fail on main).

amanbagrecha:
Bugbot's stale-readiness race fixed in f8570906: `builtin_mount_ready` is now a lifecycle flag `run_automount` clears before its force-detach and sets True only after its own `attach_mount` re-attaches this run — so a prior run's `"mounted"` can't reach the frontend's sticky cache. Test: `test_run_automount_marks_builtin_ready_only_after_attach`.

amanbagrecha:
Second Bugbot finding (reconnect leaves builtins never ready) fixed in 5b25673e: `poll_once` now syncs the builtin readiness flag from observed mount state, gated on `_automount_completed` so it only kicks in after run_automount's startup force-detach+remount (keeping the earlier stale-True fix intact). A reconnect-repaired builtin self-heals within one poll interval. Tests: `test_poll_once_self_heals_builtin_after_out_of_band_reconnect` + `test_poll_once_ignores_surviving_mount_before_automount`.

amanbagrecha:
Third finding (failed-attach overridden to ready) fixed in ce6ec74e. Split the directions: `_builtin_ready` is set True ONLY by an operation that actually attached the mount this run (`run_automount`'s attach, or `reconnect_mount` — now hooked, which also resolves the reconnect finding); `poll_once` only ever CLEARS it on an observed drop. A mount lingering from a failed force-detach or surviving a prior run can no longer flip readiness True off an ambiguous "mounted". Tests: `test_poll_once_never_marks_builtin_ready_from_observation`, `test_poll_once_clears_builtin_ready_on_drop`, `test_reconnect_marks_builtin_ready_on_success`.

amanbagrecha:
Reworked in edfd05fd: both the set- and clear-from-observation directions in `poll_once` were the same inference race (a transient post-attach snapshot reads not-mounted; a lingering/prior-run mount reads mounted). Removed poll_once's involvement entirely — `_builtin_ready` is now attach-owned: set True only by `run_automount`'s attach or `reconnect_mount` (lifecycle.py, which resolves the reconnect finding), reset False by run_automount at each pass; the health monitor never touches it. Deterministic in both directions.

amanbagrecha:
Bugbot check is green. The remaining inline 're-reports the reconnect finding on run_automount — it's addressed: reconnect_mount sets the flag on a successful builtin reconnect (lifecycle.py:826, edfd05fd), covered by test_reconnect_marks_builtin_ready_on_success. run_automount is just one of the two attach paths that set it.

amanbagrecha:
Final build from `edfd05fd` (all four fixes; Bugbot green) validated end-to-end on the real installer: **build 19.6 min, install 4.6 min, supervisor readiness ~1.4 s** (server log: boot→`/api/desktop/ready` 200), supervisor+child stay up, **no dialog**. Note: the browser window still paints slowly on a cold start, but py-spy shows that's the first-run Fused engine bootstrapping the showcase UDFs' project venvs via `uv` subprocesses (+ the WinFsp mount attach) — a separate first-run cost, not the readiness/mount path fixed here.

PR #433, opened 2026-08-10
OpenIndex integration: persistent file index powering explorer search (phases 1-3)

iamsdas:
## Summary

- **In-folder search is now instant and cross-session**: the explorer's recursive search is served from a persistent file index (names/sizes/dates only, no contents) instead of re-walking the tree, with the live streamed walk kept as a silent fallback for anything the index doesn't cover. This also gives Windows/Linux real search where Spotlight doesn't exist.
- **The engine** is OpenIndex's indexer ported into `fused_render/index/` (config/ignore/store/scan/fsevents/runner/worker/query + its specs): path-sorted parquet partitions with a min/max manifest for prefix pruning, a `dirs.parquet` signature cache that makes rescans incremental (unchanged folders cost one check), a detached worker process driving an `events.jsonl` progress log, and a macOS FSEvents fast path. De-globalized into an explicit `IndexConfig`; storage lives under `home_dir()/index`; OpenIndex's raw `sql` action is deliberately not exposed.
- **Scans schedule themselves**: server startup kicks an incremental scan of the user's home (default root; ~7–8 s cold, ~3 s warm at 634k files), debounced 15 min. During a rescan the previous index generation keeps serving (generation-numbered partitions, manifest swapped last — also closes the engine's crash-atomicity gap) and the search box shows an "indexing…" caveat.
- **Preferences → Indexing**: live status, Re-index / Full scan / Delete index, and an editable skip-rules list whose save triggers the reconciling rescan the engine's ignore fingerprint expects.
- **Mount safety is enforced twice**: every fused-render home's mounts dir is ignored by rule and refused as a scan root (`is_mount_backed`), and the walk never crosses onto another filesystem (`root_dev`) — a background crawl of an rclone NFS mount can wedge it permanently.
- Two bugs found by verifying against a live server, fixed here: the worker spawn used `start_new_session=True`, which forces `fork()`+exec and dies in PROJ's atfork handler once the server has loaded pyproj/rasterio (now posix_spawn-compatible, `setsid` in the worker); and a worker that died without a `run_end` reported `scanning: true` for 24 h (now reported dead after 5 min of no activity).
- **Hardened by three review rounds** (two full code-review passes + Cursor Bugbot), all findings fixed with regression tests. Highlights: gitignore parity for the served corpus (the index answer now prunes exactly what the walk prunes, cached per index generation); SQL/glob path escaping so a quote or `[` in `$HOME` can't break compaction; a flock'd store lock serializing writers; per-root ignore fingerprints with legacy migration; case-insensitive partition pruning; status/scan endpoints that handle multiple roots honestly; skip-list comments preserved verbatim.
- **Search performance**: typing used to freeze the UI on index-scale corpora — the ranking comparator built a fresh ICU collator per comparison (~4.5 s per keystroke at 200k hits). One hoisted `Intl.Collator` + precomputed depth brings it to ~90 ms, and scoring now runs as a debounced, sliced background job that never blocks the input.
- **Search results hold still and tell the truth**: while a search is on screen, watch-driven refreshes and index generation swaps are deferred — results never reshuffle under the cursor; the fresh corpus is adopted at a real boundary (typing, clearing the search, or an in-app rename/create/delete). The list shows the **top 100 matches** (ranking still runs over the full corpus; the chip reads `top 100 of 4,880+` and a footer says how many more exist — no infinite scroll, a narrower query beats scrolling). Progress text only says what is true: "Searching…" while anything is pending, an entries-scanned count only during a real fallback walk, and "No matches" only once the scan for the current query has actually finished. Searching the filesystem root `/` now uses the index when covered (a `rstrip("/")` made it always fall back). Results land on their top hit: the first result is auto-selected (Enter opens it hands-free) and follows the ranking as it refines — unless the user moved the selection, which then holds across re-ranks and query changes for as long as its row stays in the results.
- **Explorer polish while in the area**: folders this app mutates are searched live until the next scan covers them; shed table columns (MODIFIED/SIZE at narrow widths) now collapse instead of leaving a dead strip (WKWebView ignores `calc(%)` on fixed-layout cells — verified in the real engine); the relevance/sort chip next to the search box is removed (column headers still cycle asc → desc → relevance).

## Visuals

```mermaid
flowchart TD
    A[Explorer in-folder search] --> B{GET /api/index/search<br/>covered?}
    B -->|yes| C[Index corpus — instant,<br/>last completed generation]
    B -->|no / error / no index| D[GET /api/fs/walk<br/>streamed live walk]
    C --> E[Client fuzzy scoring<br/>unchanged]
    D --> E
    S[Server startup] -->|debounced 15 min| W[python -m fused_render.index.worker]
    P[Preferences: Re-index / Full scan] --> W
    W -->|events.jsonl| ST[/api/index/status/]
    ST -->|scanning + has_index| F["indexing… caveat in search chip"]
    W -->|generation swap, manifest last| C
```

## Usage

- Search any folder in the explorer — results come from the index when it covers the path, silently from the live walk otherwise. While a rescan runs, the match counter shows "· indexing…".
- **Preferences → Indexing**: view status (files indexed, last updated, coverage), trigger *Re-index* (incremental) / *Full scan* (from scratch) / *Delete index* (search falls back to walking until the next scan), and edit the skipped-folders rules; saving rules that differ from what the index was built under starts the reconciling rescan automatically.
- API (mutations guarded by `X-Fused`): `POST /api/index/scan` (`{"root"?, "full"?}`), `POST /api/index/cancel`, `GET /api/index/status|runs|stats|lookup|search|config`, `POST /api/index/config`, `POST /api/index/delete`.

## Design trade-offs

Decisions a reviewer should know about, and what was traded away:

- **Client-side ranking, server-side corpus.** The corpus (up to `MAX_CORPUS = 200k` entries per folder) is fetched once and every keystroke filters it locally — typing is network-free, and fuzzy ranking is *identical* whether the index or the live walk produced the corpus (no rank-flip when the source swaps). Traded: the whole candidate set crosses the wire once, and the client does the scoring work. The scoring hot path is bounded by a hoisted `Intl.Collator` (the per-compare collator was 4.5 s/keystroke at 200k hits; now ~90 ms) plus a debounced (150 ms, skipped ≤20k entries), 20k-per-slice scan job that yields to the main thread between slices.
- **Corpus cap at 200k, shallow-first.** A folder whose raw corpus exceeds the cap is truncated (shallow entries win) *before* gitignore filtering; `truncated: true` tells the client the answer is partial. Traded: matches buried below the cap in enormous trees aren't candidates. Bounded payload and memory won.
- **`fresh` is informational; coverage is the gate.** The client uses the index whenever it `covered` the folder, even mid-rescan — the last completed generation keeps serving and the "indexing…" chip is the staleness signal (explicit product decision). Traded: results can lag the disk between scans. For files *this app* mutates, the mutated folder is marked and searched by live walk until the next scan covers it; out-of-band edits wait for the next scan (startup, debounced 15 min, or manual).
- **Gitignore parity is applied at fetch time, not scan time.** The scan worker knows nothing about git; `/api/index/search` pipes the corpus through the same `check-ignore` oracle the walk uses, cached per (root, index generation) — ~0.4 s once per compaction, ~0.05 s after. Traded: first search after a scan pays the git queries. Gained: a `.gitignore` edit takes effect without a rescan, and the engine stays a dumb file catalog.
- **Default scan root is the whole home directory** (not project roots): a whole-home scan costs seconds and an index covering one project answers almost no search the explorer asks. Traded: first-boot does a home crawl (with the default skip rules), and the ignore list matters more.
- **Ignore rules are name patterns, fingerprinted per root.** The index's skip list is deliberately not gitignore (that's the server layer's job, above). Rules edits fingerprint-mismatch each root and trigger reconciling rescans per stale root. Raw lines (comments/blanks included) are persisted verbatim and cleaned at use, so a comment-only edit doesn't rescan.
- **One store, serialized writers.** Compactions and deletes serialize on a flock'd store lock (fd-held, so a crashed worker releases it); a scan request for a root that's already scanning joins the live run instead of racing it. Traded: no parallel compaction — acceptable, compaction is seconds.
- **Windows lock parity is untested by CI** (no Windows runner): the `msvcrt` branch polls `LK_NBLCK` until acquired to match flock semantics.
- **Deferred revalidation: stillness beats freshness mid-search.** Directory-watch bumps and index generation swaps accumulate while a search is on screen and are adopted only at explicit boundaries (query change, search exit, in-app mutation). Notably, *focus is not a boundary*: pane remounts and WebKit focus restoration fire `onFocus` with no user intent, so refocusing only warms an idle walk against the pinned generation. Traded: mid-search results can lag out-of-band disk changes until the next keystroke.
- **Display cap at the top 100.** Fuzzy rank past the first hundred stops being actionable, so the tail isn't rendered and there is deliberately no "load more" (it answers the wrong question). The counter always reports the true total; keyboard nav and auto-select operate only on visible rows. A guard test keeps the cap constant out of the render layer.
- **No SQL surface.** OpenIndex's raw `sql` action stays out; the only reads are typed endpoints. AI/SQL query features would need a guarded read-only view (phase 4).
- **Mount safety over completeness.** Mount-backed paths are refused as scan roots and the walk never crosses filesystems — an rclone NFS mount can be permanently wedged by a background crawl, so those trees are simply not indexed (the live walk still serves them).
- **Known engine gaps carried, not fixed here:** in-place edits preserving size+mtime are invisible to the incremental scan; `stats` is a full index scan per call (fine until an analytics UI leans on it).

## Test Plan

- [x] ~139 new Python tests across 8 files (`tests/test_index_*.py`): ignore grammar, config, parquet store + generation swap, incremental scan semantics, runner control plane (incl. posix_spawn-compatible spawn kwargs and dead-run reporting), query/lookup, router API, and mount safety (`_no_kernel_on_mount`). Full suite: **5095 passed**, 1 pre-existing failure (`test_calls.py::test_an_abandoned_merge_closes_the_day_s_files`, fails on main too).
- [x] Frontend: `index-corpus`, `index-caveat`, `revalidate`, `result-cap` unit tests (incl. red-first tests for the generation-swap deferral and a source guard on `prefetchWalk`); `bun test` 378 pass, 8 pre-existing failures in untouched `shortcut-chord.test.ts`.
- [x] Verified live in the browser against a real home scan (634k files, 7.6 s cold / 2.6 s warm): index-served search (4,880 matches instantly, zero walk requests), stale-index serving mid-rescan with the "indexing…" chip, Preferences tab (status, buttons, rule editor), delete → walk fallback → re-index cycle.
- [x] Verified live (batches 7–8): capped list is exactly 100 rows + footer; results stay frozen through background churn, blur/refocus, and a full completed rescan of a 639k-file home, then update on the next keystroke; no "0 entries scanned" placeholder anywhere; no "No matches" flash mid-scan.
- [x] Regression verified live: `POST /api/index/scan` from a running server now produces a completing worker (`events.jsonl`, `run_end complete`); before the spawn fix the worker died pre-Python with an empty log.
- [ ] Reviewer: open any folder, type in the search box; then Preferences → Indexing → Full scan and watch the live progress + "indexing…" caveat.


iamsdas:
## Review findings addressed (8/8 confirmed + 4 cut findings)

All findings from the adversarial code review are fixed, each with a regression test that was verified red against the unfixed code where practical:

| Finding | Fix | Commit |
|---|---|---|
| Concurrent scans corrupt the store | `compact()` runs entirely (manifest read included) under an flock-backed `store_lock`; writers serialize, generations stay unique | 56a214e1 |
| LIKE `_`/`%` unescaped in compact's outside-predicate | `like_literal()` + `ESCAPE '\'` — scanning `/x/proj_a` no longer deletes `/x/proj-a` | c8fb9fe3 |
| Same in `search_under`/`stats` prefixes | Same helper; `my_dir` search no longer lists `my-dir` files | c8fb9fe3 |
| Threaded-scan `done` race drops entries | Whole work list counted before any submit | 1d61e657 |
| Ignore fingerprint global, stamped by first root | Per-root `root → sig` map (old format read compatibly); config save rescans **every** stale root | e817ffac |
| Kernel `isdir` before the mount guard | Guard (pure string work) always first; scheduler's own `isdir` removed | 8e3779af |
| Delete-index races an in-flight compaction | `delete_store` takes the store lock; compact re-checks the cancel flag after acquiring it | 56a214e1 |
| No gitignore parity in the corpus | Served entries pass through the walk's own check-ignore oracle (repo-toplevel or standalone-marker scoped), cached per index generation | 6fd5b853 |

Cut findings also fixed: `scanning` now `any(running)` across runs; a root of `/` survives config writes (roots normalize as paths, not ignore patterns); file rows exactly filling the cap no longer drop all dirs with `truncated:false`; capped corpus orders shallow-first (breadth-first character preserved). Deliberately left: the client ignoring `fresh` (documented decision — coverage is the gate, the indexing chip is the staleness signal) and the pure-refactor cleanups.

**Verification:** full suite 5028 passed (25 new tests), 1 known pre-existing failure. Live against the dev server: corpus for the fused-render repo filtered from ~4k to 1,050 entries with zero gitignored leakage, 0.42s first call after compaction, 0.05s cached.

iamsdas:
Second follow-up push (`6fd5b853..ea8b433c`, 16 commits): all 5 Cursor Bugbot findings, all 8 findings from a second full code-review pass, plus three UX/cleanup items.

## Bugbot findings

| Finding | Fix |
|---|---|
| Ignore save starts overlapping scan (High) | `552b6e51` — `runner.start` now joins a live run of the same root (liveness-checked, not mere run-dir presence) instead of starting a second worker |
| Ignore stamp migration drops roots (High) | `0d60131d` — migrating a pre-per-root `ignore_applied.json` carries the old global sig forward as a per-root fallback until each root is individually stamped |
| Config POST returns wrong roots (Medium) | `a68a8075` — POST response now identical in shape to GET (`roots` = effective, `configured_roots` = raw) |
| Windows store lock times out (Medium) | `6b79f1d8` — nt branch polls `LK_NBLCK` until acquired, matching flock semantics (untested by CI — no Windows runner) |
| Corpus cache ignores limit (Medium) | `9e7756cd` — the gitignore-filter cache only stores/serves genuine full-corpus responses |

## Second code-review pass

| Finding | Fix |
|---|---|
| Compaction SQL interpolates store paths unescaped — a quote in `$HOME` breaks every compaction | `46ea3d29` — all path interpolations through `_sql()`; shard files enumerated with `glob.escape` + explicit `read_parquet([...])` list (DuckDB can't escape `[` in globs at all); same class fixed on the reader side in `query.py` |
| Index staleness after in-app fs mutations (rename → old name still found) | `ea8b433c` — the six fs mutations funnel through one seam in `api.ts`; a mutated folder is marked and in-folder search walks it live. Stale-while-rescanning behavior unchanged; the three doc sites that promised an unimplemented `fresh` fallback now tell the truth |
| Re-index / Full-scan buttons only scanned `roots[0]` | `16578a3c` — rootless `/api/index/scan` fans out to every configured root |
| Status showed a finished run's frozen numbers while another root still scanned | `6cb58341` — status prefers a running run's fields |
| Saving preferences destroyed `#` comments in the skip list | `f48f37bc` — raw lines persisted verbatim, cleaned at use; comment-only edits no longer trigger a rescan |
| Case-insensitive lookup vs byte-wise partition pruning returned zero rows for lowercase anchored queries | `391d0960` — compaction stores `min(lower(path))`/`max(lower(path))` bounds; pre-upgrade manifests keep byte-wise pruning and self-heal on the next scan |
| Status polls re-parsed every run's event log from scratch (O(n²) over a scan) | `1ae9feac` — run state memoized on the log's `(size, mtime)`; liveness still re-evaluated per call |
| Index default ignore names diverged from the walk's (source-flip inconsistency) | `22838e43` — one shared ignore floor; the walk uses it, the index extends it |

## Also in this push

- `950f4163` removes the relevance/sort chip next to the search bar (column headers still cycle asc → desc → relevance).
- `b7743357` folds `getJsonSignal` into `getJson`.
- `4a6515da` keys the `IndexConfig` rules cache on the ignore list — no more manual `_rules = None` resets.

Every fix was reproduced red before implementation. Full suite: **5041 passed**, 133 skipped, 2 known pre-existing failures (`test_calls` Linux-only `/proc/self/fd` path; `test_apps_api` xdist flake that passes in isolation). Frontend: build + `tsc` clean, 286 pass, 8 pre-existing `shortcut-chord.test.ts` platform-detection failures (file untouched by this branch).

iamsdas:
Third push (`ea8b433c..dd4273ce`): merge conflict with main resolved, plus performance and layout fixes verified in the live app.

| Change | Commit |
|---|---|
| Typing froze the UI on index-scale corpora — root cause was the ranking comparator constructing an ICU collator **per comparison** (~4.5 s/keystroke at 200k hits, measured). One hoisted `Intl.Collator` + precomputed depth → ~90 ms, ordering verified identical | `5b8fb1a5` |
| Scoring is additionally debounced (150 ms, skipped ≤20k entries) and sliced (20k/slice, yields between slices); pending scan shows "Searching…" instead of a false "No matches" | `d8fa1fda` |
| Dead blank strip between SIZE and the pane divider at narrow widths: shed columns now collapse to rendered zero-width columns. First fix used `calc(100% - 96px)`, which WKWebView silently discards on fixed-layout table cells — caught by live-browser measurement, reworked px-only and re-verified at all three width steps | `0913b389`, `79f15110` |
| Relevance/sort chip removed; `getJsonSignal` folded into `getJson`; `IndexConfig` rules cache keyed on the ignore list | `950f4163`, `b7743357`, `4a6515da` |
| **Merge of `origin/main`** (5 commits, ~4.6k lines): one conflicted hunk in `Listing.tsx` — both sides deleted different controls from the search-bar row, resolved by keeping both deletions | `9a4bbc2a` |
| Bugbot round 3, "Join blocks ignore-rule rescan" (High): a rules save mid-scan joined the in-flight worker still running under the old rules, so the promised reconciling rescan never ran. `runner.start` now joins only sig-matching runs (full covers incremental, not the reverse) and cancels+restarts on mismatch; cancel-pending runs are never joined | `dd4273ce` |

Full suite post-merge: **5083 passed**, 133 skipped, 1 known pre-existing failure (`test_calls`, Linux-only `/proc/self/fd`). Frontend: `tsc` clean, build green, 342 pass + 8 pre-existing `shortcut-chord` platform failures. Column-collapse geometry, search-under-load, and the indexing chip re-verified in the running app after the merge.

iamsdas:
**Bugbot round 4 — both findings fixed** (plus a search-UX batch from live testing), pushed through `c6455e4c` (includes a merge of latest `main`).

### Review findings

- **False empty search during scan (High)** — confirmed and fixed in `a223e998`. `onPublish` dropped the job's `done` flag, so `scanPending` cleared on *any* publish for the current query — an intermediate zero-hit slice or the empty no-corpus placeholder publish could show a confident "No matches" while the scan was still running. Scan state now carries `done`; pending clears only on a final publish for the current query. Folded into a single "what's in flight" ladder for the placeholder: rows → show them; scan pending → "Searching…"; streaming walk with progress → "No matches yet — still searching (N entries scanned)"; settled → the truthful terminal message. The always-zero "(0 entries scanned)" text is gone: the counter only ever moved during a real fallback walk, so it now appears only there.
- **Root path search always misses (Low)** — confirmed and fixed in `b8ce9e3c`. `search_under`'s `rstrip("/")` turned `/` into a falsy `""` → `covered: false` every time. Same `or "/"` normalization `stats` already used; everything downstream already special-cased `/`. Regression test lives in `test_index_search.py` (whose fixture actually seeds a root-dir row) and was mutation-tested against the fix.

### Also in this push

- **Deferred revalidation (`bed7ed4e`)**: watch churn and index generation swaps no longer reshuffle results mid-search; the fresh corpus is adopted at explicit boundaries only (typing, clearing, in-app mutation). Root-caused a live regression where a focus event could smuggle a newer generation past the deferral — focus is ambient (pane remounts, WebKit focus restoration), not a gesture, so `prefetchWalk` now only warms an idle walk against the pinned generation, with a source guard test.
- **Top-100 display cap (`a77e9ce5`)**: ranking still runs over the full corpus; the list renders the top 100 with a true-count chip (`top 100 of 4,880+`) and a "N more matches not shown" footer. Infinite scroll removed.
- **Merge with `main` (`c6455e4c`)**: main's crumb-bar search chip redesign (terse chip + full sentence in title/aria) kept; the cap and caveat semantics layered onto it via `cappedAway`/`visibleHits` so the cap constant stays out of the render layer (guard test enforced).

### Verification

- Python: 5095 passed, 1 known pre-existing failure (`test_calls`, fails on main).
- Frontend: tsc clean, build clean, `bun test` 378 pass + 8 known pre-existing `shortcut-chord` failures.
- Live against a real 639k-file home index: cap + footer exact, results frozen through background churn, blur/refocus, and a full completed rescan (then updated on the next keystroke), no "(0 entries scanned)", no mid-scan "No matches" flash.

iamsdas:
**Bugbot round 5 — all three unresolved findings fixed in `666cae0f`.**

- **False Escape claim blocks auto-select (High)** — confirmed. A query change empties the results for one commit and the listing reconcile clears the selection with them; `nextSearchSelection` read that as an Escape claim and refused to auto-select forever after the first refine. The empty commit now gets no judgment at all, and a null-lead claim whose anchor row is no longer in the results dies with its anchor (a real Escape — cleared while the row was still visible — is still honored).
- **Stale corpus survives index lifecycle (High)** — confirmed. Nothing ever re-keyed the corpus fetch after Delete index / a completed rebuild, because the filesystem didn't change so no dir-watch refresh arrived. New `indexLifecycleCount` signal (`lib/index-freshness`), bumped on delete and when the status poller observes a new `last_completed_at`, composed into the generation the revalidation deferral already manages — so mid-search it adopts at the same explicit boundaries (deliberately: a background scan finishing must not swap results under a reader), and immediately when no search is on screen.
- **Mutated folders still hit index first (Medium)** — confirmed. A folder already marked dirty now starts the live walk before any request instead of waiting out the index round-trip; the resolution-time check stays for the mutation-lands-mid-flight race.

Per instruction this round was pushed without the usual live-verification pass; tsc/build clean, existing suites unchanged (only the known platform failures).

PR #498, opened 2026-08-13
Home's folder cards open the folder's page, not its listing

iamsdas:
## Summary

- Clicking a folder card on the Explorer home page (Repos / Artifacts tabs) landed you in a file listing. It now opens the folder's **page** — `index.html`, else the first non-hidden top-level `.html` by name — and only falls back to the listing when the folder has none.
- This extends **D269** ("a folder with a top-level page IS that page") to the last surface that still showed a folder without honoring it; the /apps hub cards and the preview pane already did this.
- Costs **zero extra requests**: `FolderStack` already lists the folder to draw the card's preview, and the click target now reads that same listing through the prefetch cache.
- Directory **bookmarks are deliberately not retargeted** — see below.

## Visuals

```mermaid
flowchart TD
    A[Click folder card] --> B{Listing complete?}
    B -->|unresolved, unreadable, truncated| C[Folder listing]
    B -->|yes| D{Top-level .html, not gitignored?}
    D -->|index.html, else first by name| E[Open that page]
    D -->|none| C
```

Two guards live in the card, not in the shared rule — `lib/app-entry.ts` stays a byte-for-byte twin of `templates/shared/app_entry.py::entry_html`:

- **Gitignored pages are filtered out before the rule runs**, so a card never opens a page it didn't draw (the Repos tab's own preview already drops `ignored` entries). Filtering rather than vetoing matters: with an ignored `coverage.html` beside a real `page.html`, a post-hoc veto would drop the card to its listing even though a good page exists.
- **A truncated listing falls back to the folder**, so the shell can't resolve a different entry than the server's full `os.listdir` would.

One `folderOpenTarget()` value per render feeds both the anchor's `href` and the click handler, so a plain click, a middle-click and a Cmd-click cannot land in different places. Before the listing resolves the href carries the folder — a link that is late, never one that is wrong.

## Usage

Not separately invocable — it changes where an existing card goes. Open the home page, pick the **Repos** or **Artifacts** tab, and click a card for a folder holding an `index.html`.

Nothing else moved:

| Surface | Behavior |
|---|---|
| Home folder card (Repos/Artifacts) | **Now opens the folder's entry page** |
| Directory *bookmark* card | Unchanged — opens its saved URL |
| Navigating into a folder from a listing | Unchanged — listing, page auto-selected (D263) |

A bookmark is a URL the user saved *while standing on it* — a recorded navigation, the case D269 explicitly leaves alone. Retargeting it would also drop the view params saved with the URL, contradict the `armBookmark` Update-bookmark tracking armed against that URL, and break on bookmark URLs that aren't fs paths at all (`/_tab`, `/_panel`).

## Test Plan

- [x] `frontend/src/apps/explorer/lib/app-entry.test.ts` — 4 new cases for `folderOpenTarget`: entry page resolved as a file, two-page folder picks the first by name, no-page and empty folder answer the folder, `null` entries (unresolved *and* unreadable) answer the folder. Also stubs `location` so the file runs solo, which it previously could not.
- [x] `frontend/src/apps/explorer/BookmarkCards.test.ts` — 7 new cases for `folderCardTarget`: the rule through the card, null listing, ignored-only folder, an ignored page not outranking a real one (×2), `ignored: false`/absent meaning "real", and truncated (×2). Both guards mutation-checked: removing either turns the matching tests red and nothing else.
- [x] Full frontend suite green (887 pass / 0 fail); `tsc --noEmit` clean; module-boundary check clean.
- [ ] A Repos card for a repo with a gitignored top-level `.html` (e.g. `coverage.html`) and no other page still opens the repo listing; with a real second page, it opens the real one.
- [ ] Home → Repos/Artifacts: left-click a card for a folder with a top-level `.html` lands on the page, not the listing.
- [ ] Middle-click and Cmd-click that card open the same destination in a new tab.
- [ ] A folder with no top-level `.html`, and an unreadable folder / dead mount, still navigate to the listing — no dead link, no spinner.
- [ ] Network panel: one `/api/fs/list` per card, not two.
- [ ] A directory bookmark card still opens its saved URL with its saved params, and the breadcrumb's Update-bookmark affordance still arms.


iamsdas:
Wrong surface — this changed the explorer home page's folder cards; only the /apps page cards needed the change. Closing in favor of a fresh PR.

PR #565, opened 2026-08-17
The scheduled-message banner warns; the chat stays open

AkshilVT:
## What

A pending scheduled message that names the session on screen used to **close this chat's composer**. That block is withdrawn; the banner stays.

> "when there is a scheduled message, we block the chat. Let's not do that. Let's keep the banner, but let's not block the chat for now."

## The block, and everything that was part of it

Grepping the state variable turned up five readers, and all five go:

| what | where |
| --- | --- |
| `box.disabled = blocked` | `applyComposerBlockState` |
| the placeholder swap to `"Waiting on a scheduled message…"`, and `BOX_PLACEHOLDER` which existed only to restore the real one | same function |
| `if (schedBlocked()) return;` — above the queue branch, so it also caught queued lines and annotation-/screenshot-only sends | `submitChat` |
| `schedBlocked()` itself, now that nothing asks | — |
| `#box:disabled { opacity: .6; cursor: not-allowed }` — dead once nothing disables the box | stylesheet |

The one remaining reader is the banner itself, plus `schedLoadTaskRow`, which is now gated on `schedPending.length` rather than on "blocked". The state is renamed for what it does: `schedBlockers` → `schedPending`, `applyComposerBlock`/`applyComposerBlockState` → `applySchedNotice`/`applySchedNoticeState`, `clearComposerBlock` → `clearSchedNotice`. The DOM ids (`#schedblock`, `sb-*`) are left alone — they name the banner element, and renaming them would churn ~60 lines of CSS for nothing.

## The new wording

The old line was `"Blocked — a scheduled message runs in this chat."`, which claims a lock that no longer exists. The half that was always true is the half that stayed:

> **A scheduled message runs in this chat. Anything you send now joins its context.**

(`A repeating message…` for a repeat; `N more after it.` unchanged.) Two clauses — what is coming, and what sending now costs. No apology, no "please", no word for a state the page is not in; a test forbids `Blocked`, `read-only` and `Waiting on` in that line.

## Unchanged

The reason line's position, the task row (ring · `TASK-nnn` · name · state · time — no folder, no session id), the row's click through to the task on the Tasks calendar, `Stop the repeat` / `Cancel this message`, the two-press confirm with its Escape and press-outside disarm, the refusal sentence, and the fail-open behaviour when `/api/schedule` cannot be read.

**The composer still does not move.** Measured on the running server: `#composer-chat` top is `679.77px` with the banner shown, `679.77px` with it hidden, `679.77px` shown again — 0.00px, because `#logwrap` is `#chat`'s only `flex: 1` child and the height comes out of the transcript.

## Verified on a live server (port 2162, both themes)

Driven through CDP against a real pending entry on a real session:

- banner up, `box.disabled === false`, placeholder still `"Reply to Claude…"`, send button enabled — in dark **and** light;
- a message typed with the banner up reaches the send path (`action: "start"` fired with the typed text; `runPython` stubbed so nothing spawned Claude), and the banner stays up afterwards;
- repeat: `Stop the repeat` → `Cancel every future run` armed → Escape disarms → two presses cancels the template, and the banner falls through to the next pending message with the one-off copy;
- one-off: a single press cancels and the banner goes.

## Docs

No new D-number (main is at D307 and another branch is racing for the next one), so both edits are adjustments to existing text:

- **SPEC.md SCH-13c** already ended *"the user is never blocked from their own chat — the machine waits, not the person"* — the banner, the block, and the withdrawal are recorded there, against the rule they belong to.
- **DECISIONS.md D306** gains this as its `(8)`: the pollution reasoning is not retracted (it is what the banner *says*), but it does not buy a dead chat.

Worth knowing: neither file described the block before this PR — the commits that built it did not touch the docs — so this adds the record rather than correcting a contradiction.

## Base

Base is `main` **deliberately**, so this PR also shows the commits of the unmerged `agent/20260816-next` it sits on top of. Only the last commit is mine.

## Gate

`node --check` on the inline script, `bun test` (1264 pass), `npm run build`, `pytest tests/` — 7135 pass. Failures left: `tests/test_calls.py::test_an_abandoned_merge_closes_the_day_s_files` (known), and `tests/test_deploy.py` ×2 + a `test_notebook_kernel.py` error, all environmental (`this server's Python has no pip module`), none of them touching this change.

---

> [!NOTE]
> **High Risk**
> Large scheduling and task-model change (queue semantics, session chaining, many API/UI surfaces) plus behavior change when users chat during pending scheduled runs — context pollution is possible by design.
> 
> **Overview**
> **Scheduled chat (this PR’s stated focus):** Pending messages for the open session still show the composer banner (task row, cancel/stop repeat), but the composer is **no longer disabled** — no placeholder swap, no `submitChat` guard. Copy is a heads-up: *anything you send now joins its context*, not “Blocked”. **D306** and **SPEC SCH-13c** record the withdrawal.
> 
> **Tasks = sessions (bulk of the diff):** Docs and **SPEC** add **D306** / **SCH-13a–15**: one task per Claude session and folder project, **unbounded catch-up** with a visible/cancellable queue (optional `FUSED_RENDER_SCHEDULE_MAX_LATE`), recurring **coalesce** instead of 120s skip, **session_learned** writeback for repeating threads, **`at` vs `ran_at`**, defer while a turn is live, **`run-now`/`resend`**.
> 
> **Frontend:** New **`/api/tasks`** types and queue endpoints; sidebar **Tasks** replaces **Inbox** link; **NewJobModal** — repeat checkbox, title/ask split, past-due allowed, delete task, `session_learned` on edit; **QueueDock** merges the schedule queue into **DownloadManager** (one activity card); **jobsSummary** counts queued vs running.
> 
> Design spec: `docs/superpowers/specs/2026-08-16-tasks-threads-messages-design.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5b6c14c57cebf4bcfc337bd6f7f7d583c7ca3ad3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AkshilVT:
Closing — superseded by a change of direction in review. The branch and its worktree are left in place, so this can be reopened if wanted.

PR #566, opened 2026-08-17
A chip per scheduled message, not one per task per day

AkshilVT:
The calendar drew **one chip per task per day**, anchored at that day's earliest run, with later runs folded into a `+N` badge and named in the popover. That rule was rejected in review:

> *"if I have a task at 3 a.m., then at 5 a.m., and then 7 a.m. As of now, we only show the 3 am task, and when you click on it, we show the 5 and 7 am as a thread. I want all these three to show."*

So: **a chip per scheduled message, at its own time.** Three runs at 3/5/7 draw three chips on three hour lines.

### The popover is untouched

That was the condition on the whole change — *"when you click on them you just see the thread as is no change"*. Every chip of a day still carries that day's full message list, so all three chips of a 3/5/7 task open the identical task-scoped panel: same thread, same `TASK-nnn` pill, same Run now / Re-run footer. Verified by clicking each of the three.

What holds three chips together as one task is the shared colour and the `↻`, unchanged.

### What came off

* **The `+N` badge and its `also 5 AM, 7 AM` accessible-name clause.** With every run placed they could only ever count and name chips already on screen, so they are gone rather than left reading `+0`.
* **`dayTone`** — the fold that made one chip answer for a whole day's outcomes. Each chip wears its own run's tone now, so a 9am that ran clean stays clean beside a 2pm that died. (An hourly day where the morning ran and the afternoon has not now *shows* that split; one chip used to have to pick a single wash.)

`templateId` and `projected` also became per-message, which fixes two fudges the anchor rule needed: a one-off no longer wears the `↻` of a daily rule it merely shares a day with, and a day holding one materialized run plus three ghosts no longer has to choose one outline for all four.

### What an hourly day looks like

Intended volume, and legible. 24 chips, one per hour line, each reading `Hourly sweep ↻ 12:00 AM` at full column width — a 60-minute gap clears `CHIP_MIN`, so nothing lane-splits. Verified in both themes on a rule anchored tomorrow 00:00.

The one thing that did need fixing is the **two-lane split**, which stopped being an edge case the moment a task can land on an hourly rule's own slot. A week column halves to ~78px, which cannot hold `Pull news ↻ 3:00 AM`, and the old `lanes >= 3` threshold kept the clock and ate the title down to `P`. Two lanes now drop the clock in the week range — the hour ruler and the chip's own accessible name still carry it — while the wider 4-day range keeps both until three lanes. A `min-width: 34px` floor means a very dense day overlaps in Google's cascade rather than shrinking every chip to a sliver.

Everything the calendar recently gained is kept: chips place by `at` not `ran_at`, archived tasks draw no chip, the five-word status vocabulary, the task-scoped pill, Run now / Re-run in the footer.

### Spec

`docs/superpowers/specs/2026-08-16-tasks-threads-messages-design.md` §8 argued *for* the old rule and stated its accepted cost ("a 7pm run has no chip at 7pm"). It now describes what the code does, says plainly that the earlier rule was tried and rejected in review, and records how the old objection (one task shown many times a day) was answered rather than overruled. The out-of-scope note on a per-day chip cap says what the new ceiling actually is. `DECISIONS.md` D306 clause (7) got the same treatment.

### Gates

`bun test` (1270 pass), `npm run build`, `npx tsc --noEmit`, `pytest tests/` — the only failures are `tests/test_deploy.py` x2, which fail identically on a stashed tree (this venv has no `pip`), plus the known macOS `test_calls` one.

### Base

**Base is `main` deliberately, and this branch sits on top of the unmerged `agent/20260816-next`**, so the PR shows that branch's commits too. Expected — rebase before merging.

---

> [!NOTE]
> **High Risk**
> Touches scheduling semantics (unbounded catch-up, recurring coalesce, session chaining, defer-on-live-turn), new task/queue APIs, and calendar density—bugs can double-run sessions, drop work, or misplace chips.
> 
> **Overview**
> **Calendar placement** now follows **one chip per scheduled message at its own time** (not one chip per task per day with a `+N` for later runs). The design spec and **D306** record that the old rule was rejected in review; chips of the same task still share colour and `↻`, and the popover/thread behavior is unchanged. Per-run outcome styling replaces day-level folding (`dayTone` / `+N`).
> 
> The diff also lands the broader **task = Claude session** model on the client and in docs: **`GET /api/tasks`**, per-message read/mark-all-read, **`getTasksScheduled`** for the calendar window, queue **`GET /api/schedule/queue`** with cancel APIs, and schedule fields such as **`session_learned`**, **`at` vs `ran_at`**, **`new_task_each_run`**, **`run-now` / `resend`**, and **`max_late_seconds: null`** for unbounded catch-up (**SCH-3b** / **SCH-13** in **SPEC**).
> 
> **UI:** sidebar promotes **Tasks** and drops advertised **Inbox**; **NewJobModal** gets repeat-as-checkbox, optional title, past-due as “run ASAP”, learned-session carry-through on edit, and delete-task for templates. **QueueDock** folds the schedule queue into the single bottom-right **DownloadManager** activity card (one header/count/lifecycle).
> 
> **Note:** Calendar grid implementation (e.g. `ScheduleCalendar` / `taskChips`) may live on the stacked base branch; this diff’s file list is mostly API types, shell composition, and specification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5419760aa69a031227c07982491911c37bf4b754. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AkshilVT:
Closing — superseded by a change of direction in review. The branch and its worktree are left in place, so this can be reopened if wanted.

PR #571, opened 2026-08-17
/apps: lazy-mount live preview iframes near the viewport, offload when scrolled past

vasu2001:
## What

Cards on `/apps` without an authored `preview.png` render the app live in a scaled iframe. `loading="lazy"` only defers the *first* load — once scrolled through, every iframe stayed mounted, so a big workspace accumulated a whole sandboxed page + JS runtime per card.

Each thumb now carries an `IntersectionObserver` (`useNearViewport`) with an 800px vertical margin: the iframe mounts shortly before the card scrolls into view and unmounts once well past, showing the existing monogram placeholder in between.

Key detail: the observer is rooted at the `.apps-page` scroller, not the default document viewport — the document never scrolls, so `rootMargin` expansion on it was silently cancelled by the scroller's own clip (cards only mounted once fully visible; observed live before the fix).

`preview.png` thumbs and monograms are untouched — an `` costs nothing to keep mounted.

Also logs D322 in DECISIONS.md.

## Verification

Verified live via CDP against a 22-app workspace: iframes mount ~800px before entering view (2048 card mounted at +435px below the fold), unmount once well past (−1555px above → unmounted), band-edge cards at −841px stay mounted. `tsc --noEmit` clean, 1414 frontend tests pass.


vasu2001:
bugbot run

PR #604, opened 2026-08-18
Tasks: bigger lanes, lanes that remember, and a list that remembers

AkshilVT:
Stacked on #596 (`agent/20260818-unread-rows`) — retargets to `main` once that merges.

Four things the Tasks page was getting wrong, all of them about the page forgetting or under-showing what the reader asked for.

## Board

- **Lanes open on 20 cards and reveal 20** (was 10/10). The lane scrolls either way, so the window is about how much is rendered, not what fits; ten was two presses to read a busy column. The `Show N more` label is arithmetic over the same constant, so it cannot say ten while twenty arrive.
- **An empty lane starts rolled up; every other lane starts open — Archive included.** Archive was hard-coded closed, which hid a triaged pile behind a 52px rail; an empty lane still took a full 260px column to say nothing. One rule now, `tasks-lib.laneCollapsed`: empty ⇒ rail, otherwise ⇒ column.
- **Your own toggles stick.** The old localStorage key stored *the set of collapsed lanes* — defaults included — which cannot be told apart from a choice, so the first visit froze every lane forever. It now stores a lane → boolean map of **presses only**, under a new key (the old rows are the wrong shape, and reading them as choices would resurrect the freeze). A lane nobody has touched keeps following the rule as it fills and empties.

## List

- **Comes back to the rows you had open, where you left them.** Opening a task's chat leaves the page, and returning gave a fully collapsed list scrolled to the top — three threads out of ninety meant three trips down the same scrollbar. The open task keys and the list's scroll offset are remembered in **sessionStorage** (per tab, per sitting: a week-old offset restored into different rows is a surprise, not a memory).

Three details in there worth knowing:

- A restored row was never *toggled*, so nothing fetched the rest of its thread — it would show the listing's three messages with no button left to ask for the other twenty-three. The restore makes that trip itself, once.
- Rows grow as their threads land, so the target offset is unreachable for the first few frames: it is re-applied every render until the content is tall enough, and the window closes after 3s so a list that can never be that tall stops pinning itself to the bottom.
- A scroll **the reader** made cancels the restore; one this code just made does not.

This is `.tasks-list`'s own `scrollTop`, not the window's — it has nothing to do with, and cannot fight, the chat's `msg-anchor` scroll on the explorer page.

## Tests

New unit tests for the two pure rules (`laneCollapsed`, `parseLaneChoices`, `parseListMemory` — including the junk-in-the-store cases), plus source assertions that the views ask those functions instead of keeping a second copy of the answer. `bun test`: 1639 pass / 0 fail. `npm run build` and `tsc --noEmit` clean.

---

> [!NOTE]
> **Low Risk**
> Client-only Tasks UI state (session/local storage and scroll/expand behavior); no auth or API contract changes. Edge-case bugs could affect scroll position or lane collapse until refresh.
> 
> **Overview**
> Improves the Tasks **Board** and **List** so the UI remembers layout and scroll instead of resetting on every visit.
> 
> **Board:** Lane paging goes from **10 → 20** cards initially and per “Show more”. Collapse behavior moves from a frozen “collapsed lanes” array (Archive always shut) to **`laneCollapsed`**: empty lanes start as rails, non-empty lanes start open, with **per-lane overrides** stored in a new `localStorage` map (`LANE_CHOICE_KEY`) that records only explicit toggles.
> 
> **List:** **`sessionStorage`** (`LIST_MEMORY_KEY`) saves expanded task keys and `.tasks-list` **scrollTop**. Returning from chat restores open rows and scroll; restored expansions trigger a one-time full thread fetch. Scroll restore retries until content height catches up (3s cap), ignores programmatic scroll echoes during restore, clears scroll when a filter empties the list (not on initial empty fetch), and debounces writes.
> 
> **`tasks-lib`:** Adds `parseListMemory`, `parseLaneChoices`, and `laneCollapsed` with defensive parsing. Tests cover the rules plus source assertions that views call the shared helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08a5b6066e4f3425a1763ebf55c9d4364ffca92f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/src/shell/ScheduleTaskViews.tsx
+++ b/frontend/src/shell/ScheduleTaskViews.tsx
@@ -23,7 +23,14 @@
 // one page. Those live in styles/schedule.css; everything this file adds (the
 // accordion, the thread rows, the id chips, the unread dot) is in
 // styles/tasks.css.
-import { Fragment, useEffect, useMemo, useRef, useState } from "react";
+import {
+  Fragment,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import type { DragEvent as ReactDragEvent } from "react";
 import {
   cancelScheduledMessage,
@@ -40,7 +47,9 @@ import { BOARD_COLUMNS } from "./schedule-lib";
 import type { BoardColumn } from "./schedule-lib";
 import {
   EMPTY_FILTERS,
+  EMPTY_LIST_MEMORY,
   LANE_CHOICE_KEY,
+  LIST_MEMORY_KEY,
   UNREAD_LABEL,
   archiveIntent,
   basename,
@@ -69,6 +78,7 @@ import {
   openThreadIntent,
   opensElsewhere,
   parseLaneChoices,
+  parseListMemory,
   projectOptions,
   relativeWhen,
   settleMarkAllRead,
@@ -89,6 +99,7 @@ import {
 import type {
   ArchiveStatus,
   LaneChoices,
+  ListMemory,
   OpenThreadIntent,
   TaskFilters,
   TaskRunIntent,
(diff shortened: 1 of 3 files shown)


AkshilVT:
Stacked on #598 (base `agent/20260818-tasks-nav`), not on main. Both branches rewrite `ScheduleTaskViews.tsx`/`tasks-lib.ts`/`tasks-lib.test.ts`, and merging them as siblings conflicts, so this one rides on top of nav.

**Retarget this PR to `main` once #598 merges.** No rebase should be needed at that point — the nav commits will already be in main's history.

Conflicts resolved during the stack rebase were all additive-vs-additive (two alphabetical import lists, and three appended blocks that both branches placed at the same spot); both intents kept in every case.

One real interaction, and it already works: qol's list-memory restore re-opens rows that were never toggled, so it goes and fetches their full threads through the same `showMore` path nav's chevron-only `toggle` uses (`ScheduleTaskViews.tsx`, the `restoredThreads` effect). Gate on the stacked branch: `npm run build` clean, `bun test` 1680 pass / 0 fail.

PR #651, opened 2026-08-19
Canvas previews are signed off the listing's critical path (D360)

vasu2001:
## Problem

`/canvases` was measured at **2.21s warm / 3.95s cold** on a 7-canvas account before rendering a single card. Breakdown of `GET /api/canvases/list`:

| leg | time |
|---|---|
| spawn python + `import fused` + `fused._env` | 0.35–0.50s |
| `api.list_collections(whose="self", lite=True)` | 0.90s |
| `GET /collection/sign-image` — **one blocking call per canvas with an uploaded preview, in a `for` loop** | 0.93s **× N** |

That last term is the design bug: listing latency scaled **linearly with the account's canvas count**, entirely for decoration. A 20-preview account waits ~18.6s. Nothing on the page needs a thumbnail to be usable — name, UDF count, modified time, cloned pill and the click target are all in hand before the sign loop starts.

## Change

Split along the line the data already draws.

- **`_fused_canvases_list.py`** keeps a public `https` `preview_image_url` inline (free — already in the list payload) and reports only the `fused_uploaded_preview` sentinel as `preview_pending`, with no round trip. Routing the free case through the batch would make the fast path pay for the slow one.
- **`POST /api/canvases/previews`** signs a batch of collection ids via the new **`_fused_canvas_previews.py`** shim on an **8-thread pool** (pure network waits, so the GIL is irrelevant). Headers are generated **once outside the pool** — `api._generate_headers()` may refresh the credential, and N threads racing to refresh one is both wasted work and the shape that trips a token endpoint's rate limit. Ids arrive as JSON on **stdin**, not argv, which has a length limit a large account would reach.
- `POST` rather than a guarded `GET` for one reason: the ids are a list, and a comma-joined query string for a large account runs past what proxies carry. It still takes the same `X-Fused` guard as the other executing endpoints.
- **`Canvases.tsx`** paints cards immediately and fills thumbs in when the batch lands. `fillPreviews` is deliberately **not awaited**, and tracks asked-ids in a **ref, not state** — reading `previews` state there would make `fillPreviews`, and through it `refresh`, a new function on every thumb that landed, and `refresh`'s effect would re-run forever.

Failure postures, deliberately unequal:
- a **failed signature is never an error** — null for that id, and the card keeps its letter monogram, identical to a canvas with no preview;
- a failed **batch** un-marks the asked-set so the next refresh retries, rather than costing thumbs until a reload;
- the **external-CLI path answers `{"previews": {}}`** rather than erroring — that path (`canvas list`) never reports a pending preview, so empty is truthful, not degraded.

`PREVIEWS_MAX_IDS = 500` bounds one request's work — far above any real account, there so a malformed body cannot ask for unbounded signing.

## Measured, same account

| | before | after |
|---|---|---|
| `GET /api/canvases/list` warm | 2.21s | **1.39s** |
| `GET /api/canvases/list` cold | 3.95s | 1.76s |
| previews, 20 ids | ~18.6s (sequential, blocking the page) | **3.37s** (off the critical path) |

Cards paint ~0.8s sooner, and the part that scaled with canvas count no longer blocks anything.

## Tests

`tests/test_canvases.py` — **75 passed** (7 new): pending-preview reporting, batch signing (asserting the shim reads ids from stdin), guard + input validation, the `PREVIEWS_MAX_IDS` cap, the external-CLI empty answer, expired-credentials → 401, and not-signed-in → 409. Frontend `tsc --noEmit` clean.

Full suite: 8194 passed, 9 pre-existing environmental failures (`test_claude_health.py`, `test_calls.py`, `test_fused_cli_export.py::test_the_pin_is_discoverable_at_runtime` — these read installed distribution metadata / probe the `claude` CLI, and none import canvases).

## Not in scope

A **server-side TTL cache over the whole listing** was the other option on the table. It is orthogonal and still worth doing — it makes the repeat visit free while leaving the cold visit's cost where it is, and it composes with this change rather than competing with it.

DECISIONS.md D360 carries the full argument and the rejected alternatives.

---

> [!NOTE]
> **Low Risk**
> Listing UX and a new guarded batch-signing endpoint; no auth, sync, or data-model changes beyond optional `preview_pending` on list entries.
> 
> **Overview**
> **Moves private-bucket canvas preview signing off `GET /api/canvases/list`** so the gallery can render before the control plane finishes N sequential `sign-image` calls.
> 
> The list shim (`_fused_canvases_list.py`) now only returns public `https` preview URLs inline and marks `fused_uploaded_preview` entries with **`preview_pending`** instead of signing them. A new **`POST /api/canvases/previews`** endpoint runs **`_fused_canvas_previews.py`**, which signs ids from stdin on an **8-worker thread pool** with auth headers generated once. **`Canvases.tsx`** paints cards from the list, then **`fillPreviews`** (not awaited) batches pending ids via **`getCanvasPreviews`**, merging signed URLs into separate state and using a ref to avoid refresh loops; logout clears cached thumbs.
> 
> Per-id sign failures return null (monogram fallback); a failed batch retries on the next refresh. External CLI path returns empty previews; **`PREVIEWS_MAX_IDS = 500`** caps batch size. **DECISIONS.md D364** documents the rationale.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84033713397cdf8430864e0e7e1a7d45b0a65f6f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vasu2001:
### Checked against the hosted workbench: there is no batch endpoint to use instead

A reviewer's obvious question about `_fused_canvas_previews.py` is "why fan out per id instead of asking the control plane to sign the batch?" Answer: **no such endpoint exists.** Audited the `application` repo's generated OpenAPI surface (`fused-js/src/api/schema.d.ts`) — every sign route is singular, GET, one id query param: `/v1/collection/sign-image`, `/v1/udf/sign-image`, `/v1/files/sign*`. No plural, no POST-with-id-list.

Nor does the list endpoint hand back signed URLs. `preview_image_url` in the lite payload is the raw DB column, verbatim (`server/server/crud/collection.py:79-84`) — the same three cases this PR now branches on. A public `https://…` URL only exists after an explicit opt-in copy-to-public-bucket flow (`POST /collection/public-preview-image`), which fires on the Git-commit path.

**The hosted workbench pays exactly the same N sign calls we do.** It feels fast for reasons that are all now true of this PR, plus two that aren't yet:

| workbench does | this PR |
|---|---|
| N sign calls run **in parallel** (SWR per card, HTTP/2) | ✅ 8-thread pool |
| cards paint from the list payload; thumbnail fills in later | ✅ |
| skips signing for direct public `http(s)` URLs (`isDirectPublicPreviewUrl`) | ✅ inline, no round trip |
| module-level **negative cache** so a thumbnail-less canvas is asked for once per page load | ~ `previewsAskedRef` does this within a page session; lost on reload |
| — | — |
| SWR 60s dedupe on image keys | ✗ no server-side cache |

Two follow-ups worth their own PR, both now better-founded:

1. **Cache signed URLs server-side, keyed on `collection_id`, ~50min TTL.** `sign-image` presigns with `ExpiresIn=60*60` (`server/server/endpoints/collection.py:1413`) — a one-hour budget, which makes a 50-minute cache safe rather than a guess. Warm loads then do **zero** sign calls, and it subsumes the negative-cache gap above.
2. **Issue the list request directly instead of through the child process** (~0.4s of `import fused` + `fused._env`). `list_collections(whose="self", lite=True)` is just `GET {base_url}/collection/self/lite?skip=0&limit=100` with a bearer token (`fused-py/fused/api/api.py:617-663`) — exactly what the workbench client does. Deliberately **not** done here: it means reimplementing `_generate_headers`' auth in fused-render, against the repo's standing choice to isolate `fused` behind a child process. Worth a decision row of its own, not a quiet change.

Also worth recording, since it bounds what any client-side work can achieve: part of the ~0.9s list call is an **N+1 inside the control plane** — `CollectionLite.udf_ids` iterates a lazy relationship (`server/server/models/collection.py:82-87`) and `get_collections_by_user` uses no eager loading, so `/self/lite` fires one extra SELECT per collection. Not fixable from fused-render; the only lever a caller has is `limit`.

PR #689, opened 2026-08-20
Word-timed matching for claude's spoken annotations (D393)

vasu2001:
Leverages the word timings from #686 (D392) in the claude template's spoken annotations.

**Before**: a walkthrough's transcript was matched to clicks per SEGMENT, and a whisper segment is a sentence or several — the repo's own 7s recording comes back as one span holding three separate comments. So "this button is wrong" *click* "and this chart too" put both halves on the first note and left the second empty.

**Now**: `annRecEnd` sends `fused.ai.transcribe({ path, words: true })`, and `annRecAssign` flattens each segment into timed units before the unchanged nearest-click match — a WORD per unit when the engine timed words, the whole segment when it did not. The sentence splits at the click that interrupted it; the intro rule (everything before the first click seeds the composer as the message's prompt, D378) is the same rule on finer units and just cuts mid-sentence now.

**Degradation is per segment, with no capability probe** — D392 answers `words` best-effort rather than refusing it, so an engine without word timings leaves the key off, indistinguishable here from never asking. A wordless reply, a mixed reply, and a segment whose words are not all timed (falls back whole, rather than dropping the untimed ones) all behave like today. No words are lost on any path.

`words: true` is sent unconditionally: it costs an extra forward pass per decoded window plus a one-time Metal compile, which on seconds of speech does not pay for a conditional, and D392's decode change (hallucination pruning is gated on the flag, so segment COUNT can differ) is irrelevant here — nothing on the page reads counts.

## Tests

`tests/test_claude_annotation_modes.py`: mid-sentence split by word timings, the word-grained intro cut, a mixed reply (worded + wordless + partly-timed segments), and the `words: true` call itself. The pre-existing segment-only tests are the regression check that the old path is untouched.

`pytest tests/ -k "claude or doc_ or ai_runtime"` → 1595 passed; the 7 failures are pre-existing env leaks in `test_claude_health.py` / `test_ai_metrics.py` (a real `claude` binary on this machine's PATH), unrelated to this diff.

## Not covered by tests

Real mic capture and real transcription, as ever in this suite — needs a manual smoke: record a walkthrough on a Mac with MLX Whisper, click mid-sentence, confirm each note holds only what was said around its own click.


vasu2001:
Heads-up on the D-number: open PR #687 also mints a **D392** row (which is already taken on main by #686), so it will have to renumber — most likely into the D393 this PR uses. Whichever of #687/#689 merges second should bump its row (the duplicate-id doc test will catch it either way).

PR #700, opened 2026-08-20
Native capture: fused.capture.screen / audio / screenshot (D409, D410, SPEC §45)

vasu2001:
`fused.capture` — the page can record the screen, record the microphone, and grab a still, natively.

```js
const src = await fused.capture.sources();     // never prompts; hide the button if !src.video.available
const rec = await fused.capture.screen({ audio: "both", rect: [0,0,1200,800], onTick });
const out = await rec.stop();                  // {path, url, mime, seconds, bytes}
const words = await fused.ai.transcribe({ path: out.path, words: true });

await fused.capture.audio({ path: "note.m4a" });
await fused.capture.screenshot({ rect: [0,0,600,400] });
```

### Why native, when a page can already call `getDisplayMedia`

- **System audio** — a browser cannot capture it on macOS at all.
- **No picker per recording** — the grant belongs to FusedRender.app (the server runs in-process there), not to Chrome.
- **The result is a file whose path is known before the recording stops**, so `fused.ai.transcribe({path})` is the next line instead of a blob round-trip through `writeFile`.
- **It survives the page**: a recording is a server-owned job row, so navigating away does not end it (`list()` / `attach(id)` get the handle back).

### Shape

`screen()`/`audio()` resolve **when the recording is running** with a handle `{id, path, url, state, stop(), cancel()}` — a recording is a session of unknown length, so the same idiom as `trackJob`/`watchJob`. `screenshot()` has no handle and no row: it is milliseconds.

**Three endings, one of which destroys the file.** `stop()` keeps it, `cancel()` and the manager's ✕ delete it, and hitting `maxSeconds` (default 30 min) is a **stop** — a page can be closed mid-recording, and then the ✕ is the only control left, so a cap that discarded would leave no ending that kept the file. `atexit` finalises the rest: a .mov with no `moov` atom does not play.

### The two Apple APIs, and why only one of them survived

`SCRecordingOutput` writes the .mov and `AVAudioRecorder` the .m4a — **no sample buffer passes through Python**, which is why the backend is ~300 lines and a 90-minute recording costs no CPU here.

`AVCaptureSession` + `AVCaptureAudioFileOutput` was the first choice (it takes a device id) and **cannot work in this app**: `stopRecording` marshals with `performSelector:onThread:waitUntilDone:YES` and blocks until a run loop runs the work. `start` and `stop` are two HTTP requests on two different uvicorn worker threads, and funnelling every call through one owned run-loop thread did *not* fix it — which identified the target as the main thread. Sampled stack: `-[AVCaptureAudioFileOutput stopRecording]` → `-[_NSThreadPerformInfo wait]`, pinned indefinitely, with a perfectly valid file already on disk. A packaged app has a main run loop (rumps); a source install has none, so that API would have worked in the DMG and hung for anyone running from a checkout. `AVAudioRecorder` has no affinity — 5 ms cross-thread stop — and the whole thread mechanism was deleted with it.

Its one cost is **refused rather than hidden**: audio-only records the system's current input, so `capture.audio({device})` rejects and names where a microphone *can* be chosen (a screen recording's `audio: "mic"`).

`/usr/bin/screencapture` was rejected on capability first, purity second: no system audio, no audio-only, no stop that is not a signal.

### Honest about the machine

- **The probe never prompts.** `CGPreflightScreenCaptureAccess` + `AVCaptureDevice.authorizationStatus` for permission, `CGGetActiveDisplayList` for displays — never `SCShareableContent`, which raises the dialog. The dialog rides the first real capture.
- **The floor is macOS 15** (`SCRecordingOutput`; 14 for the still) against `LSMinimumSystemVersion` 11.0, so a 12–14 Mac gets `available: false, reason: "needs macOS 15"`. Windows/Linux the same shape with their own reason, and `start` is a 409 naming the machine — never a 500, and never an absent route, because a page must be able to ask. No `via` field: one implementation, nothing to branch on.
- Retina was a bug found by looking: `SCDisplay.width` and `CGDisplayPixelsWide` are both **points** (1800 vs 3600 real pixels here), so the first recordings came out half-size and soft. The scale comes from the display mode.

### Packaging

`NSMicrophoneUsageDescription` (an app that touches the mic without it is killed, not prompted), `com.apple.security.device.audio-input` in the hardened-runtime entitlements (without it the mic is refused whatever TCC says), and both pyobjc wheels in core deps with a darwin marker — **3 MB installed with transitives**, measured against D276's budget before being added.

### Verification

Real captures on macOS 26.6, through the endpoints:

| what | result |
|---|---|
| screen, `audio: "both"`, 800×600 rect | h264 **1600×1200** + one mixed aac stream, 4.13 s |
| microphone | valid aac, 2.99 s |
| still | 3600×2338 png |
| start → active → job row → stop → re-stop | 200 / listed / `unit: "s"`, `owner: server`, cancellable / 200 / 404 |
| cancel | file deleted, row `cancelled` |

`tests/test_capture.py` (29 tests) stubs the backend and covers what decides behaviour: the refusals that must land before a microphone turns on, where the file goes, the job row, and all three endings. Full suite: 8639 passed; the 20 failures are pre-existing and environmental on this machine (missing `rasterio` after D276, a real `claude` binary on PATH, a `/proc/self/fd` read, one xdist ordering flake) — none touch capture.

### Deliberately not in v1

Window-id capture, pause/resume, webcam, multi-track output, live frames to the page. The bridge is a forever contract, so the surface is small on purpose.

`screenshot()` is in, and it is what makes two existing problems solvable — the hand-rolled tab-capture crop in `claude/template.html` and D349's cross-origin-pane limitation both exist because a browser still needs a readable document and a share prompt. **Nothing in the claude template is rewired here**; that is a separate change to a page with its own tests.

### Still to smoke by hand

The packaged .app: the first-capture TCC prompt (attributed to FusedRender, not a terminal), the mic prompt with the new usage string, and a recording that outlives the tab that started it.


vasu2001:
Applied a fable review pass — four real bugs, three cuts (8b47908).

**The one worth reading:** `atexit.register(stop_all)` is **dead code in the packaged app**. Every quit surface there ends in `os._exit` (DM-9 — reaching `__cxa_finalize` aborts on a native extension's destructor), which runs no atexit handler. So a quit mid-recording left an unplayable .mov behind a row that said "recording" — and `test_stop_all_finalises_everything_on_the_way_out` passed the whole time, because it called `stop_all()` directly rather than exercising the registration. Now a `"capture"` rung in `app.quit_teardown` (before the unmounts: a recording writing under a mount holds it busy) plus an ASGI shutdown handler for the plain server, with a test that reads both call sites.

Also fixed:
- the client memoized the settled stop **result** instead of the promise → a double-clicked stop button fired a second request that 404'd as an unhandled rejection;
- `recordingOutput_didFailWithError_` stored an error nothing read until `stop()` → a stream dying at minute two ticked "Recording" for the rest of its cap over a file nothing was writing (the watchdog asks the backend each tick now, best-effort so a second platform need not implement the hook);
- `_settle` returned the same nothing whether the file went quiet or the deadline passed → a still-growing movie could be handed back as done.

Cut from the public surface while it is still free to:
- **`onTick`** — a recording publishes a job row and `fused.watchJob(jobId)` is the documented way to read one; the whole watcher block went with it.
- **`format` on the still** — the output **filename** picks png vs jpeg, so a `path` and a `format` can no longer disagree. Verified with `file`: `shot.jpg` really holds JPEG bytes, an unknown extension is a 400.
- **audio-only's `source`** — it ran the shared validator (which names 'mic', 'system' and 'both' as valid) and then refused two of them a line later. One check, one message.

`list()`/`attach()` were kept against the same pressure: after a reload they are the only route back to a live recording's `stop()`, and ✕ discards — without them a reload can only destroy what it finds.

Re-verified after the changes: screen + `audio: "both"` → h264 1600×1200 + aac, 4.07 s; mic → valid 2.07 s aac; png 3600×2338 and a real JPEG by filename. Suite 8643 passed, same 19 pre-existing environmental failures.

vasu2001:
### Windows and Linux backends (D407)

`fused.capture` now works on all three platforms, through one API with no field telling a page which backend served it.

**Split by capability, not by platform** — because the Windows half is a constraint, not a preference. There is no OS API a non-packaged Python process can reach that will write a movie with system audio in it: `AppRecordingManager` needs MSIX identity, `Windows.Graphics.Capture` hands out D3D surfaces with no muxer, Media Foundation has no screen source, and ffmpeg's Windows inputs are dshow/gdigrab/vfwcap — **no WASAPI indev**, so no loopback without a third-party driver. Chromium already does what a native recorder would (WGC + WASAPI loopback; the PipeWire portal on Linux), hardware-encoded.

| | mechanism |
|---|---|
| still, Windows | GDI `BitBlt`, DPI-aware, per-monitor, optional cursor composite |
| still, Linux | `org.freedesktop.portal.Screenshot` over dbus-fast — the only interface that works on Wayland *and* X11 on every desktop |
| recording, both | the page's own `MediaRecorder` → chunks over a WebSocket → `capture/_sink.py` writes the file |

One sink for both platforms is what deleted the expensive half of the plan: no portal ScreenCast state machine, no PipeWire fd hand-off, no `gst-launch-1.0` detection, no gst-libav container question.

**What is unchanged:** the path is decided at start (CP-2), it is a job row with the same three endings, `fused.ai.transcribe({path})` is still the next line.

**What differs is refused with a pointer, never ignored:** `display`/`rect`/`cursor` on `screen()` (the share picker owns the region), `display`/`cursor` on a Linux still (the portal has neither), a `path` whose extension contradicts the container. Going the other way, `device` now works on `audio()` there — it does not on macOS.

**CP-8's "no `via` field" survived a second implementation.** `transport`, `streamToken` and `client` exist on the wire so `runtime.js` can open the socket and merge a browser probe; all three are deleted before a page sees anything, with a test that fails if one leaks into the handle.

**Four load-bearing orderings** on the streamed path: the share picker opens before the server allocates anything (a cancelled picker leaves no row and no file); chunks go through one promise chain (`Blob.arrayBuffer()` is async — two chunks sent out of order are a corrupt container); the socket opens before `recorder.start()`; and an `eos` round-trip precedes the stop request, since the stop travels on another connection and would otherwise close the file ahead of the last timeslice.

**A closed socket is an ending, not an error.** The encoder lives in the page, so a reload ends the recording — the one real regression against macOS. It fails safely: fragmented mp4 and WebM are playable *as written*, so there is no `moov` atom to lose; the file is kept, the row is marked done, and `list()` stops showing a recording that is not happening. The handle is built in the topmost same-origin window (D72's climb), so iframe navigation and switching files do not end one.

**One bug found by testing, invisible without it:** returning from a websocket endpoint does not close the socket, so the byte-ceiling break left the page encoding into a file nothing was reading. The explicit `ws.close()` in the `finally` is what makes every server-side ending visible to the page.

**Rejected:** bundling LGPL ffmpeg (40–60 MB and *still* no system audio); WGC through `MediaStreamSource` (per-frame Python + a D3D copy per frame for hours); a native helper .exe (a new build/signing pipeline); PyAV with `x11grab` (frames through Python, dead on Wayland); enumerating displays on Wayland (no client may without prompting — CP-7).

**Dependencies:** `pillow` (both) and `dbus-fast` (Linux) promoted from the desktop extras into core with markers. Both already arrive in the packaged builds for tray icons; leaving them there reproduces the mistake the `pyobjc-framework-Cocoa` comment describes — a source install running the same pages and reporting `screenshot.available: false` forever. dbus-fast is pure Python with manylinux wheels: no GTK, no PyGObject, no system package.

**Not verified on hardware** — there is no Windows box and no Linux desktop in this environment, and the ctypes calls, the portal round trip and the browser's own behaviour are the parts with no test. So the pure halves are split out and run on every platform (monitor union, rect offsetting, the portal's three reply endings, its file URI), and the sink is tested **end to end through the real WebSocket**, where a test can simply be the browser: 38 new tests covering the token, the `eos` ordering, both failure clocks, the byte ceiling, the cap and the closed-socket ending.

Suite: 8829 passed. The 9 failures are pre-existing and environmental — 7 in the `claude`-on-PATH family, one `/proc/self/fd` read, and one order-dependent `test_ai_worker_base` case that passes in isolation.

Smokes that need a human: the Chromium share picker on Windows, the portal picker on a Wayland compositor, and a mic grant in each browser.

vasu2001:
### Follow-up: a token leak, and the bridge now has a test that runs it

Three fixes on top of the Windows/Linux commit.

**1. `streamToken` was reachable on an unguarded GET.** `_Session.public()` carried `transport` and `streamToken` whenever the handle had them, and `active()` feeds `GET /api/capture` — which is unguarded on purpose, since a probe has to be readable while a page draws a record button. So every live recording's bearer token was available to anything that could make a GET (the guard given away), and `transport` on a `list()` row was exactly the `via` field CP-8 forbids. Both fields moved to `opening()`, used by the start reply alone; nothing else needs them, because `attach()` re-adopts a recording's *controls*, never its encoder. My leak test read only the handle `runtime.js` builds, so it saw none of this — it reads the GET's rows now too. This invalidates the "with a test that fails if one leaks" claim in my previous comment; it is true as of this commit.

**2. Nothing in the suite executed the bridge.** Every other check reads runtime.js as text, so a stray brace would ship a file that breaks every page in the app rather than one feature. Two additions:
- `node --check` as a test;
- `tests/_capture_bridge_probe.mjs`, which runs runtime.js **verbatim** against a DOM and media stub — the `_git_view_probe.mjs` pattern, for the same reason that file exists.

The probe pins what only running it can. The observed order is now asserted:

```
picker → POST /start → ws.open → recorder.start:1000 → ws.chunk
       → recorder.stop → ws.chunk → ws.eos → POST /stop → ws.close
```

which is the four load-bearing orderings in one line, plus chunks arriving as `[3, 1]` (the order produced — `Blob.arrayBuffer()` is async, and two swapped clusters are a corrupt container), plus `handle.leaks == []` and `client` stripped.

**3. Two Windows corrections in code no test here can reach.** The `EnumDisplayMonitors` callback declared LPARAM as `c_double` — pointer-sized and integral; it limps along on x64 because the value is unused, and is still the wrong ABI. And `_region`'s comment claimed `rect` arrives in points "like the other backends" while the code scales nothing under `PER_MONITOR_AWARE_V2`. The rule is now stated once and is true on all three: **a `rect` is in the units the `displays` entry it applies to reports** — points on macOS, physical pixels on Windows.

Suite: 8836 passed. Failures unchanged and pre-existing — 7 in the `claude`-on-PATH family, one `/proc/self/fd` read, and one xdist-only collection error in `test_fs_raw_bearer_proxy.py` that passes in isolation.

PR #706, opened 2026-08-21
llamacpp-text: GGUF text generation with Vulkan GPU offload and any-repo model resolution

iamsdas:
## Summary

- **Adds two text-generation runners that read GGUF quantizations through `llama-cpp-python`:** `llamacpp-text` (`llama.cpp (CPU)`, every platform) and `llamacpp-text-vulkan` (`llama.cpp (Vulkan)`, Linux x86_64 + Windows AMD64). Non-Apple machines previously had **no quantized text path at all** — `transformers-text` loads unquantized safetensors on a CPU torch wheel. `unsloth/Qwen3.5-9B-GGUF` Q4_K_M is **5.68 GB against 19.33 GB** for the bf16 `Qwen/Qwen3.5-9B`: newer and larger for under a third of the download.
- **Measured on real hardware, llama.cpp beats `transformers-text` on both hardware tiers, on every axis at once** — 4.2× throughput on GPU, 2.4× on CPU, at ~⅓ the download and ~⅓ the memory. Full table below. This is the x86 re-measurement `formats.UNLOADABLE_QUANT` explicitly asks for, and it makes removing `transformers-text` a follow-up worth doing (stacked branch, not this PR).
- **The Vulkan variant is the only GPU text path here that needs no vendor SDK** — it covers AMD, Intel and NVIDIA alike. `transformers-text-cuda` needs CUDA; `transformers-text-rocm` needs ROCm and Linux.
- **Registered below all three `transformers-text` rows, so `auto` resolution never reaches these rows.** Opt-in on purpose, and the reason is packaging rather than performance: the wheels come from the maintainer's own index, not PyPI.
- **The pin (`==0.3.29`) was chosen by a per-platform zip-integrity audit, not by "newest version resolves".** Of all 16 macOS arm64 wheels on the maintainer's index, only **4** (0.3.18, 0.3.24, 0.3.26, 0.3.29) pass `zipfile.testzip()`. Every failing wheel's sha256 matched its own published GitHub digest, so re-downloading never helps — upstream publishes the bad bytes ([#1650](https://github.com/abetlen/llama-cpp-python/issues/1650), open since Aug 2024). [#2352](https://github.com/abetlen/llama-cpp-python/issues/2352)'s claim that 0.3.30 is unaffected was checked directly and is **false**. 0.3.29 is the newest release where all four shipped platform tags pass; both vulkan-index wheels re-audited **PASS**.
- **Any GGUF repo resolves, not just a curated list (D407).** `formats.pick_gguf_file()` maps a bare Hub repo id to exactly one `.gguf` by excluding-by-shape, ranking by quant suffix, then refusing rather than guessing when two unranked candidates remain — a wrong pick is a multi-gigabyte download of the wrong file, not a `FileNotFoundError`. Curated `(repo, filename)` pairs still resolve exactly as before. Deliberately **no `repo:Q4_K_M` id grammar**, which would touch every page, preference and cache tag that treats a model id as a Hub repo id verbatim.
- **GPU offload actually happens.** `llama_text.load()` previously never passed `n_gpu_layers` and llama-cpp-python defaults it to `0`, so both runners ran **100% on CPU** — the Vulkan variant would have installed a 182 MB GPU wheel and accelerated nothing, and the macOS wheel's Metal backend was never engaged. No test caught it because the hardcoded `device="cpu"` was, under those conditions, correct. Offload is now decided by `llama_cpp.llama_supports_gpu_offload()` at call time and sized empirically, with `worker_base.STATE["device"]` reporting `cpu` / `gpu` / `gpu (partial)`.
- **Engine names are qualified by HARDWARE, never by format.** The CPU row was briefly `llama.cpp (GGUF)`, which qualified the wrong axis — both rows read GGUF through the same `runners/llama_text.py`, so the format cannot tell them apart while the wheel index can. Both names now carry the hardware, matching the precedent `transformers-text` set with `(CPU)`; a test groups rows by label stem and rejects any name that advertises a format its sibling also reads.
- **The Local card's engine tag names the engine FAMILY, hardware stripped** (`llama.cpp`, `Diffusers`), via an explicit `Runner.family_label` on all 13 rows. A tag on a Local card is a *format* claim about a file on disk, and all three Diffusers rows read the identical safetensors — so `(ROCm)` there answered nothing and leaked which machine happened to be looking. The hardware-qualified name stays on the tag's hover and aria-label.
- **Amends SPEC AI-2a, AI-2c and AI-11** (which had refused llama.cpp outright as sdist-only — true of PyPI, not of the maintainer's index) and widens `test_ai_runner_deps.py`'s wheels-only guard into a general rule: any non-PyPI `[[tool.uv.index]]` a runner declares must carry `explicit = true`.
- **Bug fixed in passing:** `_llamacpp_platform()` consulted `platform.system()` only, so Windows ARM64 and non-x86_64/aarch64/riscv64 Linux were advertised as available despite no matching wheel. Now checked per-OS against the real published tag set.

## How resolution and offload work

```mermaid
flowchart TD
    A[Model id] --> B{Key in GGUF_RECIPES?}
    B -->|yes| C[Curated repo + filename]
    B -->|no| D[Bare Hub repo id]
    D --> E[pick_gguf_file: exclude by shape, rank by quant, refuse if ambiguous]
    C --> F[Fetch exactly one .gguf]
    E --> F
    F --> G[Chat template read from the file's own metadata]
    G --> H{llama_supports_gpu_offload}
    H -->|false| I[n_gpu_layers=0, device=cpu]
    H -->|true| J[gguf_block_count reads layer count]
    J --> K[Try full, then 2/3, then 1/3, then 0]
    K -->|construction succeeds| L[device = gpu or gpu partial]
    K -->|last candidate fails| M[raise: a real error, not a sizing problem]
```

The single-file fetch is a fact about the curated repos rather than a design choice: `unsloth/Qwen3.5-*-GGUF` ship no root `tokenizer_config.json` or `config.json` at all — vocabulary and chat template live inside the GGUF, which `llama_cpp.Llama.metadata` exposes at load time.

Sizing is empirical because it has to be: llama.cpp clamps `n_gpu_layers` to the model's layer count and **never to available VRAM**, and the wheel exposes no free-VRAM query at all (no binding for `ggml_backend_dev_memory`). `llama_supports_gpu_offload()` resolves to `ggml_backend_dev_by_type(GPU/IGPU) != nullptr` — a query against ggml's live device registry, not a compile-time flag — which keeps the shared `llama_text.py` variant-agnostic and covers Metal on Apple Silicon with no Apple-specific code.

`pick_gguf_file`'s auxiliary-exclusion list was **wrong on first attempt and fixed by measurement**. A scan of ~200 real `text-generation` + `gguf` repos (2637 filenames) found `mtp-` anchored was incorrect (the real form is `*-mtp.gguf`), `projector` had to be added (an mmproj-shaped file containing no "mmproj"), and `specul`/`speculative` were tried and rejected as false-positive on a real fine-tune name. mmproj files genuinely appear under `pipeline_tag: text-generation`, so this exclusion is load-bearing — without the scan the picker would have offered multimodal projectors as chat models.

## Why two runners rather than one

The Vulkan wheel bundles `libggml-cpu` alongside the Vulkan backend, so a machine with the loader but no GPU degrades to CPU compute. But the degradation is neither free nor total:

| Missing piece | Behaviour |
|---|---|
| Vulkan **driver ICD** (no GPU registered) | Falls back to CPU compute — works, just no acceleration |
| Vulkan **loader** (`libvulkan.so.1` / `vulkan-1.dll`) | `import llama_cpp` fails outright |

That second row is why the CPU variant stays. `libggml-vulkan.so` declares `DT_NEEDED libvulkan.so.1`, and `libggml.so`/`libllama.so` need `libggml-vulkan.so.0` in turn — a hard ELF link resolved at load, not a `dlopen`. The Vulkan wheel is also ~8× the CPU wheel's size (181.77 MB vs ~23 MB on Linux), so serving it to loader-less machines costs bandwidth for nothing. `registry._vulkan()` therefore refuses on a missing loader (hard failure) *and* on a missing ICD (wasted download, the same reasoning `_cuda`/`_rocm` already use), gated to exactly the two tags the index publishes at 0.3.29.

Platform gate for the CPU row: Windows, Linux, and Apple Silicon macOS. **Intel macOS is excluded** because the wheel index publishes no macOS x86_64 tag at all — `uv sync` would have nothing to install behind a Load button.

## Usage

Opt-in via the Engines tab, or by naming a GGUF directly:

```js
await fused.ai("Summarise this table", { model: "Qwen3.5-9B-Q4_K_M.gguf" });
await fused.ai("Summarise this table", { model: "unsloth/Qwen3.5-4B-GGUF" });
```

A curated id is the GGUF's own filename, treated as opaque and never parsed for structure; a bare repo id resolves through `pick_gguf_file`. Hub search on the Discover tab now populates this engine — `Runner.hub_filter_tags` ANDs `filter=gguf` into the query so a repo yielding no usable GGUF is dropped rather than offered with a Load button that cannot work. Verified against the live Hub API: `library=gguf` is a **trap** (returns `facebook/opt-125m` and other mixed `library_name` rows), while `expand[]=siblings` returns the full filename list in the list response for free — which is what makes resolution at row-construction time cheap.

## Benchmarks — Linux + AMD, measured

Radeon RX 9060 XT (RADV GFX1200, 16304 MiB), Ryzen 5 9600X, 30 GB RAM, Linux x86_64. Qwen3.5-4B, identical prompt, 128-token cap, runs back-to-back and uncontended:

| engine | build | download | load | first token | tok/s | peak RSS |
|---|---|---|---|---|---|---|
| **llama.cpp** | **Vulkan GPU (Q5_K_M)** | **2.93 GB** | **1.6 s** | **0.14 s** | **53.4** | **3.14 GB** |
| transformers | ROCm GPU (bf16) | 8.7 GB | 8.3 s | 1.84 s | 12.6 | 9.18 GB |
| **llama.cpp** | **CPU (Q5_K_M)** | **2.93 GB** | **1.3 s** | **0.71 s** | **6.4** | **3.44 GB** |
| transformers | CPU (bf16) | 8.7 GB | 5.9 s | 2.9 s | 2.7 | 8.37 GB |

All four produced correct, fluent prose, so this is like-for-like on quality. The CPU-to-CPU row is the fair one to read first — no GPU advantage on either side, and llama.cpp still wins by 2.4× on speed, 3× on download and 2.4× on memory. `transformers-text-rocm`'s availability gate was checked and is **honest** on RDNA4 (torch 2.13.0+rocm7.1 reports `cuda_avail True` on gfx1200 and genuinely generated on the GPU) — transformers got a fair fight and lost.

## GPU verification — now done on Linux + AMD

Everything in the offload path was originally verified by reading llama.cpp and llama-cpp-python source, with none of it run on a GPU. Closed on the hardware above:

- [x] `import llama_cpp` from the Vulkan wheel succeeds and enumerates the RX 9060 XT
- [x] `llama_supports_gpu_offload()` returns `True` with a working driver, and `False` with the loader present but the ICD hidden (`VK_DRIVER_FILES=/nonexistent.json`) — both halves of `_vulkan()`'s documented reasoning reproduce
- [x] Both llama.cpp rows report available through the registry gate
- [x] `Qwen3.5-4B-Q5_K_M` loads in 0.9 s, `block_count` 32, schedule `(32, 21, 10, 0)`, reported device `gpu`
- [x] Generation is coherent — the first real GGUF generated through this runner's own code path on any platform
- [ ] **The backoff does NOT engage on AMD — see below.** This box cannot be checked.
- [ ] The `win_amd64` wheel's import on a clean Windows box with only the stock MSVC runtime remains untested.

Repro, for anyone else with a GPU — note the command previously given here was broken (`--index-url` *replaces* PyPI, so every transitive dependency vanishes and it fails with `Because typing-extensions was not found in the package registry`). Use the mechanism the runner actually declares:

```bash
cd fused_render/ai/runners/llamacpp_text_vulkan && uv sync
.venv/bin/python -c "import llama_cpp; print(llama_cpp.llama_supports_gpu_offload())"
```

For a dev environment, `uv sync --extra dev --extra bundled` — **not** `--all-extras`, which fails on Linux with `FileNotFoundError: '/usr/bin/sw_vers'` because the `app` extra pulls the macOS-only `pyobjc-framework-webkit`.

## Open finding: the offload backoff never engages on AMD

`_offload_schedule()` exists so an over-commit degrades to partial offload rather than failing the Load button. **On Linux + AMD it never steps.** With 15960 MiB of 16304 MiB already held, `load()` succeeded three consecutive times, each reporting `device='gpu'` with all 32 layers, taking usage to 16222 MiB. The loop catches `Exception` per candidate — but amdgpu/radv appears to satisfy an over-commit by **evicting other clients' buffers** rather than returning an allocation failure, so there is no exception to catch and the first candidate always "succeeds".

The consequence is not theoretical: the evicted client was the compositor, and the desktop session crashed. That is precisely the failure this schedule was written to prevent, and on this stack it is the failure the schedule *produces*, because a successful-looking full offload is indistinguishable from a safe one. `registry.py`'s `diffusers-image-rocm` note already carries a kernel-log-backed "shared ring" desktop-stability warning, so the hazard is known in-tree for the image path and the text path needs the same treatment.

Deliberately **not** re-tested: the card drives the display, so reproducing means crashing the session again. It should be confirmed on a headless GPU before anyone designs the fix, and a fix probably cannot be reactive — it likely needs to size the offload against *free* VRAM up front rather than discover the ceiling by hitting it.

This is a real limit on the claim above, and it is why these rows must stay opt-in.

## Test Plan

- [x] `tests/test_ai_llamacpp_worker.py` — recipe resolution, uncurated repo resolution, locally-cached GGUF discovery, chat rendering against the embedded template, streaming, BOS handling
- [x] `tests/test_ai_formats.py` — GGUF architecture detection (`gguf_architecture` / `is_text_gguf`) against llama.cpp's own `LLM_ARCH_NAMES` set; `pick_gguf_file` ranking, shape exclusions, ambiguity refusal, `UD-` tiering
- [x] `tests/test_ai_runtime.py` — registry ordering (`auto` must not move), the exhaustive runner-code set, the platform gate matching the published wheel tags, the hardware-vs-format naming rule, and `family_label` on every row
- [x] `tests/test_ai_models_api.py` — the engine payload carries the family name beside the hardware one, at **both** router return sites
- [x] `tests/test_hub_models.py` — filter-tag ANDing, sibling expansion, the new drop reason
- [x] `tests/test_ai_runner_deps.py` — the widened `explicit = true` guard
- [x] `bun test` (2062 pass) and `bunx tsc --noEmit` clean
- [x] Full suite on Linux: **8920 passed**. The 21 failures were each traced and none belong to this branch — 18 reproduce identically on `origin/main` (a real `claude` binary present, plus filesystem-dependent `browse_mount`/`index_scan` tests), and 3 are repo-wide source sweeps walking into `runners/*/.venv/site-packages`, which fail in any checkout where a runner venv exists and pass on a clean worktree. That last one is a **pre-existing bug unrelated to this PR** (the app syncs runner venvs in place, so any developer who has loaded a local model sees it; CI never does) and is being fixed separately.
- [x] Wheel integrity audited for all four shipped platform tags of 0.3.29; the vendored llama.cpp (`f05cf467`) confirmed to carry `qwen35`/`qwen35moe` in source and via `strings` on the installed `libllama.dylib`

**Note on CI:** the last completed run on this branch (`0060bd51`, before the naming commits) failed only on `test-python-windows` → `test_rc_cancellation.py::test_stat_timeout_stops_the_job_and_reports_indeterminate` (`assert 0 == 1`) — a job-cancellation timing test with no connection to this PR, and a likely Windows flake.

### Still unverified

- Whether GPU driver installers actually place `vulkan-1.dll` in `System32` on real Windows machines. Stated as a hint in the code, mirroring `NVCUDA_DLL`'s own documented caveat.
- Windows loader/ICD paths — the Linux ones are now confirmed on hardware; the Windows equivalents are verified only against Khronos's `LoaderDriverInterface.md`.

## Follow-up (not this PR)

Given the benchmark, `transformers-text` and its two hardware variants no longer earn their place: llama.cpp is faster on both tiers at a third of the download and a third of the memory, and offering two engines that do the same thing worse is not a feature. Removal is on a stacked branch, since it depends on these rows existing — stripping transformers before this lands would leave Linux with no local text engine at all. Recorded there as D408, along with the two accepted costs: non-PyPI wheel provenance, and `llamacpp-text-vulkan` becoming the only GPU text path on Linux/AMD while the backoff finding above is open.


iamsdas:
## Verified on Linux + AMD — the unchecked boxes are now closed, with one real finding against the safety claim

This PR's description says "None of it has been run on a GPU — this was developed on a Mac with no discrete GPU." I ran it on one.

**Hardware:** Linux x86_64, Ryzen 5 9600X, 30 GB RAM, Radeon RX 9060 XT (RADV GFX1200, 16304 MiB VRAM), `llama-cpp-python==0.3.29` from the `whl/vulkan` index.

| Check | Result |
|---|---|
| `import llama_cpp` from the Vulkan wheel | OK — enumerates the RX 9060 XT |
| `llama_supports_gpu_offload()` with the driver present | `True` |
| Same, loader present but ICD hidden (`VK_DRIVER_FILES=/nonexistent.json`) | imports fine, returns `False` — exactly the split `_vulkan()`'s comment predicts |
| Registry gate | both llama.cpp rows report available |
| Load `Qwen3.5-4B-Q5_K_M` | 0.9 s, `block_count` 32, schedule `(32, 21, 10, 0)`, `device='gpu'` |
| Generation coherence | correct, fluent prose — the first real GGUF generated through `runners/llama_text.py` on any platform |
| Throughput | **53.4 tok/s**, first token 0.14 s, peak RSS 3.14 GB |

The dependency chain in `llamacpp_text_vulkan/pyproject.toml` is accurate as written: a missing *loader* is a hard `import llama_cpp` failure, while a missing *ICD* silently falls back to the bundled CPU backend. Both halves reproduce.

---

### Finding 1 — the offload backoff never engages on AMD, and the safety claim does not hold here

This is the one that matters. The PR argues that an over-commit degrades to partial offload rather than failing the Load button, and `_offload_schedule()` exists to deliver that by stepping `32 → 21 → 10 → 0`.

**It never stepped.** With 15960 MiB of 16304 MiB already held by other allocations, `load()` succeeded **three consecutive times**, each reporting `device='gpu'` with all 32 layers on the card, pushing usage to 16222 MiB. The backoff loop catches `Exception` per candidate — but amdgpu/radv appears to satisfy an over-commit by **evicting other clients' buffers** rather than returning an allocation failure, so there is no exception to catch and the first candidate always "succeeds."

The consequence is not theoretical: the client whose buffers got evicted was the compositor, and **my Hyprland session froze and the desktop crashed.** That is the failure mode this schedule was written to prevent, and on Linux + AMD it is the failure mode the schedule *produces*, because a successful-looking full offload is indistinguishable from a safe one.

Worth noting `registry.py`'s own `diffusers-image-rocm` note already carries a "shared ring" desktop-stability warning backed by a kernel log — so this hazard is known in-tree for the image path, and the text path now needs the same treatment.

I have **not** re-tested this, deliberately: the card drives the display, and reproducing it means crashing the session again. It should be confirmed on a headless GPU before anyone designs the fix. A fix probably cannot be reactive — it likely needs to size the offload against *free* VRAM up front rather than discover the ceiling by hitting it.

### Finding 2 — the repro command in the PR description is broken

```
uv pip install --index-url https://.../whl/vulkan llama-cpp-python==0.3.29
```

fails with `Because typing-extensions was not found in the package registry`. `--index-url` *replaces* PyPI rather than adding to it, so every transitive dependency vanishes. The mechanism the runner actually uses is correct — `[[tool.uv.index]]` + `explicit = true`, which is precisely why that construction is right — so the fix is just to tell readers to `uv sync` in the runner folder:

```
cd fused_render/ai/runners/llamacpp_text_vulkan && uv sync
```

### Finding 3 — `uv sync --all-extras` cannot work on Linux

Anyone following it to set up a dev environment hits `FileNotFoundError: '/usr/bin/sw_vers'`: the `app` extra pulls `pyobjc-framework-webkit`, which is macOS-only. `uv sync --extra dev --extra bundled` is the working invocation on Linux and is what a contributor guide should say.

### Aside — verifying this PR locally produces 3 spurious test failures (pre-existing, not from this PR)

Not caused by this branch, but it will bite anyone who verifies it. `test_git_posix_spawn.py` (×2) and `test_subprocess_encoding.py` are repo-wide source sweeps, and they walk into `fused_render/ai/runners/*/.venv/lib/*/site-packages/`, flagging `sympy` and `torch` internals as offenders. They pass on a clean worktree and fail in any checkout where a runner venv exists — and since the app `uv`-syncs runners *in place*, that means **any developer who has ever loaded a local AI model.** CI never sees it because CI has no runner venvs. The sweeps should exclude `runners/*/.venv`.

---

### Context for a follow-up: llama.cpp dominates `transformers-text` on both tiers

Since this PR establishes the GGUF path, here is the head-to-head that follows from it. Same machine, same model family (Qwen3.5-4B), same prompt, 128-token cap, back-to-back and uncontended:

| engine | build | download | load | first token | tok/s | peak mem |
|---|---|---|---|---|---|---|
| llama.cpp | Vulkan GPU | 2.93 GB | 1.6 s | 0.14 s | **53.4** | 3.14 GB |
| transformers | ROCm GPU | 8.7 GB | 8.3 s | 1.84 s | **12.6** | 9.18 GB |
| llama.cpp | CPU | 2.93 GB | 1.3 s | 0.71 s | **6.4** | 3.44 GB |
| transformers | CPU | 8.7 GB | 5.9 s | 2.9 s | **2.7** | 8.37 GB |

llama.cpp wins **both** tiers on **every** axis at once — 4.2× throughput on GPU, 2.4× on CPU, at a third of the download and a third of the memory. All four outputs were correct fluent prose, so this is like-for-like on quality. This is also the x86 re-measurement that `formats.py`'s `UNLOADABLE_QUANT` note explicitly asks for before anyone revisits the quantization decision.

For the record, `transformers-text-rocm`'s availability gate is **honest** on RDNA4: torch 2.13.0+rocm7.1 reports `cuda_avail True` on gfx1200 and genuinely generated on the GPU. transformers got a fair fight and lost by 4×.

PR #710, opened 2026-08-21
Claude chat: show and stream a turn this page did not start (D406)

AkshilVT:
## The report

A Claude Code background shell task finished → the harness injected a `<task-notification>` turn → the agent woke and worked for minutes — and the chat showed **no "Thinking…" indicator, no streaming, nothing at all**, until a few "continue"s later produced a reply. The `<task-notification>` also rendered as a raw XML message bubble.

## Root cause (two, and the first is the real one)

**1. `_poll` latched `done` on the first `result` row.** One `claude -p` process runs *several* turns: a turn that started a background command is woken by the harness when the command finishes, and everything it then says goes into the **same** `out.jsonl`, after the `result` that closed the first turn. Verified on real run dirs on this machine — two `result` rows in one file, the second carrying `origin: {kind: "task-notification"}`.

`done` was set at the first `result` and never taken back, so by the time the woken turn started, `pollLoop` had already removed the working line, cleared the `run` param and called `setRunningUi(false)`. The state variable that missed the turn is `done` (agent.py `_poll`), and the page-side consequence is `activeRun`/`body.running` being released for a run that was still going.

`done` is now **per turn**: `idle or not alive`, where `idle` means "the last row in the file is a `result`". The wake's first row flips it back to False.

**2. Nothing ever asked again.** The page's only "is something running that I am not showing?" question was `adoptLiveRun` — a ~3s window around *opening* a chat. After that, the running state could only be armed by this composer's own `sendMessage`, i.e. the state machine was about *who sent the message* rather than *what the session is doing*. Every externally-initiated turn missed it (harness wake, the same chat open in another pane, a canvases spawn). There is now a standing watch on a 10s timer (`liveWatchTick`) that runs one `adoptLiveRun` lap and derives "running" from the **server** (`live_run`, which reads the run dirs' pids), never from local send events. It attaches `quiet` so it can never print a user line the reader did not send.

**Rejected: `done = not alive`** (built first) — the process outlives a finished turn for as long as the background command runs, so the chat would sit busy and queue everything the user typed for an hour. Per-turn `done` + a standing watch costs ≤1 watch interval of lag and keeps the chat composable in between.

## Secondary (included)

`<task-notification>` is now a compact system chip instead of raw XML. `_history` filtered exactly two synthetic `user` records by name (`<local-command`, `<command-name`) and let the rest of `_MACHINERY_DROP` through verbatim; those are now dropped as turns, and a task-notification becomes a `notice` segment — one muted line from the CLI's own `<summary>`, rendered as **plain text, never markdown**. One renderer reads both shapes (the live `system`/`task_notification` row and the transcript's XML record), so a restored conversation and a streaming one show the same chip.

## Test plan

- `pytest tests/ -q`: **8 failures, all pre-existing on clean `origin/main`** (verified by stashing): 6 × `test_claude_health.py`, `test_ai_metrics.py::test_a_missing_claude_binary_is_counted`, `test_calls.py::test_an_abandoned_merge_closes_the_day_s_files`. 8770 passed.
- New tests: `test_claude_stream.py` (done semantics: result-then-nothing ends the turn; a result the run has spoken past does not; the second result does; a dead process always does; the delta-less `result` text fallback is not delayed), `test_claude_agent_segments.py` (notice segment live + from history; the other machinery records dropped; a message that merely *mentions* the tag is still the user's), `test_claude_template_segments.py` (node probes: the chip is plain text, never markdown; it keeps two replies apart), `test_claude_live_run.py` (node probes: the watch asks once a lap, adopts a foreign run, and is quiet when there is no session / a turn is attached / the gate is held / a lap is in flight, and passes `quiet`).
- `cd frontend && npm run build`: clean (no frontend source touched).

## Verified live

Dev server from this worktree, claude template on a demo folder, with a fake **alive** run dir plus a session transcript, i.e. a turn this page never sent:

- Opening the chat: the transcript's `<task-notification>` renders as `▸ Background command "pytest -q" completed (exit code 0)` — no XML bubble.
- Appending the wake rows to the run's `out.jsonl`: within one watch interval the page armed `body.running`, showed `✻ Composing… (20s · ↓ 11 tokens) stop · esc`, swapped the send button for a stop square, drew the notice chip inline and **streamed the woken turn** — with nothing typed into the composer.
- Appending the closing `result`: chrome released, `run` param cleared, transcript finalized. No console errors.

## Notes

- **Known gap, recorded not closed:** a turn driven from a terminal (`claude --resume`) has no run dir, so `live_run` cannot see it. Catching that means tailing `~/.claude/projects` and guessing turn boundaries; D406 records the trade-off.
- Narration lives in the template/agent comments as usual; DECISIONS.md gains **D406** (next free after D405 on `origin/main`). No SPEC rule covers the run lifecycle, so nothing there needed renumbering.


AkshilVT:
## Follow-up: the two-tab report was latency, not arming

**Which hypothesis was real: (b).** (a) is disproven — a lap that catches a run *mid-turn* does arm the chrome (verified again below, and in the original PR evidence). What happened in the owner's test is that the only trigger was a 10s interval: by the time the lap looked, the short turn had finished, so `resumeRun` took its **done-branch** and repaired the transcript silently. Hence "same output ~10s later, no chrome". (c) was not needed as an explanation.

### What changed (commit `cc698a83`)

1. **The tab that starts a turn now pokes the others.** `noteChatActivity` already stamps `localStorage` as a poll loop *opens* (for the shell's tasks store), and a `storage` event fires in every other document and never in the writer — exactly the shape this question needs. The two-tab case now costs no polling and lands in ~1s. Key is a single shared constant.
2. **`visibilitychange` + `focus` lap immediately**, so a backgrounded tab catches up on return instead of waiting out an interval.
3. **Interval 10s → 5s, and skipped while `document.hidden`.** It is now only the backstop for the one starter that pokes nothing this page can hear: a **server**-started run (scheduled message, canvases spawn). Tradeoff recorded in D406: a hidden tab can miss a server-started run until the reader returns — the same repair it used to get — and chrome nobody is looking at is worth nothing.
4. **`quiet` became dedupe, not silence.** Suppressing the run's user line unconditionally was right for a woken run whose message is already on screen and *wrong* for another tab's send: the second tab showed an answer hanging under no question (I hit this in the first live pass). `quiet` now means "do not print a message this transcript is already showing", asked of every user bubble — where `matches` asks only the last one, because `matches` also decides whether to STRIP rows. The **done-branch takes the same test**, so a short turn made elsewhere and finished before the first look is now rendered *whole* (prompt + reply) instead of dropped.

### Live evidence (cmux, two real tabs, one session, real `claude` turns)

- Tab B (hidden) sends "List 8 colours, one per line" → **tab A at +1s**: `body.running`, `✻ Noodling… (1s · medium effort) stop · esc`, send button swapped to the stop square, and the other tab's prompt bubble printed above it. Screenshot in thread below shows it mid-turn at 9s.
- Turn ends → tab A: chrome released, `run` param cleared, reply complete (`red/blue/green/…`). No console entries, no duplicated bubbles in either tab.
- The earlier fake-live-run repro (message already on screen) still adopts without re-printing the prompt.

### Tests

- `tests/test_claude_live_run.py`: +6 — the storage poke adopts (and another store key does not), returning/focusing laps at once, a hidden tab does not spend the interval (and the interval is 5s), the `onScreen` predicate answers over the whole transcript, and both render sites consult it.
- `tests/test_claude_schedule_pill.py`: its source anchor widened to the *condition* it pins (`neverShown` is one way into that append, no longer the only one) — it was matching the literal string I reworded.
- Full suite: **8774 passed**, failures = the same 8 pre-existing ones from the PR body, plus `test_ai_worker_base.py::test_a_refused_body_this_cannot_frame_ends_the_connection`, which **flakes on clean `origin/main`** too (verified: 1 failed / 1 failed / all passed across three stashed runs, different parameter each time).

### FYI acknowledged

Real background-task wakes can't be exercised in-app (`claude -p` kills background tasks at turn end). The per-turn `done` fix still stands on its own — it is what lets *any* second turn in one run dir be streamed — and the watch is what covers server-initiated resumes and other tabs.

PR #732, opened 2026-08-21
Playground: try a local model without writing a page

vasu2001:
## What

A **Playground** tab on **/ai-models** — first in the strip and the page's new default — where anyone picks a local model from a sidebar and uses it immediately, plus the doors that lead people to it (Home cards, primary-nav sidebar row) and the doors out of it (the app-builder seed).

### The tab

- **Sidebar** = `GET /api/ai/catalog` verbatim, grouped **Chat / Images / Transcription / Embeddings**, showing a new curated `nickname` (no quantization noise), the `size_gb` figure ("how much disk will this cost me"), loaded/downloaded state, and the curator's `note`. Unavailable capabilities stay visible with their reason (HF-8's lesson). Cached (uncurated) downloads appear automatically (D323).
- **Chat stage** (text-generation): streamed NDJSON, Stop (cancel without unload, AI-1a), tok/s from the done frame, collapsed `<think>` blocks, and the AI-5 cold-start dance — 409 `model_loading` → watch the load job → retry once. Advanced drawer: temperature / top-p / max tokens / system prompt on the server's own clamps; a standing default system prompt keeps first answers short and jargon-free.
- **Image stage** (text-to-image): job-backed render, live preview poll while denoising, **settled-params echo** (the server snaps/clamps/invents the seed — the caption states what ran), seed reuse, session thumbnail strip. Defaults benchmarked to 512×512 at guidance 1 so the first picture lands in seconds, not minutes.
- **Transcription stage** (speech-to-text): **Record** (MediaRecorder) or **Pick a file** → `POST /api/fs/upload` into `~/recordings` → transcribe by path; progress in seconds of audio; words stream in via the partial-JSONL tail (runners/partial.py). A recording is replayable and re-runnable across models.
- **Embeddings stage** (new capability, from #728): a search-by-meaning demo — prefilled lines and a query that shares no words with its best matches, one `/api/ai/embed` batch (query + up to 63 lines), dot-product ranking (vectors come back unit-length, so dot = cosine), bars scaled to the best match because raw SigLIP cosines (~0.2–0.4) would read as broken; raw score on hover. Rides the same cold-start 409 dance as chat.

### The doors

- **Home strip**: one card per capability, each linking to `?cap=` which lands on that group's default model. The strip measures its width and renders `slice(0, shown)` — the embeddings card is appended last on purpose, so it's the first one a narrow window drops.
- **Sidebar**: AI Models is primary nav (row + rail item after Canvases) with a Beta tag; the resident-model dot moves off the Settings trigger.
- **App builder**: model cards open the Playground; the Playground's "Build an app with this" opens the /apps composer with a seed naming the model, the settings the user dialled in (read off the URL), the matching `fused.ai.*` call — including `fused.ai.embed` — and a pointer at the `fused-render-ai` skill so the spawned session doesn't improvise the contract.

### The restructure (D420)

/ai-models became **one path per tab**: a sub-app on the frontend (`apps/ai_models/` with `playground/` and `lib/`), a lib on the backend. The playground moved out of `shell/` wholesale; shared helpers (URL params, engine labels) live in `lib/`.

## Decisions baked in (flag if wrong)

1. **Playground is the default landing** — bare `/ai-models` opens it; `?tab=local` names the inventory. Old links with an explicit `tab` param are untouched.
2. **Selecting ≠ loading** (AI-4): a click renders the stage; weights move on Download/Load or the first generation. Eviction is a sentence in the header, not a modal.
3. **`nickname` is a curated field** in catalog.py beside `label` — never derived by stripping the parenthetical (AI-2c's reasoning). Cached entries fall back to `label`.
4. **URL carries the setup, never the transcript**: `model` + per-stage non-default settings via `replaceSearch` (no history spam). The embed stage's query/lines are transcript, so they stay session state.
5. **Leaving the chat cancels the stream + generation; image/transcribe jobs deliberately survive navigation** (they are job-shaped on purpose and stay visible in Activity).

## Notes

- `playground/client.ts` reimplements a small slice of runtime.js (NDJSON reader, 409 dance, job watch, embed unwrap) — the shell can't ride the page runtime; the contract is pinned by the existing server-side tests.
- No tests in this PR (explicitly deferred). `tsc` + boundaries + vite build pass; `catalog.describe()` smoke-checked for the new field.
- Follow-ups: the Local tab's cache walk still runs on mount even when landing on Playground (pre-existing behaviour, worth deferring later); playground stage params linger in the URL when switching tabs (harmless, ignored). SPEC/DECISIONS rows for the playground still owed.
- Manual smoke pending: chat cold-start on a real model, mic permission flow, image preview on a model without a fitted projection, embeddings run on a real SigLIP2 model, Home card → `?cap=embeddings`, narrow-window card drop.


vasu2001:
**Second wave — full redesign (b7f186eb):** researched LM Studio / Jan / Open WebUI / GPT4All and the OpenAI / AI Studio / fal / Replicate / Midjourney playgrounds, rebuilt the tab on what they converged on.

- **Settings now live in a right rail** (slider + number pairs, plain-language hint under each, per-control reset; persistent ≥1100px, Controls toggle below). The bottom "Advanced ▾" drawer from the first cut is gone.
- **Chat**: rendered markdown (tiny React-element renderer, untrusted text never hits innerHTML), code blocks with Copy, per-reply tokens + tok/s footer, auto-growing composer with an accent Send, starter-prompt chips on the empty state.
- **Images**: aspect-ratio chips + Quick/Finer/Max speed row over the same URL params; **steps default to the model's benchmarked count** via a new curated `defaults` hint in catalog.py — FLUX.2 klein is step-distilled, D310 measured 4 steps (~34s/image); the generic 28 made a first image take minutes. Unknown models keep the server's 28. Settled-params caption, live step progress + ETA, session history strip.
- **Transcription**: record button with red pulse, elapsed time and a live level meter; drag-and-drop + browse; seconds-of-audio progress bar.

Build + tsc green; visuals not yet eyeballed in a running app — screenshots to follow from manual smoke.

PR #736, opened 2026-08-22
map: server-managed tile engine, faster vector tiles, tile cancellation

amanbagrecha:
## What & why

Opening a remote raster (e.g. a NASA HLS / MODIS COG on Azure) in the map template hung on "Loading visible tiles…". The root problem was architectural: the tile daemon was spawned detached from a short-lived `runPython` worker, and the browser held tile URLs bound to that daemon's **ephemeral port**, so any daemon death/wedge/restart silently broke every tile — with no owner to heal it (and a spawn storm of orphans, 18 at one point).

This moves the map tile engine's **ownership into the fused-render server** and serves tiles through the stable `:1777` origin, so daemon death becomes invisible to the page. It then makes large vector tiles fast, cancels abandoned work, and fixes antimeridian rasters. Design record: `docs/MAP_ENGINE_SERVER_DESIGN.md`.

## Changes

- **Server-owned daemon** — `server/map_engine.py` supervises exactly one child (spawned with the caller's project-venv interpreter, health-checked, restarted with a describe **replay**, killed at app shutdown, modeled on `ai/supervisor.py`). `server/routers/map_tiles.py` proxies `/api/map/*` with **heal-on-failure** (restart + retry). `map_render.py` hands off ownership (−180 lines of daemon plumbing); `daemon.py` drops its singleton/idle/retire logic; `template.html` loses the client-side recover hack.
- **Deadlock/oversubscription fix** — all geo work runs on persistent pools (`RENDER`/`DESCRIBE`/`VTILE`); a thread that did GDAL `/vsicurl` work deadlocks the whole process when it exits on Windows, so handler threads never touch the geo stack.
- **Vector tile performance** — a hand-written MVT encoder (`mvt_encode.py`, drops GDAL's ~0.9s/tile overhead) plus an RTree **node-summary overview** for dense low zoom, and a persistent per-source sqlite connection. Measured on a 3.58 GB GeoPackage: **z6 16.9s → 0.02s, z8 5.6s → 0.02s**.
- **Cancel abandoned tiles** — async proxy that races the fetch against client disconnect, and a daemon that interrupts the in-flight sqlite query. Pan/zoom no longer piles up work, and one heavy layer no longer blocks other tabs (abandon a 17s tile → next tile 2.2s vs 21–29s; `/api/config` stays responsive during a render).
- **Antimeridian rasters** — a MODIS sinusoidal tile near 180° describes with `west > east`; drop that invalid crossing `bounds` (MapLibre silently rejects it and never adds the source → the "only a bbox, no imagery" symptom), fit across the seam, and outline the real footprint.
- **Test-home leak seal** — importing `supervisor.__main__` repointed `FUSED_RENDER_HOME` at the real `~/.fused-render` for the whole pytest worker (Windows-only, invisible in CI); restore the env around that import and add a session conftest guard that fails loudly if the real staged templates change.

## Tests

Curated (not bloated): new `test_map_engine.py`, `test_map_tiles_router.py`, `test_vector_tiles.py`; retired the moot daemon-lifecycle tests; e2e trimmed 20 → 14 focused cases + an antimeridian guard. **70/70 map unit tests pass; 14 e2e + 1 skip** against a worktree server. Verified end-to-end against real datasets (HLS, MODIS, a remote Maxar COG, and the GeoPackage), including heal-after-daemon-kill.

## Notes

The fix reaches the installed app only on rebuild/reinstall. Deferred (in the design doc): a static PMTiles/precomputed pyramid for very large vectors; in-render GDAL cancellation for raster tiles (currently queue-dequeue only).


iamsdas (changes requested):
Thanks for this — the diagnosis is right, and `engine_host.py` + `/api/engines/*` is exactly the abstraction we want: generic child supervision, health-check, replay, proxy, disconnect-cancel, with "nothing here knows what a tile is". That part I'd merge happily.

The blocker is the layer above it. `server/routers/tiles.py` and `fused.tiles` in `static/runtime.js` put geospatial vocabulary into the core server and the shared JS bridge — and as far as I can tell they aren't buying us anything:

- **It's redundant.** By the time `map_render.py` returns its descriptor, it has already ensured the engine, registered the replay, and rewritten `tile_url` / `vtile_url` / `job_url` to stable `/api/engines/map/proxy/...` paths. The page holds self-healing same-origin URLs at that point — the daemon-death bug is fixed. `template.html` can point MapLibre straight at them.
- **It duplicates ownership.** `map_render._URL_KEYS` comments that the URL shape is "this template's knowledge" — then `tiles.py` hardcodes that same shape again (`f"/api/tiles/{layer_id}/{{z}}/{{x}}/{{y}}.png"`, `_PUBLIC_KIND`, `bounds`/`minzoom`/`maxzoom`) in the server. One fact, two owners, and the second one is core code that has to change whenever the map descriptor does.
- **The security rationale doesn't hold.** The docstring calls it "an ALLOWLIST, not a passthrough", but the proxy only gates POST behind `X-Fused` — GET is open, because MapLibre's tile requests can't send the header. A page can hit `/api/engines/map/proxy/tiles/...` directly. The `{layer}` traversal guard is worth keeping, but it belongs on the proxy where it protects every engine.
- **`_describe_layer()` reimplements `/api/run`** — re-branching on `shell_prefs.effective_engine()`, picking `_engine.run_python` vs `to_thread(run_python, ...)`, re-shaping the error. Executor dispatch inside a router is a sign the layer is reaching for something it shouldn't need.
- The server also now reaches into `templates/` by path, which is the boundary `test_templates_decoupled.py` exists to protect; the PR has to widen that guard (excluding `templates/*/tests`) to accommodate it.

So the ask is a deletion, not a redesign:

1. Drop `routers/tiles.py`, `tests/test_tiles_router.py`, the `fused.tiles` block in `runtime.js`, and the wiring line in `app.py`.
2. Move the `{layer}` bare-token validation onto the generic proxy.
3. Leave every geospatial concept — tile URL shape, descriptor kinds, `colormap` / `rescale` / `stretch` / `renderMode` — in `map_render.py` and `template.html`, where they already live. Styling knobs are just render params.
4. Nothing geospatial in `fused_render/server/`.

On generalizing further: once a second template wants supervision, I'd have the template *declare* its engine (entry point, health path, public routes) as data the host reads, so a new engine is a manifest rather than a new router. But that's speculative today — the caller-supplied `ensure` payload is simpler and correct, and I'd rather not build the manifest until there's a second caller.

Everything else here is welcome and largely independent: the pool/deadlock fix, the MVT encoder and RTree overview (z6 16.9s → 0.02s is great), disconnect cancellation, the antimeridian fix, the test-home leak seal. Happy to take those plus the engine host right away and land the rest once the public surface is gone.

---

**Summary:** the primitive is right and we want it. The geo-specific face on top of it is redundant — the page already gets stable healing tile URLs from the descriptor — and it costs us duplicated ownership of the map template's own knowledge inside core server code and the shared bridge. Delete that face, keep everything else.

iamsdas (approved):

PR #739, opened 2026-08-22
Tasks board reads chat turns off the last message, not a 45s window (D420)

AkshilVT:
## What

PR2 of the task-sync work (#719 was PR1). Fixes the reported limitation: chat-template tasks "take some time to update on board when in progress and when done".

- `tasks._live` now decides a chat turn's liveness with `session_liveness.transcript_turn_open` (D415's last-message rule) instead of the 45-second activity window. A plain-text assistant row is a turn that ended; a user row or a `tool_use` block is one in flight. Done shows on the next poll instead of up to 45s late.
- The window's other lie is gone with it: a late plain-text closing row no longer flips a finished run back to In Progress (symptom 3's shape).
- `transcript_turn_open` gains the row-timestamp half of the stale ceiling: housekeeping appends (`ai-title`, away summary) bump the file mtime hours after a turn died, and the mtime gate alone read the dead turn's last user row as in flight. The deciding row's own timestamp is now checked against `STALE_TAIL_SEC` too — this also hardens the chat page's working line, which shares the rule.

Scheduled runs untouched: the store's `turn` stays the authority (PR1). The tail read stays for `active`, the timestamp `_verdict_outvotes_live` measures against.

## Tests

- 3 new matrix tests: finished chat reply is Done the moment it lands; mid-`tool_use` chat turn stays In Progress; work past the verdict keeps In Progress (TASK-001, now on an honest `tool_use` tail).
- Reclassified pin: `test_transcript_past_echo_window_reads_in_progress` → `test_late_closing_text_does_not_flip_the_board_back` (board now agrees with the dock).
- New liveness pin: stale user row under a fresh mtime is not in flight.
- Full suite: 9085 passed; remaining failures are this machine's known local set (claude_health/ai_metrics need a real `claude`, test_calls abandoned-merge, ai_worker_base — all fail identically on clean main).

## Accepted trade-off

A turn left open by a dead process reads In Progress up to `STALE_TAIL_SEC` (90s) instead of the window's 45s — the same ceiling D415 accepted, lying late rather than wrongly.


AkshilVT:
claude --resume a606c6d4-7d4d-47a6-8701-ea1906a5fb91
note to myself

PR #757, opened 2026-08-23
Model cards say what they are, whether you have them, and how much of them (D436–D440)

AkshilVT:
## The problem

The AI Models tab had four card states and its cards' surfaces drew exactly one of them (loaded). Three of the four are facts about the disk — so *"which of these do I already have?"*, the question a grid of models is read for, could only be answered by reading each card's buttons one at a time. A user also hit a genuine dead end in this tab, and a partly downloaded card lied about its own progress.

### 1. The card's surface states the disk facts (D436)
- **Downloaded** — the green wash plus the search results' own `✓ downloaded` chip in the footer.
- **Loaded** — the *same* wash, set apart by its tinted border and 3px rail. Two strengths of one green read as two shades of a thing rather than as two things; a card either has an edge or it does not, which is a distinction made at a glance.
- **Partly downloaded** — an amber boundary at the fraction fetched; flat wash where no denominator exists, because *"some of this is here and we cannot say how much"* is a different sentence from *"2% of it is"*.
- **Not downloaded** — unchanged.
- All washes are **background only** (no border, rail, radius, padding), so a card changing state cannot resize and shift its neighbours mid-scroll.

### 2. One hue per engine (D436)
The engine tag gives up the accent — now spoken for by the ✓ chip — for one of the calendar's categorical hues per engine **family**, assigned by a named table rather than hashed, so the single lime hue is spent deliberately. A hardware-qualified label resolves to its family's hue; a family this build cannot place stays muted rather than taking a colour we cannot back. The **unavailable** tag keeps its dashed border and warning hue, where the state outranks the identity.

### 3. The dead end, given a way out (D437)
Reported from a user's screen: a cancelled download left a folder holding one 40-byte `refs/main` and not a blob. Everything the page said was true — *partly downloaded* (no snapshot **is** the evidence, D424), *40 B* (the folder), *Unrecognised* (no files → no task → no capability) — and the card still had no working control: the primary button was a **disabled** Download whose hover said to delete it, and the delete was an unlabelled glyph third in a row of four. The user deleted the folder in Finder.
- Reproduced first; the delete endpoint was fine, the card was the dead end.
- Where a resume is impossible, the primary control becomes a **labelled Delete** onto the same confirm dialog.
- A weights-only fetch removes such a folder on its way out — guarded on there being **no snapshot AND no blob of any kind**, so it can never take bytes a resume would have used (D275/AI-5i), and it reads the folder rather than the job's outcome, so it is a no-op after a success.

### 4. The tab reads "Models" (D438)
With recommendations in every row and a Hub search above them, most of what the tab offers is not local. Path stays `/ai-models/local` — it is in bookmarks and every `tabHref` cross-link.

### 5. A download in flight fills its card (D439)
Green boundary at the job's progress, on **all three** card kinds — a multi-GB fetch deserves more than 3px of bar in a card watched from across a carousel. Amber is a boundary that has stopped moving, green is one that has not; `partly downloaded` cannot draw that, since a partial wears the tag while resuming.
The fraction is the **larger** of the two readings available, a monotonic guard: pressing Download used to collapse a 94% fill to the new run's few percent. The job's total is the denominator wherever there is one, gated on it counting **bytes** — a step-counted stage has a perfectly good total that is not bytes, and dividing disk bytes by it drew a full card from a four-step job.

### 6. The fraction counts bytes that ARRIVED, and a running pull can be stopped (D440)
The fetcher **preallocates** a part file to the full length of the file it is fetching, so a 1.6 GB download reported 1.6 GB on disk the instant it began — the card drew a nearly-full boundary while the job row on the same card said `243 MB / 1.5 GB`.
- `size` stays what the page **prints** (allocated bytes are what the folder costs); the listing now also carries **`fetchedBytes`** — the scan total with one correction per part file, taken from the sidecar cursors `flush()` writes only *after* fsyncing the data, which is the accounting a resume itself trusts.
- A part file with **no** sidecar counts zero, never its length (positive evidence only). hf's `.incomplete` is not corrected — that writer appends, so its length already is its progress. A sidecar's own bytes are excluded: a fraction made of bookkeeping is not a fraction.
- **The ✕**: while a pull ran the primary button was a disabled "Downloading…" and the trash was disabled too, so a multi-GB fetch started by mistake could not be stopped from the card that started it. `CancelButton` moved to `shared/` and the repo card wears the same one, under the manager's own rule about when a job may be cancelled.

## Verification

Full lifecycle on a real 1.6 GB repo, reloading at every stage:

| stage | fill | job row | controls |
|---|---|---|---|
| downloading | 13.6% → 47.8% | 209 MB → 736 MB | ✕ present |
| **reload mid-download** | 73.7% | 1.1 GB / 1.5 GB | ✕ present, trash correctly disabled |
| cancelled | amber 84.4% | — | Download + trash back, ✕ gone |
| **reload while paused** | 84.4%, identical | — | hover names the 1.4 GB actually here |
| resumed | forward, never back | 1.5 GB / 1.5 GB | ✕ present |
| finished | green, no gradient | — | ✓ downloaded + Load |

Also checked in the browser: every card state in light and dark; computed styles per card (distinct hues for MLX LM / MLX Whisper / MLX FLUX / MLX Embeddings / Transformers Embeddings, dashed amber for the unavailable Diffusers tag); the labelled Delete driven end to end on a synthetic refs-only shell (folder gone, no Finder); hub search results wearing the same wash for a model already on disk; card heights identical across states within a row; console clean.

Gates: `bun test` **2235 pass**, `npm run build` (boundaries + tsc + vite) clean, `pytest` **772** across `test_ai_models_api` / `test_hub_models` / `test_theme` / `test_ai_runtime`. New tests cover `engineHue`, `partialFraction` (including the preallocation case and the never-backwards rule), `jobFraction`, `emptyShell`, the shell sweep, and the durable-byte accounting (no sidecar → 0, torn sidecar → 0, over-claiming sidecar clamped, hf `.incomplete` counts its length, finished repo reports `fetchedBytes == size`).


AkshilVT:
Reopening to re-trigger CI — GitHub did not dispatch a run for 745edd7c.

PR #761, opened 2026-08-23
Playground: apps that can use the selected model, from showcase metadata (D442)

vasu2001:
## What

Community/showcase apps can now declare in `metadata.json` which AI they drive:

- `ai_capabilities` — catalog capability ids ("text-generation", …) the app works across (any model of that capability).
- `ai_models` — exact wire model ids (repo id or GGUF filename) when the app only works with specific models. Wins over the capability list: a hit is a recommendation.
- `ai_model_param` — the URL query param the app reads its model from. Required for a listing, and never literally `model` (the shell's sidebar owns that key — every AI app in the repo already namespaces: `chatModel`/`imageModel`/`asrModel`).

The Playground draws a strip under the stage ("Use it in an app") listing matching showcase apps for the selected model:

- capability match → listed
- `ai_models` hit → listed first with a **Recommended** badge
- `ai_models`-only app whose list misses the current model → not listed
- click → opens the showcase clone's copy at `/explorer/view/…?<ai_model_param>=<selected model id>`, plus the usual `touch` marker

Catalog read is `action:"catalog"` only — hidden entirely on no-cache/error/no-match; cloning stays the hub Showcase tab's job.

## Structure

New files only touch PlaygroundTab.tsx by two lines and append a self-contained `.pg-apps-` CSS block, so the concurrently open #760 redesign merges clean.

- `frontend/src/apps/ai_models/playground/appMatch.ts` (+ tests, 7 cases)
- `frontend/src/apps/ai_models/playground/PlaygroundApps.tsx`
- DECISIONS.md D442

Companion metadata PR in fused-render-community-apps declares the fields on local-chat / local-image / local-transcription.

## Tests

`bun test`: 2242 pass (7 new). `tsc --noEmit` clean.


vasu2001:
Updated per owner direction: the strip now reuses the hub's `AppPreviewCard` (same card as /apps and /home — authored preview.png, live-iframe hover, `.apps-cards` grid) fed a synthesized `AppInfo` for the showcase clone's copy. The card gains an optional `href` override for the model-handoff query string, which `hrefFor` deliberately never carries; recommended matches wear the badge slot ("Recommended for this model").

PR #763, opened 2026-08-23
Warm workers for apps: a zero-config /api/engine that keeps a script's process alive

amanbagrecha:
## What & why

Every `fused.runPython("./x.py", …)` spawns a fresh subprocess that imports the module, runs `main()`, and exits — so an app talking to a remote service re-pays the whole cost each call (measured for the S3 Browser: ~7s/read, ~2.5s of it just `import botocore` + client build; a bucket switch fires two ≈ ~14s). The engine host from #736 can keep a worker alive, but only for built-in templates.

This adds a **zero-config warm variant of `/api/run`**: an app page calls `fused.engine("./x.py").call(params)`, the server keeps that script's worker process alive between calls, and `main()` is re-invoked in the already-warm process — so module imports run once and module globals (a warm client cache) persist. No manifest, no registration, no templates-root requirement; the author swaps one call and keeps the same `main()`. Design doc: `docs/ENGINE_HOST_APPS_DESIGN.md`.

## How it works

- **`POST /api/engine`** (`server/routers/engine.py`, + `/api/engine/forget`) — same body as `/api/run` (`{py, html, params}`), `X-Fused`-guarded, resolves the file/interpreter exactly as `run.py` does, then forwards to the warm worker via the engines proxy (heal-on-failure, cancel-on-disconnect).
- **`engine_worker.py`** — the standard warm worker: fits the engine_host child contract, imports the target module once, re-imports on mtime change, returns `_child.py`'s exact envelope. Calls run **concurrently** in the one process (only import/mtime/`main`-lookup is locked; `main()` runs outside it), with per-call stdout captured via a contextvar so concurrent output never scrambles. A multi-process pool remains an option for CPU-bound `main()`.
- **`engine_host.ensure_app()`** — bring-up keyed by the resolved `.py` path (no templates-root check, no caller-supplied interpreter), reusing #736's spawn/health/heal/reap/shutdown; adds idle-retire.
- **`projectenv.interpreter_for()`** — the same interpreter choice `/api/run` makes (bundled `sys.executable`, else the project venv).
- **`fused.engine(py)`** in `static/runtime.js` — `{call, forget}`, sharing `runPython`'s latest-wins stale-cancel channel and headers, degrading to `runPython` when the server is unreachable (so hosted/exported pages still work).

Opt-in throughout: `/api/run` is unchanged and stays the default.

## Testing (dev server, isolated `.dev-home`)

Verified end-to-end: warmth (a trivial app: 41ms warm vs 658ms cold; real botocore via the S3 Browser: ~0.3s warm vs ~2.9s cold), in-process **parallelism** (two concurrent `sleep=1` calls in 1.04s, one process), **mtime-refresh** (edit → reload), **heal-on-failure** (killed worker → transparent restart), **idle-retire** (workers reaped after idle), per-call **stdout isolation**, the **`X-Fused` 403** guard, graceful S3-error handling with worker survival, and shutdown-reaping.

## Notes

Phase-2 follow-ups noted in code/docs: a multi-process pool for CPU-bound `main()`, a per-origin LRU cap, `/api/env/install` integration for unbuilt venvs, and a hard per-call watchdog. The map template's existing engine path is unchanged.

---

> [!NOTE]
> **Medium Risk**
> New long-lived process path that executes user `.py` with persistent module state, but it is X-Fused-gated and reuses the same resolve/interpreter trust model as `/api/run`. Idle-retire, heal, and at-most-once forwarding bound the extra lifetime risk.
> 
> **Overview**
> Adds an opt-in warm path for app scripts so `fused.engine("./x.py").call(params)` reuses a long-lived worker instead of a fresh subprocess per `runPython` call. Same `{py, html, params}` body and result envelope as `/api/run`; `/api/run` is unchanged.
> 
> The server resolves the file and interpreter the same way as run, then `engine_host.ensure_app` spawns shipped `engine_worker.py` (keyed by path hash). The worker imports once, reloads on mtime, runs concurrent `main()` with per-call stdout, and is idle-retired after 15 minutes. Proxy logic moves into shared `engine_forward` (keep-alive pool, heal, cancel-on-disconnect, 60s budget, at-most-once for `/call`).
> 
> Runtime `fused.engine` shares stale-cancel with `runPython` and falls back to it only on a network TypeError. Unbuilt venvs return 409 rather than installing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db667125a3710151c0dd4267d811d38ec53f84e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iamsdas (approved):
the tests seem to be all failing

amanbagrecha:
> the tests seem to be all failing

The only CI failure is a pre-existing main breakage (ai-playground.css colour literals from PR #771) — unrelated to this map PR, and main itself is red on it. How do you want to handle it?

amanbagrecha:
### Code review

Two correctness issues, plus one minor consistency note.

**1. A failed POST can double-execute `main()`.** On the warm app path (`/api/engine` → `_forward` → `_proxy`), a POST that reuses a pooled keep-alive and fails with anything other than `RemoteDisconnected` (e.g. `ConnectionResetError`/`IncompleteRead` — what you get when the child drops the connection *after* it already started running `main()`) makes `_proxy` `return None`. `_forward` treats `None` as "unreachable", restarts the worker, and **re-sends the same POST body**, running a side-effecting `main()` a second time. The comment says returning `None` avoids this ("let `_forward` heal it instead of risking a double-execution"), but `_forward`'s heal path *always* retries after the restart — so it causes the double-execution rather than preventing it. Since `None` means both "provably never sent, safe to restart+retry" and "may have run, don't retry", there's no way to express the latter.

https://github.com/fusedio/fused-render/blob/6878cacda67befdaaa09f2219d6ea70813107b69/fused_render/server/routers/engines.py#L172-L186

https://github.com/fusedio/fused-render/blob/6878cacda67befdaaa09f2219d6ea70813107b69/fused_render/server/routers/engines.py#L210-L219

**2. A 504'd call is marked idle while `main()` is still running, so the idle reaper can kill it mid-flight.** On a `call_timeout`, `_proxy` severs the socket but the worker keeps running `main()` (by design — "we neither heal nor retry"). `routers/engine.py`'s `finally: mark_idle(engine_id)` then fires immediately, zeroing `_busy` and stamping `last_used`. If no other call reaches that worker within `APP_IDLE_RETIRE_S` (15 min), `reap_idle_app_workers` sees `_busy == 0` and terminates the process while the original `main()` is still executing — a silent truncation of a genuinely-busy worker, which is exactly what the heal-safe busy count was meant to prevent.

https://github.com/fusedio/fused-render/blob/6878cacda67befdaaa09f2219d6ea70813107b69/fused_render/server/routers/engine.py#L82-L89

https://github.com/fusedio/fused-render/blob/6878cacda67befdaaa09f2219d6ea70813107b69/fused_render/server/engine_host.py#L500-L508

**Minor:** both handlers in `routers/engine.py` hand-roll `json.loads(await request.body() ...)` + `isinstance` + `try/except`, where every sibling router (`run.py`, `engines.py`) takes the same `{py, html, params}` body via FastAPI's `Body(...)` with no manual parsing.

https://github.com/fusedio/fused-render/blob/6878cacda67befdaaa09f2219d6ea70813107b69/fused_render/server/routers/engine.py#L39-L44

PR #771, opened 2026-08-23
Playground: the stages' shared surface rebuilt — cog settings, sample pills, image editing, download and memory facts

vasu2001:
Nine passes over the five playground stages' shared surface (`controls.tsx` + `ai-playground.css`), so every stage moves together.

**Three D rows, not nine** (D464, D465, D467). Six were written and then dropped: a row that restates its own diff is noise in a file a fresh session is supposed to be able to continue the project from, so only the passes that constrain future work definitively keep one — the layout traps, the sample-row rules, and the file-picking policy. The rest of that reasoning lives in comments at the code it explains. D430's "rejected: a gear icon on the summary" is amended, since that clause is the one thing here a future session would otherwise believe.

## 1. Config to a cog, starters out of the empty state

- `AdvancedPanel` splits into `StageHeader` (title + a settings cog right-aligned on the same row, `aria-expanded`, lit while open) and a controlled `ConfigPanel` driven by each stage's own `configOpen`, unmounted while closed.
- The chevron `<details>` summary under the input is gone. This reverses D430's "rejected: a gear icon on the summary" — the caret did carry the affordance, but it sat in the one slot the examples needed, so D430 is amended rather than left to mislead.
- Starter prompts leave the centered empty state for one horizontal row directly under the input — no "TRY ONE" label. Still shown only until there is a result to read. (Pass 4 turns that row into pills.)

The old fold sat *below* the input, so on the stages with starters the column read input → Config → examples → result: the settings and the examples fought for the same slot and the one wanted first came second.

Transcribe was included unasked — it shares `controls.tsx`, and a lone chevron-Config on one of four stages is two dialects.

## 2. Image stage: a 480×272 16:9 default, full-width result

- `DEFAULTS` moves off 512², the `16:9` chip is re-footed onto that exact pair so a fresh stage lights a chip, and `shot.width` drops `min(80%, 0.8·w px, 50vh·ratio)` for a flat `100%` with the run's ratio supplying the height.
- **272, not the 275 asked for**: the image route floors every side to a multiple of 16 (`side - side % 16`, `ai_runtime.py`), so a 275 default would render 272 while the slider, the chip title and the URL all said 275.
- Accepted cost: a portrait render is now tall — 9:16 at a 680px column is ~1200px high, where the dropped `50vh` term used to cap it. The wide default is what makes that the uncommon case.

## 3. The settings card leaves the column, and the pair centers (D464)

- `.pg-work` grows a second grid column: a 240px track holding a `.pg-config-card` rail absolutely positioned into it (`inset: 0`), with the visible card (`.pg-config-inner` — bordered, `--bg-alt`, a SETTINGS caption) sticky inside the rail so it holds while a long result scrolls past.
- The track reserves the width, the abspos rail keeps the card out of the row sizing — an in-flow card would size its own grid row and shove the composer down by its whole height. (`grid-row: 1 / -1` is not an option: negative line numbers resolve against the explicit grid, and there are no explicit rows.)
- **Final placement: the pair centers, the column never resizes.** The input column keeps its 680px and slides left by half the card's footprint, so column + gap + card is centered on the stage. Clicking the cog changes no width at all — nothing inside the column rewraps, where an earlier gutter version let the column shrink to ~480px and rewrapped the prompt, the result and the pills on every toggle.
- Gated on `@container (min-width: 1000px)` over a `container-type: inline-size` `.pg-stage` — 936px for the pair plus 32px of edges. The viewport is a poor proxy (shell sidebar + `cc-main` padding + the 300px model rail take ~630px before the stage gets any). Below the threshold the same card renders in flow under the input, as in pass 1.
- The model banner is the fixed landmark: it widens to the pair's own width (`min(680 + gap + card, 100% - 32px)`) on a 16px gutter, so its edges are where the pair's edges land and it does not move when the cog opens. An intermediate pass slid the banner with the column, which jumped the model's name 128px sideways per click.
- Two notes for anyone editing the block: `--pg-card` / `--pg-gap` are declared on `.pg-stage` *outside* the container query, because a `@container` rule can never target its own container. And `.pg-config-card`'s doc comment closed early, leaving three lines of prose as stray CSS; fixed here.
- The apps strip at the foot of the stage now carries the banner's own `min()` expression — the expression, not a copied number — so the two bands bracketing the work column cannot drift apart, and its heading drops caps + tracking for 13px sentence case ("Use it in an app" is an invitation, not a filing category).

## 4. Example prompts become a rotating row of sample pills (D465)

- An authored sample is `{ name, icon, prompt }` (`Starter`), drawn as a hug-the-content pill — icon plus a two-or-three-word name, no fill, round rotate button immediately after the last pill. The detailed prompt lives in the hover title.
- **Eight per stage, four on screen**, rotate steps the window modularly through the rest (no shuffle, so clicking back around lands on the same pills).
- **The count is measured, never truncated**: a layout effect compares the row's `scrollWidth` to its `clientWidth` and drops one pill at a time (floor of two); a `ResizeObserver` re-asks for the full four when the column grows back.
- The prompts themselves are rewritten to be worth running: each asks for a length, a format or an exclusion rather than naming a topic.
- One shared 19-glyph icon vocabulary (`starterIcons.tsx`, capabilityIcons' grammar) rather than one drawing per sample.
- `StarterCards` is generic over the sample, so a pick can set more than a prompt: **the embed stage's samples are whole scenarios** — the query plus the six lines it is searched against, each set built so the three semantic matches share no word with the query. `run()` takes both as arguments, since state set in the same click is not readable yet.
- The row belongs to the four stages that take a prompt (text, image, embed, and video as of pass 6). Read-aloud scripts for the transcribe stage were built and then removed by request: its input is a recording, and a row of suggested sentences beside a record button answers a question nobody with a file to drop was asking.
- `.pg-starter-row` / `.pg-starter-chip` deleted.

## 5. A three-line composer, Clear above Run beside it

- Reported with a picture: a three-line prompt in the image stage had become a scrolling two-line box. `grow()` wrote an inline px height from `scrollHeight` and only ran `onChange`, so when Clear appeared the moment a picture landed, the textarea lost Clear's width, the same text rewrapped a line taller, and the stale height turned it into a scroller.
- Fixed at both ends: the buttons move into a `.pg-composer-side` column (Clear at the top, Run at the foot, as wide as Run) so the input's width no longer depends on whether a result exists, and the two hand-rolled `grow()` copies become one `useAutoGrow()` that re-measures on a **width** change only — reacting to height would loop on what the callback does. `field-sizing: content` would delete the hook outright and is deliberately not used — Chromium-only, and this sheet is not a Chromium-only sheet; noted on `useAutoGrow` itself, since that is where somebody would go to delete it.
- The composer opens three lines tall (a `min-height`, not `rows` alone — the inline height would collapse it back). A `min-height: 72px` floor on the column reserves both slots, or the embed stage's one-line input would jump taller the moment a result gave it a Clear. The column is exactly its two buttons tall, so a tall prompt no longer strands Clear adrift mid-box, and Clear holds 8px off Run's right edge — flush, a bare label squared against a filled button's corner reads as a mis-set pair.
- The recording toolbar's Clear opts out of the top alignment: that row is a horizontal toolbar around an `<audio>` element, not a composer.

## 6. Hand the image stage a base image to edit (D467, SPEC AI-11j)

Some image models can be given a picture to edit, not just a prompt (AI-9f, mflux-only) — and nothing in the app could reach that.

- The composer grows two pills beside Generate: **Add an image**, which raises the OS file dialog and points the render at the path it returns, and **Webcam**, which opens a viewfinder under the prompt and writes the frame it captures.
- **Which models take one is the server's answer**: `/api/ai/catalog` computes a per-entry `acceptsImage` from the resolved engine (`engine_options`) and then mflux's own edit recipe — the two gates `/api/ai/image` refuses with — so the affordance cannot appear where the request would 400. False on every non-image capability rather than True by vacancy, since the engine table is an exception list. `tests/test_ai_runtime.py` asserts the flag and the endpoint agree per entry rather than pinning repo ids on either side.
- **A picked file is never copied.** A browser strips the path from `<input type=file>`, so `dirpicker` gains `pick_file` behind `POST /api/fs/pick-file` (and `menubar_pin` a file-only `NSOpenPanel`), sharing one body with the folder chooser; `files` is passed only when true, so every existing backend caller still sees the two-argument signature.
- **An edit renders at the picture's shape**, scaled to a 640 longest side, snapped down to a multiple of 16, never upscaled, floor 256 — the settings panel shows the numbers that actually run, and "Set a size" is still the way to a full-size render. With the size pair left off the route fits the longest side to 1024 (~4.5× this stage's own default pixels), which is why an attached screenshot took minutes. A probe that has not answered falls back to the server's derivation, never to 480×272: a slow render beats a squashed one.
- Webcam frames go to `<cache>/image-playground`, not the user's home: `/api/config` gains `cache_dir` (branch-aware `storage.home_dir()/cache`) as the one answer for shell code that has to write scratch files. The stage mkdirs it and the subdir, since `/api/fs/mkdir` makes one level by design.
- Composer chrome: Clear floats absolutely in the box's top-right corner (10px in from both edges) instead of costing a permanent 40px row, the prompt reserves a permanent 64px lane on its right so text can never run under it (a lane that came and went would rewrap the prompt and strand `useAutoGrow`'s inline height), the attachment is a 28px thumbnail on the composer's floor beside Replace/Webcam/Generate — a button that opens the picture full size, one ✕, backdrop or Escape to close — and the viewfinder's Cancel is centred under Capture in the error colour.
- **Three of the samples bring their own photo** (Ghibli coast, Watercolour bridge, Pencil study — Unsplash's, re-encoded at 896×597 with metadata dropped into `fused_render/static/samples`, credited in a `CREDITS.md` beside them). The eleven examples are one pool, shuffled once at module scope rather than per render (three photo pills at the head read as a separate section; shuffling per render would be the slot machine D452 refused). A photo pill runs on click like every other one: what to edit is passed to `generate()` rather than read back out of state the same click just set, and the picture is measured off its blob so the first edit is fitted to its own shape. Clear drops the attachment too — it is part of the request, not the setup.

## 7. The video stage joins the cog design and loses its rail

`VideoStage` landed on main while this branch was rewriting the other four stages, so it arrived written against the vocabulary the branch removes. `StageHeader` + `ConfigPanel` replace the `.pg-rail` aside, the three-string `STARTERS` become eight `Starter` pills in the shared row, and the centered empty state goes.

Video was the last consumer of `RailSection`, `StarterPrompts`, `.pg-work-video`, `.pg-main`, `.pg-rail*`, `.pg-under`, `.pg-empty-*` and `.pg-starter-chip*`, so all of them are deleted rather than left standing.

Kept on purpose, and the reason this is not a mechanical port: the settled caption (frames are rounded to the engine's own grid, so what ran is not what was asked for), its seed button (an hour-long render nobody can reproduce is one nobody can iterate on) and the gallery strip.

## 8. A download counts where its button was, and the card answers "will it fit"

The stage header's Download button and the sidebar row's download glyph both vacated their slot the moment a pull started, so the corner the eye was already on went blank at the moment there was most to say.

- **Sidebar row**: the glyph becomes a 14px progress ring in place, sized so the row does not change shape. No total yet spins a quarter arc rather than sitting at zero, which reads as stalled.
- **Stage header**: download icon + the same ring + "done / total". Deliberately not the shared `ModelProgress` drawing — its pulsing dot and job sentence want a card's width, and this is a slot in a row of buttons.
- One `downloadFraction(job)` behind the ring, the byte line and both titles, so they cannot disagree about one download.
- **Pointing at the byte figures offers a Cancel.** A download is the one thing on this card a reader changes their mind about, and the figures are what they are looking at while doing it — so the way out lives under the pointer that is already there, not in a fourth button on a row read while nothing is downloading. Progress and Cancel share one grid cell (sized by the wider of the two, so the hover repaints instead of reflowing) and swap by opacity, not `display`: a display-toggled button cannot be focused, so `:focus-within` would never fire and the way out would be pointer-only. Offered only when the download manager says the job is cancellable and has not already been asked — the rule `CancelButton` applies on the Local tab. The button is a pale word at the figures' own 12px, error hue at 85% with an underline; the icon stays put, since it says which kind of work the slot is about.
- **`_fit_verdict` is read again**, as a fourth `.pg-hero-fact` ("Memory"). It had been on the wire unread since the state line carrying it was deleted, so the card named the download in gigabytes and said nothing about whether the thing would run. **All three verdicts are worded** — "Runs comfortably here" / "Tight fit on this machine" / "Likely too big for this machine" — because warning-only made it unreachable: on a 34GB machine every model the sidebar offers is "easy", so the one fact answering "can this run here" appeared nowhere. Only `null` draws nothing, because a verdict over an unmeasured size is a lie.
- The two verdicts that ask the reader to act differently take `--warning` and `--error` on the text and its underline, not a filled badge that would out-shout the figures. **"Runs comfortably here" takes no green on purpose**: green on this page means *running* (the live dot, the loaded badge, D461), and a second meaning for one hue on nearly every model dilutes the one that matters. No dotted underline either — under a coloured word it read as a defect, which is a high price for a mark pointing at a tooltip that only elaborates.

## 9. A capability switch clears the query string

The stages share one query namespace and each nulls only the keys it owns, so switching capability carried the abandoned stage's tuning over — and `prompt`, `steps`, `seed` and `w`/`h` are spelled the same in several stages while meaning different things in each. A cross-capability pick now rewrites the URL as `?model=<id>` and nothing else (`resetParams`, a keep-list of one); a pick within a capability keeps everything, so two image models can still be compared at one size and seed. Written as a keep-list rather than a clear-list: the list of keys to *drop* is what goes stale when a stage gains a parameter.

The embed stage also loses its prefilled query and its intro paragraph — the starter pills say both, about a concrete example.

## Fixes and polish carried along

- **`think.ts`**: a `</think>` nobody opened still hides the reasoning. Two models on the Models tab close a thinking block they never open — Macaw-OptiQ's chat template prefills `<think>` into the generation prompt (so the tag is read, not written), LFM2.5 never emits one at all. `splitThink` demanded the pair, so both leaked the whole trace plus a stray tag into the answer body. A closing tag with nothing opening it now means the block was open from the first token (vLLM's rule; llama.cpp's `thinking_forced_open`), and an empty closed block discloses nothing rather than drawing an empty "Thought process". Moved out of `TextStage.tsx` so the rule has tests.
- **`client.withModelReady`**: a model that loses its slot mid-load is waited for, not reported. "`<model>` is still loading (loading)" was not a broken run — the stage retried AI-5's cold-start 409 once, and one model per capability is resident, so any other surface asking for a different text model evicted ours between the job row going done and the retry landing. Asking again wins. Bounded at four rounds, because two surfaces trading the slot forever is worse as a spinner that never resolves; a 409 with no row to watch paces itself. Lifted out of `TextStage` and `EmbedStage`, which each carried the same dance. The image stage needs none of it — `generate_image` waits inside the render job server-side.
- **The text stage ships no system prompt.** `DEFAULT_SYSTEM` is deleted and `system` starts empty, so a run sends exactly what a bare `fused.ai` call sends. The old four-sentence default was a second author in every reply — verbosity, formatting and reasoning length arrived steered with nothing on screen saying by whom, which is the wrong footing for a page whose reader is comparing two local models. `system=` now rides the URL only when someone writes one, and the field's reset reads "clear". No D row: emptying a default is not a decision (the owner's call); the reasoning sits in a comment where `DEFAULT_SYSTEM` used to be.
- **Catalog**: the three video entries carry their parameter count and quantization, which shipped empty — so the model card showed two of its four facts blank on every video model. Filled from the publishers per AI-2c (LTX-2.3 int4/int8 named "MLX 4-bit"/"MLX 8-bit" to read down the same column as the other MLX rows; MiniMax H3's own "33B-parameter dense, single-stream Transformer", no quantization, since these are the published bf16 weights).
- **The video section gets Home's video glyph.** `capabilityIcons` had no text-to-video entry, so the rail drew the FALLBACK — the mark meaning "a capability this build has not learned yet" — over a section with two working models in it. Copied rather than shared, for the reason that file's header gives.
- **The transcribe stage's uploads land in the app's cache too**, `<cache>/transcribe-playground` rather than `~/recordings` — same rule as the webcam frame above: `~/recordings` is a folder the user browses, holding the capture feature's own takes, where a playground upload is a file nobody can tell from one they made on purpose.
- The starter pills lose their hover fill.
- **The embed stage's search starts at the top of its composer.** `align-items: flex-end` is a rule about a growing box; an input does not grow, so the search text sat on the button column's 72px floor with an empty third of the card above it.

## Checks

`npx tsc --noEmit` clean. `node scripts/check-boundaries.mjs` clean (346 files). `bun test` over the new/changed frontend units — `think.test.ts`, `imageInput.test.ts`, `client.test.ts`, `lib/params.test.ts` — 41 pass. `pytest tests/test_ai_runtime.py` — 434 pass. `pytest tests/test_doc_duplicate_ids.py` — 3 pass (no duplicate or dangling doc ids after the row removals). Visual smoke is the owner's.


vasu2001:
Added on top: the image stage can now be handed a base image to edit (D454, SPEC AI-11j).

- `/api/ai/catalog` computes a per-entry `acceptsImage` — resolved engine (`engine_options`) then mflux's own edit recipe, the two gates `/api/ai/image` refuses with — so the attach affordance only appears where the request would succeed. False on every non-image capability.
- Composer: **Add an image** + **Webcam** pills beside Generate; the attached photo and the live viewfinder sit under the prompt.
- A picked file is **pointed at, never copied**: `dirpicker.pick_file` behind `POST /api/fs/pick-file` (+ a file-only NSOpenPanel in `menubar_pin`), since a browser strips the path from `<input type=file>`. `files` is passed to a backend only when true, so every existing caller/stub still sees the two-argument signature.
- A webcam frame is the one thing written — `<home>/ai/inputs`, both levels mkdir'd (`/api/fs/mkdir` makes one level by design).
- With an image attached the stage stops sending `width`/`height` (the server derives them from the base image) and the size controls collapse to one line with "Set a size" as the way back.

Owner has smoke-tested the pick + attach path in the running app; a webcam capture and an actual edit render still want a pass.

PR #780, opened 2026-08-24
Text embeddings as a capability of their own: llama.cpp (CPU + Vulkan) and MLX

amanbagrecha:
Text embeddings — bge, e5, nomic-embed, EmbeddingGemma, Qwen3-Embedding — as a
capability of their own, on llama.cpp (CPU + Vulkan) and MLX.

## Why a separate capability rather than widening `embeddings`

The existing `embeddings` capability is a dual encoder: SigLIP/CLIP, image and
text into one vector space, gated by `EMBED_MODEL_TYPES = {"siglip", "clip"}`.
A text encoder is a different thing — no vision tower, no
`get_text_features`, a pooling configuration instead — and today it cannot load
at all. Passing `BAAI/bge-small-en-v1.5` to `/api/ai/embed` currently downloads
~400 MB, starts a worker, and only then dies with
`AttributeError: 'BertModel' object has no attribute 'get_text_features'`.

`registry.py` already anticipated this, in the comment explaining why the Hub's
`feature-extraction` / `sentence-similarity` labels are withheld:

> Mean-pooling a text-only encoder is a different load path and a different
> catalog; when it ships, these two move.

They move here. Keeping the two capabilities separate also means a SigLIP model
and a text embedder can be resident at the same time — one resident model per
*capability*, so a RAG page and an image-search page no longer evict each other.

**The existing `embeddings` capability is untouched by this branch**:
no diff under `transformers_embed/`, `mlx_embed/` or `embed_common.py`, and
`EMBED_MODEL_TYPES` is unchanged.

## Engines

Following the precedent set when text generation moved off torch
(`llamacpp_text` + `llamacpp_text_vulkan` + `mlx_text`):

| | Runner |
|---|---|
| Windows / Linux CPU | `llamacpp_embed/` |
| Windows / Linux GPU | `llamacpp_embed_vulkan/` |
| Apple Silicon | `mlx_text_embed/` |

The llama.cpp pair carries the same `llama-cpp-python==0.3.29` pin as the text
pair, differing only in the wheel index — so their audit covers these and no new
one was invented. `llama_embed.py` sits at the runners root and serves both
folders, exactly as `llama_text.py` does.

## What an embedding GGUF is, and how it is told apart

Not by architecture. `Qwen3-Embedding-0.6B` and Qwen3 chat both declare
`general.architecture = qwen3`; only the embedding model carries a pooling key.
So `is_embedding_gguf` turns on **`pooling_type`**, read from the GGUF header —
an architecture allowlist could not have worked. The header readers take bytes,
so they run against a 2 MB HTTP Range peek rather than a downloaded file, which
is what makes the refusal path cheap.

## API

`POST /api/ai/embed-text` and `fused.ai.embedText({texts, model, kind})`,
matching the existing embed route's contract — including the cold-model fork
that answers `409 {"type":"model_loading","jobId":...}` rather than failing, and
unit-normalised vectors so cosine stays a plain dot product.

`kind` is the one genuinely new piece of surface. Retrieval models are
asymmetric — e5 wants `query:`/`passage:`, bge instructs the query only,
Qwen3-Embedding ships named prompts — and getting it wrong degrades retrieval
silently rather than failing. It defaults to `"document"`, the side that keeps a
corpus self-consistent for someone who never reads about prompts (and on bge
means no prefix at all). An unrecognised `kind` is **refused**, because it is the
one field whose wrong value is undetectable downstream.

## Refusing before the download

A repo that is not a loadable text embedding model is refused with a sentence
naming the repo, what it looks like and what to pass instead — **before** any
weights are fetched. Verified live against the Hub: a safetensors repo refused
off a 30-file listing, a chat GGUF refused off a 2 MB header peek, neither
downloading weights.

## Verified vs not — please read this part

**Actually run:** the GGUF header readers against four real published headers
over HTTP Range (architecture, pooling, context length all correct); the full
cold-load pipeline through a running server — env build, file resolution, a
146 MB download with progress, reaching `Llama(...)`; both refusal paths live
against the Hub; every route status and shape; the demo page rendering,
catching the 409, watching the job and surfacing the error.

**Not run — `llama.cpp` inference itself.** The pinned wheel aborts the process
with `0xc000001d` (illegal instruction) on the development machine's AVX2-only
i5-8365U, at every version tried (0.3.29, .26, .18, .12). Retrieval was
therefore validated by driving the *shipped test bodies* with real
`bge-small`/`bge-base` weights through transformers: all four assertions pass,
the prompt scheme resolves correctly, and the corpus scores 5/5 against a 2/5
keyword baseline. That proves the logic, not the llama.cpp call.

**Not run — MLX and Vulkan.** No hardware. `TESTING.md` §6 lists the three MLX
seams to check on a Mac.

## A pre-existing bug this surfaced (not fixed here)

`_llamacpp_platform` reports available for any Windows AMD64, but the pinned
wheel ships one `ggml-cpu.dll` with no runtime feature dispatch. A pre-AVX-512
x86_64 machine therefore gets a Load button and a hard process abort — which is
exactly what happened here. **This affects the existing `llamacpp-text` runner
identically**; it is untouched by this branch and wants its own fix.

## Tests

Targeted rather than a full-suite run, deliberately — the suite has pre-existing
failures unrelated to this change.

- New: `test_ai_text_embed_common.py`, `test_ai_llamacpp_embed_worker.py`,
  `test_ai_runtime_embed_text.py`, `test_ai_text_embed_retrieval.py` —
  **73 passed, 3 skipped** (skips are the opt-in llama.cpp venv).
- All 21 touched and neighbouring modules — **948 passed, 8 skipped**.
- Frontend `bun test src/apps/ai_models/lib/` — **333 pass**.
- 10 pre-existing failures baselined by stashing onto the branch point: nine
  Windows symlink-privilege errors and one metrics error-type expectation, none
  in a module this branch edits. Recorded in `TESTING.md`.

## Trying it

```bash
uv venv && uv pip install -e ".[dev]"
./.venv/Scripts/python.exe -m pytest tests/test_ai_text_embed_common.py \
  tests/test_ai_llamacpp_embed_worker.py tests/test_ai_runtime_embed_text.py \
  tests/test_ai_text_embed_retrieval.py -q
```

`TESTING.md` §2–4 cover running the server, exercising both refusal paths, and
the demo page at `docs/text-embeddings-demo.html`.

## Note on merging

This branches from `a014e786` and does **not** merge cleanly with current
`main`: `text-to-video` landed in the meantime and adds a capability in the same
places — the capability constants block, the runner rows, the catalog lists. The
conflicts are additive (both sides add a capability) rather than contradictory,
but reconciling two brand-new capabilities is a judgement call for whoever owns
the video work, so it has been left rather than resolved unilaterally. Happy to
do the merge on request.

---

> [!NOTE]
> **Medium Risk**
> Adds a new AI capability, runners, Hub format classification, and a public embed API. Existing dual-encoder embeddings are left intact, but GGUF/safetensors loader rules and Hub task mapping can change how models are tagged and loaded.
> 
> **Overview**
> Adds **text embeddings as their own capability**, separate from dual-encoder `embeddings`, so a retrieval encoder and SigLIP can stay resident at once.
> 
> Pages get `POST /api/ai/embed-text` and `fused.ai.embedText({texts, kind, model})`. Vectors are unit-normalized; `kind` is `"query"` | `"document"` (default document) with the model’s prompt scheme applied and reported back. Image `paths` and unknown `kind` values are refused rather than silently mis-embedded.
> 
> Engines follow the chat split: **llama.cpp CPU**, **Vulkan**, and **MLX** (different catalogs — GGUF vs safetensors). Embedding GGUFs are identified by header `pooling_type`, not architecture, so Qwen3-Embedding is not treated as chat. Uncurated repos are refused from a Hub listing or a 2MB Range peek **before** weights download.
> 
> Hub tags `feature-extraction` / `sentence-similarity` now map here; llama.cpp still filters Discover to GGUF. Catalog, benchmarks, and the AI Models UI pick up the new capability. MLX inference is implemented against `mlx-embeddings` 0.1.0 but not executed on Apple Silicon in this branch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1345135dd914e227fef076a78cd0e72397b6980c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,270 @@
+# Testing `text-embeddings`
+
+A new capability (`registry.TEXT_EMBEDDINGS`) that turns text into vectors
+through a text encoder — bge, e5, nomic-embed, EmbeddingGemma,
+Qwen3-Embedding. It is **separate from the existing `embeddings`
+capability**, which stays exactly as it was: that one serves dual encoders
+(SigLIP/CLIP) and can put an image and a sentence in one space; this one
+serves the retrieval models a RAG or search page actually wants. They hold
+separate resident-model slots, so you can have one of each loaded at once.
+
+Engines: **llama.cpp** on Windows/Linux CPU, **llama.cpp + Vulkan** on
+Windows/Linux GPU, **MLX** on Apple Silicon.
+
+---
+
+## 1. Run the tests
+
+Everything here runs with no model downloads and no runner venvs.
+
+```bash
+cd /path/to/wt-text-embeddings
+uv venv
+uv pip install -e ".[dev]"
+
+./.venv/Scripts/python.exe -m pytest \
+  tests/test_ai_text_embed_common.py \
+  tests/test_ai_llamacpp_embed_worker.py \
+  tests/test_ai_runtime_embed_text.py \
+  tests/test_ai_text_embed_retrieval.py -q
+```
+
+On POSIX use `./.venv/bin/python` instead.
+
+Expect **~65 passed, 3 skipped**. The three skips are
+`test_ai_text_embed_retrieval.py`'s model-backed tests, which need real
+weights — section 4 turns them on.
+
+The tests this change also touched, all of which should be green:
+
+```bash
+./.venv/Scripts/python.exe -m pytest \
+  tests/test_ai_formats.py tests/test_ai_registry.py tests/test_ai_tasks.py \
+  tests/test_ai_runtime.py tests/test_hub_models.py \
+  tests/test_ai_runtime_embed.py tests/test_ai_runner_deps.py -q
+```
+
+> **Known unrelated failures.** `tests/test_ai_models_api.py` fails 4 tests on
+> Windows without Developer Mode (`OSError: WinError 1314` — the fixture
+> creates symlinks). Those fail on `origin/main` too; verified by stashing.
+
+---
+
+## 2. Run the server from source
+
+The installed desktop app is far behind — you must run from the worktree.
+Do **not** try to reuse another checkout's `.venv` with `PYTHONPATH`: the
+editable install uses a PEP 660 meta-path finder that beats `PYTHONPATH`, so
+you silently get the other checkout's code.
+
+```bash
+cd /path/to/wt-text-embeddings
+uv pip install -e ".[fused]"     # the [fused] extra pins fused==2.9.3b7;
+                                 # without it NO model can load
+
+# The React shell must exist, or create_app raises at startup:
+(cd frontend && npm install && npm run build)
+
+FUSED_RENDER_CORE_TEMPLATES="$PWD/fused_render/templates" \
+  ./.venv/Scripts/python.exe -c \
+  "import sys; sys.argv=['fused-render','--port','1789','--no-browser']; from fused_render.cli import main; main()"
+```
+
+`FUSED_RENDER_CORE_TEMPLATES` short-circuits template staging so in-repo
+templates are read live — without it the server fails on a missing `vendor`
+directory. Port 1777 is the desktop app; 1788 is often taken.
+
+### The capability should now be visible
+
+```bash
+curl -s http://127.0.0.1:1789/api/ai/runtime -H 'X-Fused: 1' \
+  | python -c "import json,sys; [print(r['code'], r['capability'], r['available'], r['active']) for r in json.load(sys.stdin)['runners'] if r['capability']=='text-embeddings']"
+```
+
+Expected on Windows/Linux x86_64:
+
+```
+mlx-text-embed          text-embeddings False False
+llamacpp-embed          text-embeddings True  True
+llamacpp-embed-vulkan   text-embeddings True  False
+```
+
+And the catalog, smallest first, default first:
+
+```bash
+curl -s http://127.0.0.1:1789/api/ai/catalog -H 'X-Fused: 1' \
+  | python -c "import json,sys; [print(r['default'], [m['id'] for m in r['models']]) for r in json.load(sys.stdin) if r['capability']=='text-embeddings']"
+```
+
+---
+
+## 3. Exercise the endpoint
+
+### The refusals — these need no model and no network
+
+```bash
+E=http://127.0.0.1:1789/api/ai/embed-text
+J='-H Content-Type:application/json -H X-Fused:1'
+
+# 403 — the guard every mutating POST carries
+curl -s -o /dev/null -w '%{http_code}\n' -X POST $E \
+  -H 'Content-Type: application/json' -d '{"texts":["a"]}'
+
+# 400 — images are meaningless here, and the message says where to go instead
+curl -s -X POST $E $J -d '{"texts":["a"],"paths":["/photos/cat.png"]}'
+
+# 400 — an unrecognised `kind` is refused, never silently defaulted
+curl -s -X POST $E $J -d '{"texts":["a"],"kind":"queries"}'
+```
+
+### The cold-model fork
+
+```bash
+curl -s -X POST $E $J -d '{"texts":["hello world"]}'
+```
+
+```json
+{"ok": false, "error": {"type": "model_loading",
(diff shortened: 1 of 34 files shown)


iamsdas:
Thanks for this, @amanbagrecha — and sorry for the long silence on it. Rather than review it in place I want to explain where we've landed, because we've built something on top of the problem you identified here and I'd rather you hear the reasoning than just see a competing PR appear.

## Your diagnosis was right

The thing this PR is actually about — that our embedding capability could not do text retrieval — is real, and we had not noticed it. The capability today is served by SigLIP2 dual encoders, whose text tower is capped at **64 tokens**. That isn't a small chunk size, it's a wall: there is no chunking strategy that makes RAG work against a 64-token encoder. So "we already have embeddings" was not an answer to what you raised, and the prompt-scheme work in `formats.py` (the per-family query/document prefix table for bge / nomic / e5 / qwen3 / gemma-embedding) is the part of this that most needed doing. Getting that wrong silently produces vectors that look fine and retrieve badly.

## What we're changing, and why it's a different shape

We're keeping the substance and changing the structure, in one PR off `main` (branch `worktree-unified-embeddings-onnx`).

**1. One capability, not two.** This PR adds text embeddings as a capability of its own, sitting beside the existing dual-encoder one. We think that's the wrong seam. It's the same seam `TEXT_GENERATION` already resolved the other way: that capability has an optional image parameter gated per model and per engine (`_accepts_image`, computed off the cached `config.json`, never curated), because `image-text-to-text` and `text-generation` models are one capability with a model-dependent surface — not two capabilities. Embeddings are the same. A prose encoder takes text; a dual encoder takes text *or* images; `jinaai/jina-clip-v2` takes both into a single 1024-dim space and is tagged `feature-extraction` exactly like bge is. Two capabilities would put a hard boundary through the middle of a model family that already has members straddling it, and would force a user to know which capability their model belongs to before they can pick it.

So: one `embeddings` capability, with `paths` and `kind` gated per model the way `image` already is for text generation. `feature-extraction` and `sentence-similarity` map onto it.

**2. ONNX Runtime rather than llama.cpp, and the reason is the vision tower.** This isn't a judgement about llama.cpp — we use it as the default for text generation. It's that llama.cpp cannot run a vision tower, so it can serve *half* of a unified embeddings capability and no more. We'd have shipped one engine for prose and a second for dual encoders, permanently. ONNX Runtime serves both from one runner, and `onnx-community` already publishes exports for everything we curate.

It also lets us do something we've wanted for a while: **delete the `transformers-embed` family entirely.** Those three runners (CPU, CUDA, ROCm) pull torch to run one forward pass of a 375M-parameter encoder — 1.86 GB for cu130 on Windows, 5.86 GB for rocm7.1 on Linux, plus ~1.17 GB of `nvidia-*` wheels. That download is the single worst first-run experience in the product. onnxruntime is ~50 MB. MLX stays as the Apple Silicon path; the four new ONNX rows (CPU, DirectML, CUDA, ROCm) sit below it.

**3. Your prompt-scheme work is carried over with attribution.** The commit that ports the per-family query/document prefix table carries `Co-authored-by: Aman Bagrecha`. It's your design, and it survives intact — same families, same `query`/`document` kinds. We read the pooling mode off `1_Pooling/config.json` rather than assuming a masked mean, because bge pools CLS while e5 and nomic pool the mean and all three publish the same `last_hidden_state`.

## Where that leaves this PR

We'd like to supersede it rather than merge it, and we'd rather you make that call than have us just close it. If you disagree with the single-capability read, or you think there's something in the llama.cpp path we're losing, say so here — the branch isn't merged and the argument above is the whole basis for the decision, so it's worth contesting if it's wrong.

One place we'd genuinely value your input: the ONNX-vs-torch vector parity gate that licenses deleting the torch runners has been written but never executed, because it needs an interpreter with onnxruntime, torch and transformers plus both snapshots cached. If you have a machine that can run it, `CONTRIBUTING.md` on the branch has the invocation.

Either way — thanks for finding this. The capability was broken in a way we'd have kept not noticing.

PR #784, opened 2026-08-24
Tokenize the Playground lightbox and viewfinder colours, unblocking main

iamsdas:
## Summary

- **Turns `main` green.** `test_theme.py::test_shell_css_has_no_colour_literals_outside_the_palettes` has been failing since `ebf94d23` (#771, the stages' shared-surface rebuild), which landed six raw colour literals in `frontend/src/styles/ai-playground.css`. That single assertion is what reddens **all four `test-python` lanes, `test-python-windows` and `fused-engine`** — each of them reports `1 failed, 9270 passed`.
- **All six colours belong to the scrim set**, the one token group `tokens.css` keeps deliberately identical in both palettes, because what sits underneath is a photograph rather than the page. They're now named tokens with unchanged values — the same move the neighbouring `.pg-image-save` rules in this very file already made, with a comment explaining why.
- **No pixel changes, in either theme.** Every alpha is carried across exactly as authored.

## Visuals

The failing test's premise is that light mode must be able to repaint every colour. For a control on arbitrary photo pixels the answer isn't a themed value — it's a token that's *intentionally* the same in both palettes, so the rule keeps no exceptions:

```mermaid
flowchart TD
    A[colour in a stylesheet] --> B{painted on the page,<br/>or on a photo?}
    B -->|page| C[themed token<br/>light palette overrides]
    B -->|photo| D["scrim token<br/>same in both palettes"]
    E[raw literal] -.->|what #771 left,<br/>this PR removes| F[test_theme fails]
```

## Usage

Nothing user-invocable changes. Five tokens are added to the scrim set in **both** palettes:

| Token | Value | Paints |
|---|---|---|
| `--scrim-veil` | `rgba(0, 0, 0, 0.72)` | the lightbox backdrop |
| `--scrim-lift` | `rgba(0, 0, 0, 0.5)` | the picture's shadow on that veil |
| `--scrim-chip-bg` / `-hover` | `rgba(255, 255, 255, 0.12)` / `0.22` | the lightbox ✕ |
| `--scrim-letterbox` | `#000000` | behind `<video>`, in the webcam viewfinder |

Two judgement calls worth a reviewer's eye:

- **`--scrim-veil` is not `--scrim`.** The existing modal backdrop is `0.55` dark / `0.4` light; a lightbox has to make the photograph the only thing in the room, so it keeps its heavier `0.72` and does not flip with the theme.
- **`--scrim-lift` is not `--shadow-lg`.** They match in dark mode, but light mode drops `--shadow-lg` to `rgba(0, 0, 0, 0.18)` — nearly invisible against a `0.72` veil, which would quietly delete the separation between picture and backdrop in one theme only.

## Test Plan

- [x] **`tests/test_theme.py` — 133 passed** (was `1 failed`). No new test needed: the guard that was already failing is the coverage, and it fails again on the old literals.
- [x] **`ai-playground.css` was the only offender** — the guard walks every sheet in the `shell.css` barrel and now passes on all of them.
- [x] **`test_shell_css_light_palette_redefines_every_dark_token` passes**, so all five tokens are defined in both palettes — that test is why a dark-only addition can't slip through.
- [x] `tests/test_schedule_css.py` (the other suite reading these two files) — **150 passed** together with the theme suite.
- [x] **Frontend: 2437 pass, 0 fail** across 101 files; `tsc --noEmit` clean.
- [ ] **Worth one glance in the browser:** open the Playground, attach an image, click it to open the lightbox (veil, shadow, ✕ hover), then open the camera — in **both** themes. The values are unchanged, so this is confirming a no-op rather than reviewing a redesign.


iamsdas:
Pushed `beeb45da`: a **second** inherited failure surfaced once the colour-literal one was fixed — `test_doc_duplicate_ids::test_no_new_duplicate_decision_ids`, also red on `main` and also not from this PR.

#771 and #775 each added a **D464** (the Playground settings card, and the dock's read-gated retention clock). Disjoint lines, so the merge raised no conflict — exactly the silent collision that test was written for. The newer row (#775's) becomes **D469**, and `jobs.py`'s cross-reference follows it; `ai-playground.css`'s D464 reference points at #771's row and is left alone.

Not the D466 hole, though it looks free: that number was withdrawn in `66c7de5c` ("Drop six of this branch's nine D rows"), so a citation of D466 means the dropped decision rather than this one.

The 3.10 lane reported `1 failed, 9289 passed` on the previous push, so this should be the last one.

PR #791, opened 2026-08-24
Rebuild the Local AI page around one comparison per pillar

AkshilVT:
Rebuilds `render.fused.io/ai.html`. The first version shipped in #777 was written to a copy deck that mandated *"fragments over sentences"* — that rule is what made it unreadable. A visitor could not say what the product was, and the visuals did not line up with the claims next to them.

## What changed

Every pillar now states its case in plain sentences and then shows one thing that proves it. Each pillar gets a form suited to its own argument, rather than the same card pair four times:

| Pillar | Form |
|---|---|
| **Models** | Before/after ledger — the one pillar whose story really *is* a painful sequence against a short one |
| **Apps** | Four real community apps, each tile linking to its directory in `fused-render-community-apps` |
| **Search** | A hairline strip, plus a search panel and file-format chips |
| **Privacy** | A drawn boundary — `file → model → answer` inside a dashed "your computer" box, arrow out struck through |

Every paragraph that used to sit under a comparison is gone; whatever mattered moved into the comparison itself, where it actually gets read.

## Claims corrected

The first version shipped several that were wrong:

- **It compared against Finder alone** and claimed results come "only from where it decided to look." Spotlight is global, so that was false. The comparison now names **Finder and Spotlight** and argues the real difference — in-folder search, and previewing 140 file types in the same window.
- The GPU narrative ran unqualified. Off Apple Silicon the default is the CPU wheel.
- *"chases an invoice, sorts a drive"* — Invoice Generator **writes** invoices; Disk Space Visualizer **maps** a disk.
- *"across the whole disk"* — the index covers the **home folder**.
- *"Faster than Finder"* removed; the nuanced version is both more defensible and more true.

Numbers are checked against source: **140** file extensions from `fused_render/templates/registry.json`, **19** apps confirmed against the upstream community-apps repo.

## Latent bug fixed

`index.html`'s `application/fused-bundle` manifest listed `assets/*.jpg`. The exporter matches globs **within a single path segment** (`_glob_segments_match` in `fused_render/export.py`), so the showcase images under `assets/showcase/` were never bundled on export. Added `assets/showcase/*.jpg`.

Publishing was unaffected — `release.yml` uses `aws s3 sync`, which is recursive.

## Assets

Showcase tiles re-cropped to the band where each app is visibly doing its job. Two of six dropped — their source previews were an empty state and an abstract gradient, unreadable at tile width. Net **−82KB**.

## Notes for review

- Section rhythm is a single `--gap` token applied as `padding-top` only, so the space between any two sections cannot compound.
- No separators between sections, by design.
- Hard contracts preserved: `#dmg` / `#windows-exe` / `#linux-appimage` / `#version` still filled from a same-origin `fetch("/latest.json")` and work before it resolves; resting CSS is the finished page so a `document.hidden` render is complete; no web fonts, no third-party scripts, no external images.
- Contrast verified by hand (lowest pair 5.30:1) since the design detector ran without its HTML parser modules.
- Verified at 1440px and at the narrowest viewport headless Chrome allows (it clamps to 500px), plus `prefers-reduced-motion`.
- CI: `scripts/download_page/**` is not in `test.yml`'s paths filter, so test jobs are expected to skip.

---

> [!NOTE]
> **Low Risk**
> Static marketing HTML/CSS/JS only; download-link IDs and same-origin `/latest.json` behavior are unchanged. The bundle glob fix is a one-line include, not export-engine logic.
> 
> **Overview**
> Rebuilds `ai.html` so each product pillar is a plain-language claim plus one matching visual, instead of the previous essay of fragments, CSS mock windows, and a sequenced GPU “ladder.”
> 
> The hero now names three jobs (local models, local apps, fast search). Models use a before/after ledger. Apps show four real community screenshots linking into `fused-render-community-apps`. Search compares Finder & Spotlight vs in-window preview (140 types) with a search mock and format chips. Privacy is a drawn “stays on your computer” boundary. Copy is tightened to drop overclaims (Finder-only, whole-disk index, “faster than Finder”).
> 
> Also adds `assets/showcase/*.jpg` to `index.html`’s fused-bundle include so nested showcase images actually export (`assets/*.jpg` does not cross a path segment). Download buttons still hydrate from `/latest.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 486bd5420572d9d1bf4539b625bdaeaa7f7ed874. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AkshilVT:
Closing — superseded by the landing work that has since merged (#826, #847, #860, #872, #875, #885).

PR #813, opened 2026-08-24
Embeddings: fix the transformers 5 pooled-output contract, and add CUDA and ROCm rows

iamsdas:
## Summary

- **Embeddings were broken outright on every install, and this fixes them.** transformers 5 changed `get_text_features`/`get_image_features` to return a `BaseModelOutputWithPooling` where 4.x returned the pooled tensor, so every `fused.ai.embed()` call raised `AttributeError: 'BaseModelOutputWithPooling' object has no attribute 'to'` under the `transformers>=5.15,<6` this runner pins. Both towers, every platform, since the pin landed.
- **Embeddings can now run on a GPU off Apple Silicon.** `transformers-embed` is pinned to `download.pytorch.org/whl/cpu`, so a machine with a working NVIDIA or AMD card embedded in fp32 on the CPU and `_placement()`'s `cuda` branch could never fire. Two opt-in hardware rows now sit below it, gated on the existing `_cuda`/`_rocm` device probes.
- **The runner body is now shared, not copied.** `runners/torch_embed.py` holds it, and the three folders are five-line shells that differ only in which index torch comes from — the `torch_image.py` shape the image family already uses.
- **`formats.loaders()` and `DECISIVE` know about the new rows**, without which they would have been registered-but-unusable: no engine tag, no Load button, no cached repos offered, on precisely the machines that chose them.

No dtype change: `_placement()` still returns a device only and every row runs fp32, so the shared-vector-space property between `mlx-embed` and the torch rows is unchanged. fp16/bf16 was considered and deliberately deferred until the gain is measured on the two suggested models — the frozen `embed-8-texts` benchmark workload is text-only at batch 8 and would understate it.

## Visuals

Resolution for the embeddings capability. `auto` still lands on the CPU row everywhere off Apple Silicon (SPEC AI-2b); the two new rows are reachable only by picking them on the Engines tab, and all three torch rows share one body.

```mermaid
flowchart TD
    A[fused.ai.embed] --> R{registry resolve}
    R -->|auto, Apple Silicon| M[mlx-embed]
    R -->|auto, elsewhere| C[transformers-embed CPU]
    R -->|opt-in, gated _cuda| U[transformers-embed-cuda]
    R -->|opt-in, gated _rocm| O[transformers-embed-rocm]
    C --> T[runners/torch_embed.py]
    U --> T
    O --> T
```

## Usage

Embedding calls are unchanged — this restores them and adds an engine choice:

```js
const { vectors, dim } = await fused.ai.embed({ texts: ["a photo of a cat"] })
// dim 768 on siglip2-base-patch16-384; vectors are unit-normalised
```

To use the GPU, pick the row on the **AI Models → Engines** tab. The accelerated rows appear only where the device is actually present and usable:

| Row | Requires |
|---|---|
| `transformers-embed-cuda` | NVIDIA GPU, driver loaded (Windows or Linux) |
| `transformers-embed-rocm` | AMD GPU on Linux with a gfx in `ROCM_TARGETS` |

Both report what they actually got: `/health` carries the device `load()` placed the model on, so a row that fell back to CPU says so. Note ROCm torch reports through `torch.cuda.is_available()` and wants the device string `cuda`, so `/health` says `cuda` on an AMD card — torch's own naming, since HIP presents the CUDA API surface.

## Test Plan

**Full suite, diffed against a true `origin/main` baseline** (`d88499ed`), both runs on the same machine with the frontend built:

| | failures |
|---|---|
| `origin/main` | 14 |
| this branch | 18 |

The 4-item difference is entirely `tests/test_browse_mount.py`, which is **pre-existing and environmental**: run in isolation it fails **6 of 15 identically on both branch and baseline**. Under xdist a different subset surfaces per run, which is the whole of the delta. No failure is attributable to this branch.

Targeted surface: `-k "ai_ or embed or format or registry or hub_model"` → **1940 passed, 8 skipped, 1 failed**, the one failure being `test_ai_metrics.py::test_a_missing_claude_binary_is_counted`, present in the baseline set and unrelated (it resolves a `claude` binary path on the host).

- [x] `tests/test_ai_transformers_embed_worker.py` — the fake now returns a `BaseModelOutputWithPooling` shape carrying `pooler_output` **and** a deliberately different-rank `last_hidden_state`, so reaching for the wrong field fails on shape rather than passing with nonsense. This is the test that previously encoded the 4.x contract and let the bug ship.
- [x] `tests/test_ai_formats.py::test_every_registered_runner_appears_in_loaders` — the drift this PR would otherwise have caused; fixed via `TRANSFORMERS_EMBED_RUNNERS` plus `DECISIVE` membership.
- [x] `tests/test_ai_registry.py` — the three tests pinning the old two-row/no-variant shape updated, including `test_no_cuda_or_rocm_embed_variant_exists`, which existed to pin the decision this PR reverses. Renamed and re-documented rather than silently deleted.
- [x] Ordering pinned (`mlx-embed → transformers-embed → transformers-embed-cuda → transformers-embed-rocm`), and `test_AUTO_STAYS_ON_THE_UNACCELERATED_ROW_EVEN_WITH_AN_ACCELERATOR` still passes — `auto` must not silently pull a multi-gigabyte accelerated wheel.
- [x] `_SHARED_SUGGESTIONS` aliases asserted for both new rows, so neither can grow a shortlist that drifts from the CPU row's.
- [x] **Real weights, actually executed.** `tests/test_ai_transformers_embed_real_weights.py` skips by default (CI has no weights and must not fetch 1.5GB) and hard-fails rather than skipping under `FUSED_RENDER_REAL_WEIGHTS=1`. Run against the runner venv (torch 2.13.0+cpu, transformers 5.15.1): **2 passed** — both towers return 768-dim unit vectors with `cos(cat,dog) > cos(cat,bicycle)`. Confirmed it still skips cleanly with the flag unset.

**Known gaps, stated rather than implied:**

- **Neither accelerated venv was built.** `uv sync` on these folders is a multi-gigabyte ROCm/CUDA download, so the manifests are verified by reading them against their `diffusers_image_*` siblings and by the manifest tests — not by an actual install. First real install is the thing to watch, particularly the `triton-rocm` routing (ROCm torch pins it transitively and it exists only on the ROCm index).
- **The real-weights test runs on CPU**, so it pins the contract and the semantics, not the accelerated numerics. Confirming those needs hardware CI.
- **In the default suite that file is a local reproducer, not coverage.** Its docstring says so. The standing guard for the `pooler_output` contract is `FakeOutput`.
- **No performance numbers yet.** Measuring the GPU gain — and deciding whether fp16 earns its cost on the two suggested models — is deliberate follow-up.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -5747,14 +5747,20 @@ an AI Models page that could say what was on disk but not what was *running*.
   may occupy SEVERAL folders, one per hardware build of it, and then the shared
   code lives at the runners ROOT and each folder's `worker.py` is a shell over
   it** (D381). Diffusers is three folders — `diffusers_image`,
-  `diffusers_image_cuda`, `diffusers_image_rocm` — and llama.cpp is two
-  (`llamacpp_text`, `llamacpp_text_vulkan`), whose manifests are identical except
-  for the index the accelerated distribution is resolved from, and whose
-  `worker.py` is a `sys.path` insert and a call into `runners/torch_image.py` or
-  `runners/llama_text.py`. (D416 removed a third such family — `transformers_text`
-  and its `_cuda`/`_rocm` siblings over `runners/torch_text.py`, which is where
-  this rule was first written. The rule is unchanged; only one of its instances
-  is gone.) Both halves of that are
+  `diffusers_image_cuda`, `diffusers_image_rocm` — llama.cpp is two
+  (`llamacpp_text`, `llamacpp_text_vulkan`), and the transformers embeddings
+  runner is three (`transformers_embed`, `transformers_embed_cuda`,
+  `transformers_embed_rocm`), whose manifests are identical except for the
+  index the accelerated distribution is resolved from, and whose `worker.py`
+  is a `sys.path` insert and a call into `runners/torch_image.py`,
+  `runners/llama_text.py` or `runners/torch_embed.py`. (D416 removed an
+  earlier such family — `transformers_text` and its `_cuda`/`_rocm` siblings
+  over `runners/torch_text.py`, which is where this rule was first written;
+  the embeddings family above is a LATER, separate instance of the same rule,
+  not that one restored — the text tower's one-forward-pass argument that
+  justified withdrawing `transformers_text_cuda`/`_rocm` never applied to the
+  embeddings runner's image tower, which is genuine batched GPU work at
+  `embed_common.MAX_ITEMS`.) Both halves of that are
   forced rather than chosen. **The folder split is forced by PY-16**: a venv is
   keyed on the folder, so two rows sharing one folder share one environment and
   could not hold two different torch builds; and `_env_install_worker` runs a
--- a/fused_render/ai/catalog.py
+++ b/fused_render/ai/catalog.py
@@ -928,6 +928,11 @@ _SHARED_SUGGESTIONS = {
     "diffusers-image-cuda": "diffusers-image",
     "diffusers-image-rocm": "diffusers-image",
     "llamacpp-text-vulkan": "llamacpp-text",
+    # Hardware variants of `transformers-embed` — same repos, same format,
+    # different wheel — for the identical reason the Diffusers pair above is
+    # aliased rather than copied.
+    "transformers-embed-cuda": "transformers-embed",
+    "transformers-embed-rocm": "transformers-embed",
     # Not a hardware variant of the same runner — a DIFFERENT runner that reads
     # the same repos (see the comment on the embeddings block above). Aliased
     # for the same reason as the pairs above: one list to keep in step rather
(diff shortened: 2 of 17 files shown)


iamsdas:
### CI red is inherited from `main`, not introduced here

This branch is rebased onto `d88499ed` (#814), and **that commit's own CI run on `main` is failing**:

| | failing in CI |
|---|---|
| `main` @ `d88499ed` | `test_theme.py::test_shell_css_has_no_colour_literals_outside_the_palettes`, `test_canvases.py::test_opening_a_canvas_retires_the_legacy_fused_plugin` |
| this branch | `test_theme.py::test_shell_css_has_no_colour_literals_outside_the_palettes` |

So this branch fails a strict subset of what `main` already fails. `test_theme` also fails locally on both this branch and a clean `origin/main` checkout, so it is reproducible and pre-existing — it looks like fallout from the recent shadcn/Tailwind work introducing colour literals into `shell.css`.

Leaving this as a **draft** rather than marking it ready, since the checks are red — but the red is not this PR's. It should go green once `main` is fixed and this is rebased.

PR #877, opened 2026-08-26
Repos update on the user's terms: showcase de-managed, per-repo update notifications

iamsdas:
## Summary

This branch does two things. The first was the original ask; the second grew out of it, because the notification had to live somewhere and that somewhere turned out to need rebuilding.

### 1. The showcase repo stops being app-managed, and every repo says when it is behind

- **`~/Fused/showcase` is no longer synced.** It is cloned once if absent, then treated as an ordinary git repo — no fetch, no `merge --ff-only`, no background refresh on server start. The user owns the checkout, so their work is never silently advanced out from under them.
- **Every app in a git repo reports when it is behind.** Opening an app records a throttled, per-repo-root check against origin's default branch. Behind → a row appears in the status bar offering exactly one action: **Update** (fast-forward) on the default branch, **Switch** to it anywhere else. Both are non-destructive by construction; neither can conflict or touch the user's commits.
- **A dirty tree is refused, never fixed behind the user's back.** No stash, no auto-commit, no reset. The row says what is in the way and offers **Fix with Claude**.
- **Installed community copies are gone.** The `~/Fused/local/<slug>` copy-then-`git init` path, its force-overwriting update, and the Clone button are removed — apps are edited in the showcase checkout itself. Existing installs stay on disk and simply stop being managed.
- **Silence on failure is deliberate.** An unreachable remote, a repo with no `origin`, or a mount-backed repo records nothing and shows nothing. Mount-backed repos are refused ahead of the throttle: a background fetch across an rclone-NFS mount is the exact wedge that refusal exists to prevent (GT-4 / MD-11).

### 2. The status bar became four uniform sections

- **Four sections, one grammar:** `Models · Engines · Jobs · Notifications`. Each is a name plus a single circle — outlined when empty, filled when not. No counts anywhere, including accessible names. Geometry follows VS Code's actual metrics (22px bar, `padding: 0 8px`, items `0 5px` at 12px/400) rather than an approximation.
- **Engines is new, and it can stop things.** A live readout of every managed child process — template daemons, warm app workers, background app daemons — with a Stop per row. Backed by two new endpoints over `engine_host`'s existing in-memory children.
- **One panel open at a time, and the fold is never remembered.** Every section mounts collapsed on every load — a `.dl-panel` floats over the page and is dismissed by an outside click or Escape, so "open" is a statement about this moment, not a preference. The four `localStorage` keys and their helpers are deleted, not merely unread. An arrival can transiently open a section; a drained list closes it. Models and Engines never auto-open at all — running is not news.
- **Job rows use lines instead of truncation.** The old shrink ladder minced a model name to `F…` while the card had room to spare; the title now wraps and the model gets its own line.
- **Model memory leads with what is true right now.** A row reads `1.7 GB now (24 GB held)` — resident first, then the total the process is holding including the Metal pool, colour-banded against this machine's wired-memory ceiling. Both figures are labelled, because on Apple Silicon they differ by more than an order of magnitude and an unlabelled pair invites the reader to pick the wrong one. Where the OS counter is unavailable the row shows `1.7 GB now` with no parenthetical and no band, rather than a guess.
  - `held` is `max(phys_footprint, resident_size)`, read from a single `task_vm_info` call, and is documented as a **lower bound** rather than the truth. Neither counter alone is the total: RSS is blind to MLX's Metal buffers, and `phys_footprint` excludes clean file-backed pages that RSS counts. Measured on a plain interpreter here, `resident_size` is 19.2 MB against `phys_footprint`'s 9.3 MB — so a runner that mmaps its weights read-only (GGUF, torch with mmap) would otherwise render `8.2 GB now (1.1 GB held)` and paint a comfortable band off the smaller number. The parenthetical is omitted when the two are equal, compared as bytes rather than formatted strings.

## Visuals

The status bar's visibility rules are the subtle part — a section's panel is steered by a transient override that never touches the user's saved preference.

```mermaid
stateDiagram-v2
    [*] --> Closed: no stored preference
    Closed --> Open: user clicks chip
    Open --> Closed: click, outside pointer, Escape
    Closed --> Open: new item arrives
    Open --> Closed: list drains to empty
    Open --> Closed: another section opened
```

Arrival beats drain in the same tick, so a job landing as the last one finishes opens rather than flashing shut. `neverOpen` sections skip the arrival edge but keep the drain edge.

## Usage

Nothing to invoke — the notification is the feature. Open an app in a git repo that is behind `origin`'s default branch and the Notifications section carries the row.

New endpoints:

```
GET  /api/git-upstream            # repos known to be behind; never shells out
POST /api/git-upstream            # update | switch, X-Fused + root allowlist
GET  /api/engines/running         # every live managed child
POST /api/engines/{id}/stop       # X-Fused; recoverable for all three kinds
GET  /api/ai/runtime              # + footprint, basis, ceiling, OS footprint
```

Stopping any engine is safe: a template engine respawns on the next `ensure`, a warm app worker on the next call, and a background app daemon going down is the documented "quit this app now" action.

## Test Plan

- [x] `bun test` — 2874 pass. One pre-existing failure (`appCardMenu` → `appShot.ts:142`), reproduced on main's tree in isolation.
- [x] `pytest` full suite — 39 failing IDs against main's macOS baseline; the three branch-only IDs are established flakes on files this branch never touches.
- [x] `tsc --noEmit` and the frontend boundary check (415 files) clean.
- [x] Server-side: new engines-router tests (running snapshot, stop removes the child), `/api/ai/runtime` field propagation including the null-not-zero path, and `refresh_memory` propagating the OS footprint past load.
- [x] The fold-persistence removal is guarded by a test that asserts no source file in any of the four sections touches `localStorage` for the fold — checked with comments stripped, since these files legitimately explain the removal in prose. Verified by reverting: flipping the collapsed default to `false` fails it.
- [x] Browser-verified by measurement, not by eye: circle and label box centres coincide at 822.5; all four panels closed on a fresh load; bar height 22px.
- [x] Code review over the full diff, briefed to hunt leftover logic from the branch's own mid-flight reversals. 15 findings, all addressed:
  - **One real bug.** `useAutoExpandOnNew`'s drain-close was fed only one of each panel's two row sources, so a finishing download slammed the Jobs panel shut over queue rows still on screen (including a live turn's only cancel control), and dismissing the last repo row closed Notifications over failure rows. Fixed by splitting the hook's input into `ids` (announceable) and `alsoDrawn` (counts for occupancy, never announces) — a plain union would have made an arriving failure an arrival and auto-opened the panel, which D586 forbids. Auto-open behaviour is unchanged in all four sections. Row identities are prefixed per source because a scheduled run's job id and its queue row genuinely collide.
  - **One false premise**, since corrected: the `held` figure's docstring claimed RSS is a strict subset of the footprint. It is not (see above).
  - **Thirteen historical leftovers**: `SPEC.md` still specified the dismissal rule a review had removed as a bug, plus a collapsed chip with a percentage and chevron and a deleted `overallFraction`; `jobsSummary`/`NO_QUEUE` outlived the sibling helpers deleted with them; two doc blocks sat on the wrong function, leaving `clearableCount` undocumented while two files cited its doc; `neverClose` was a dead option; and a spread of comments still described three `localStorage` fold keys, a removed dot, chip counts, and a tie the arbiter can no longer reach.
- [ ] The non-Darwin fallback for the OS-footprint probe resolves to `None` on this machine and is untested against a real Linux/Windows host. It degrades to no figure rather than a wrong one.


iamsdas:
@/private/tmp/claude-501/-Users-iamsdas-Work-fused-render/4ee6c4c9-d289-4961-99e1-19a067c70671/scratchpad/notif-demo/pr-comment.md

PR #914, opened 2026-08-28
Image to 3D: one picture in, a mesh out, locally on Apple Silicon

iamsdas:
## Status: parked, not shippable — read this before reopening

This PR is being kept in draft. The integration (routes, worker, Playground UI, tests) is complete, but the underlying engine misses the product's own bar on both axes that matter — reliability and speed — even after fixing a real bug found along the way. Full detail below; the short version:

- The MLX shape engine (`hy3dshape`'s `pipeline_mlx.ShapePipeline`, via a fork pinned in `hunyuan3d_mlx/pyproject.toml`) had a genuine scheduler bug that made **every** render produce an empty mesh. Found, proven by direct arithmetic, fixed in our fork, pushed (`iamsdas/Hunyuan3D-2.1-mlx@7e62963`), and an upstream PR opened: [dgrauet/Hunyuan3D-2.1-mlx#8](https://github.com/dgrauet/Hunyuan3D-2.1-mlx/pull/8).
- **Even fixed, only 1 of 5 test inputs renders** — including 3 of the fork's own `example_images/` assets, not adversarial inputs.
- **Even the one success is too slow.** 16 steps at octree 128 took 187s (≈11.7s/step); the fork's own documented settings (50 steps, octree 256) extrapolate to roughly 9-10 minutes. The product bar is 2 minutes. There is no settings combination that is both fast enough and shown to be reliable.

Anyone reopening this should not have to rediscover any of the above — it's all recorded here and in `DECISIONS.md`.

## The scheduler bug (fixed)

`hy3dshape/pipeline_mlx.py`'s denoising loop always calls the scheduler with an **explicit ascending** sigma schedule (`sigmas = np.linspace(0, 1, num_inference_steps)`). The MLX port's scheduler substitute, `mlx_arsenal.diffusion.FlowMatchEulerDiscreteScheduler.set_timesteps`, appends a trailing `0.0` to that schedule — correct for *that class's own documented default* (a descending schedule, never actually used by this pipeline), but wrong for the ascending schedule this pipeline always passes. The reference PyTorch scheduler, given the identical explicit ascending sigmas, appends a trailing `1.0` instead.

Both schedulers compute `prev_sample = sample + (sigma_next - sigma) * model_output`. With the torch reference's trailing `1.0`, the *last* Euler step of every render is a deliberate no-op (`sigma_next - sigma == 0`). With the MLX port's trailing `0.0`, that same last step instead computes `sigma_next - sigma == -1.0` — the full predicted velocity gets **subtracted** from the sample, reversing most of the accumulated denoising trajectory in one step, at the very end of every render. This reproduced as an empty SDF volume (zero near-surface voxels, `np.concatenate` on an empty list) on every image, step count, and octree resolution tested before the fix.

Fixed in our fork by overwriting the scheduler's own last sigma after `set_timesteps` returns, matching the reference's trailing-`1.0` convention exactly — `mlx_arsenal` itself is untouched (it behaves exactly as its own docs describe; this pipeline was the caller feeding it a schedule shape its default was never written for).

## Reliability, after the fix: 1 of 5

Same settings throughout — `steps=16, guidance=7.5, octree_resolution=128, seed=42` (a cheap setting chosen specifically so an empty result would be unambiguous, not "too few steps") — scheduler fix applied:

| image | subject coverage | result | faces | SDF min/max (coarse level) | latent std (step 0 → last step) |
|---|---|---|---|---|---|
| mermaid (fork's own `tests/test_stage1_to_stage2.py` fixture) | 28.8% | **succeeded** | 112,914 | real sign crossing at both hierarchical levels | 0.935 → 0.975 (stayed healthy) |
| shaded sphere (synthetic convex blob) | 34.6% | failed | — | -0.9995 / -0.9971, mean -0.9985 (flat, one sign) | 0.935 → 0.197 (collapsed) |
| tiger (fork's own `example_images/ainimal_9.png`) | 31.9% | failed | — | -0.9995 / -0.9976, mean -0.9985 | 0.935 → 0.209 (collapsed) |
| `example_images/198.png` (fork's own) | 36.9% | failed | — | -0.9990 / -0.9976, mean -0.9983 | 0.936 → 0.223 (collapsed) |
| `example_images/052.png` (fork's own) | 37.1% | failed | — | -0.9995 / -0.9976, mean -0.9985 | 0.936 → 0.207 (collapsed) |

The real user photo that originally surfaced this whole investigation also failed, at both 16 and 30 steps, with the identical signature.

Subject coverage isn't the differentiator — the one success has the *lowest* coverage of the five. Every failure shares the same fingerprint: the SDF grid goes nearly constant around -0.998 (not just single-sign — flat to within ~0.002 across the whole query volume, which reads as a collapsed decode rather than a real-but-mispositioned object), and the DiT's own latent std collapses from ~0.935 at step 0 to ~0.2 at the final step, where the one success's std stayed at 0.94-0.98 throughout. Whether this is a residual scheduler/precision issue, a step-count-sensitivity question (unresolved — a `steps=30` retry on the real photo changed nothing measurably), or something else entirely was not run down further, because the timing numbers below make the answer moot for shipping.

## Timing vs. the product bar

- 16 steps, octree 128: 187s wall clock ≈ 11.7s/step.
- 50 steps, octree 256 (the fork's own documented/tested default): extrapolates to roughly 9-10 minutes.
- Best case found (reduced quality, and only in the 1-of-5 case that produces anything at all): ~3.1 minutes.
- **Product bar: 2 minutes.** Nothing tested clears it.

The torch reference pipeline was considered as a way to separate "port bug" from "wrong settings" but was not run: it is not installed anywhere on this machine (a fresh venv plus a separate multi-GB `tencent/Hunyuan3D-2.1` PyTorch checkpoint would be needed), and it would be *slower* than the already-too-slow MLX path — MLX being faster is this port's whole reason to exist, so clearing that bar was never going to change the shipping decision.

## Known, identified performance defects (not yet fixed — the path to 2-4x if anyone revisits)

Found by direct code reading and, where noted, verified by measurement — not fixed here, since perf tuning on a pipeline with a 1/5 hit rate wasn't worth doing until reliability was understood:

- **`mlx_arsenal.moe.moe.MoELayer.__call__` runs all 8 experts densely on every token instead of a routed top-2 gather** (config: `num_experts=8, moe_top_k=2`). Verified mathematically correct (matches an independent per-token gather-and-sum reference within float tolerance) — a real perf defect, not a correctness bug. ~4x more expert compute than the architecture needs.
- **The same `MoELayer.__call__` also forces 48 blocking `.item()` host syncs per DiT forward pass** (8 experts × 6 MoE layers), guarding a skip branch that essentially never fires at this token count — paying full sync cost for a skip that doesn't happen.
- **CFG doubles the DiT batch whenever `guidance_scale >= 0`**, including the algebraically-redundant `guidance_scale == 1` case (batch-of-2 gives an output identical to batch-of-1 at that exact value). Confirmed by direct code read of `pipeline_mlx.py`'s `do_cfg` branch; not separately measured in isolation.
- Measured at octree 128: VAE decode (`_query_sdf_volume`, both hierarchical levels combined) held at 2.3-3.3 seconds across every step count tried (4 through 50), while total wall time scaled 45-679 seconds with steps — **the DiT denoising loop dominates wall time, not decode chunking.** `num_chunks` tuning (the original perf hypothesis) was measured and found NOT to be the bottleneck; left at the fork's own default.

## What's actually built and working

- Routes, worker lifecycle, job progress (real per-phase denoise/decode ticks with mid-render cancellation), registry/catalog entries with a corrected (measured, not guessed) memory figure, and the Playground's Mesh stage UI are all complete and covered by tests — see the commit history on this branch.
- The image-preprocessing defect that first looked like the whole story (missing recentre-to-bbox, and a required-alpha-cutout-only design that was too strict) is genuinely fixed: `_prepare_cutout` ports the torch reference's recenter geometry and auto-mattes a flat, non-alpha backdrop by border colour. That fix is real and independently useful, but it was not sufficient — see the scheduler bug above, found only after the preprocessing fix didn't resolve the failure.

## Two things that still need a human (never machine-verified, and structurally can't be)

- A real end-to-end render through the native file picker, watched in the actual app.
- The 3D canvas's geometry and orbit-control pointer behaviour in `MeshStage`'s viewer.

## Original test plan (superseded by the above — kept for the record)

- [x] Automated, scoped tests — `tests/test_ai_supervisor_mesh.py`, `tests/test_ai_hunyuan3d_mlx_worker.py` (extended through this investigation), plus registry/catalog/runtime/tasks/formats/hub_models additions. Frontend: `aiModelGroups.test.ts`, `engines.test.ts`, `benchmark.test.ts`, `pick.test.ts`.
- [x] Runner venv builds and imports from scratch on Apple Silicon.
- [ ] Full pytest suite — not re-verified against this PR's final state; moot given the parking decision above.
- [ ] Real end-to-end render — see "what still needs a human" above.
- [ ] 3D canvas geometry and orbit-control pointer behaviour — see "what still needs a human" above.


iamsdas:
## CI reconciliation at `c74485fc`

**Everything passes except `test-python-windows`, and that red is inherited from `main`, not introduced here.**

| Check | Result |
|---|---|
| test-python 3.11 / 3.12 / 3.13 | pass |
| fused-engine | pass (flaked once on `test_apps_api.py::test_spawn_really_delivers_the_prompt_to_the_claude_process`, clean on re-run) |
| frontend, bundle-contents, linux-desktop, detect-changes | pass |
| test-python-windows | **fail — see below** |

### The Windows red is main's

Two tests fail, both from #915's attachment work:

- `tests/test_claude_shots.py::test_the_spawn_line_grows_a_read_rule_per_attachment_directory` — `AssertionError: only an absolute, existing, non-root directory survives`
- `tests/test_schedule_images.py::test_a_send_with_images_pre_allows_the_task_shots_dir` — a path built as `.../home/task-shots\a.png`, mixing `/` and `\`

**`main` at `3f35146d` fails both with byte-identical assertions** ([run 33172214172](https://github.com/fusedio/fused-render/actions/runs/33172214172)). They reproduced twice on this branch and once on main, so this is a genuine Windows path-separator bug in main rather than a flake — it arrived here through the merge that resolved this PR's conflicts. This branch touches neither file nor any Windows path handling.

### Local verification

The full pytest suite was run independently against a fresh `origin/main` clone and the failure-ID sets were diffed both ways. The persistent core (35 failed + 1 error on this machine) is identical in identity on both sides; the 2–4 extra reds per run differed between runs and each passed 3/3 in isolation — `-n auto` load flakes that reproduce on main too. **No regression from this branch.** The feature's own tests: 616 passed.

### Still needs a human — cannot be verified in CI

- **A real render.** No weights were ever downloaded, so the 7.4 GB pull and an actual image→mesh have never run. Needs an Apple Silicon machine.
- **The 3D canvas.** `MeshStage`'s viewer was built headlessly; layout and orbit-control pointer behaviour can't be checked that way.

`ShapePipeline.__call__`'s return shape *was* checked against the pinned fork sha (a single `trimesh.Trimesh`, not the list the PyTorch reference pipeline returns), so `worker.py`'s `mesh.export(out)` is correct as written.

PR #939, opened 2026-08-31
Get test-python-windows green, and fix the scan-cancel bug it uncovered

isaacbrodsky:
`test-python-windows` had been red on `main` on every recent commit, taking the `Test Status` rollup — and therefore **every PR in the repo** — with it. This gets it green.

Chasing it turned up one real product bug, and two pieces of test overhead worth removing while they were in view. **All 13 checks pass on `9965ba2`, `test-python-windows` included.**

| | |
|---|---|
| Windows failures | 3 → **0** |
| Suite wall clock | 291.9s → **267.1s** (measured, 2 runs a side) |
| Disk written per run | **~11GB less** |
| CI log | ~25,000 → ~2,900 lines |
| Product bugs fixed | **1** (`cancel_all_scans`) |

---

## 1. A live scan was often not cancelled — `fused_render/server/routers/index.py`

**The only production change here, and it is a real user-facing bug.** Found by CI going red, not by looking for it.

Turning indexing off is documented to cancel a scan already in flight, not merely refuse the next one. `cancel_all_scans` enumerated live runs through `_live_runs`, and its own docstring justified that by calling it a *"request-scoped cache."* **It is not.** It is a 1.0s process-wide cache, held so that ranking a keystroke does not re-fold every run directory.

So any ranked request in the preceding second answered the toggle-off from a listing taken **before the running scan existed**. The scan was not in it, `cancel` was never written, and the scan kept going. `/api/index/rank` fires on every keystroke in two search boxes, so in a running app that cache is essentially always warm — **this was the common case, not the rare one.**

| `_runs_cache` state when the toggle arrives | `cancel` sentinel written? |
|---|---|
| cold | ✅ yes |
| primed < 1s earlier | ❌ **no** |

Fixed by listing through `runner.list_runs` directly — the same call `_live_runs` wraps, so the entry shape is unchanged and the ranking path keeps its cache. Toggling indexing off is a rare deliberate act; it can afford the fold a keystroke cannot.

The existing test passed only by luck, which is why this survived. The new one primes the cache first, making it deterministic. **Mutation checked**: it fails on the old enumeration, passes on the new.

## 2. The three Windows-only `test_app_fused_dir` failures

Two fixture bugs, both in the tests — `app_fused_dir`, `claude_session_move` and `workspace_migration` were behaving as documented.

**The fake "old" root was not absolute on Windows.** `os.path.join(os.sep, "somewhere", ...)` is `/somewhere/...` on POSIX but `\somewhere\...` on Windows — drive-*relative*. `ntpath.isabs` answers True anyway, but with no drive `abspath` prepends the current one and **the string changes**. Every path function under test abspaths its input, so the fixture seeded transcripts containing `\somewhere\else\demo` while `_path_rewrites` searched for `D:\somewhere\else\demo`. Now built by an `_absent_root()` helper, with a guard test asserting the contract that actually matters — `abspath(p) == p`, **not** `isabs`.

**The bookmark URL was hand-encoded with POSIX assumptions.** `lstrip("/")` and `.split("/")` left the whole backslashed path as one segment, producing `/explorer/view/C%3A%5CUsers%5C…` — verbatim what CI printed. Now built with the shell's own `_view_url_codec.view_url_path`. On POSIX the two constructions are identical.

## 3. One test wrote 5GB of real bytes; two more wrote 3GB each

Three tests build a model file solely so something can read its length back. The largest was `b"x" * (5 * 1024 ** 3)` — 5GB of RAM to build the string, 5GB written to disk, for a file nothing ever reads.

`tests/_big_files.sparse_file` sets the length with `ftruncate` instead. `st_size` is identical, which is what `os.path.getsize` and `hub_cache`'s repo walk both sum, and it allocates nothing. It also matches what a real download leaves on disk — `worker_base.folder_bytes` documents that our segments write out of order, so a `.fusedpart` is created at its final size and filled sparsely.

| test | before | after |
|---|---|---|
| `test_a_local_cache_hit_is_preferred_over_a_remote_listing` | 67.66s | 0.00s |
| `test_a_cached_entrys_size_is_its_real_measured_footprint` | 35.55s | 0.07s |
| `test_a_second_revision_landing_in_an_EXISTING_repo_updates_its_size` | 14.05s | 0.08s |

That is ~117s of worker time, which on 4 cores under `-n auto` is ~25s of wall clock — the other workers absorb most of it. The **~11GB of disk** is the part that may matter more on `windows-latest`.

Deliberately **not** applied to `test_fetched_bytes_reads_the_sidecar_not_the_part_files_length`: download progress is measured by allocated blocks precisely because a preallocated `.fusedpart`'s length is its eventual size rather than what has arrived, so a sparse one would describe a download that had not started. `tests/_big_files`'s docstring records that as the one place the helper is wrong.

## 4. A test that raced a 1.0s deadline

`test_index_search::test_the_run_listing_is_not_re_read_on_every_keystroke` fires five ranked requests and asserts `list_runs` ran once — which holds only if all five HTTP round trips fit inside `RUNS_CACHE_S`. On a loaded runner they do not, and it reads `assert 2 == 1`. It failed on two separate Windows runs before being fixed here.

It now asserts the shipped constant is a real window — so a regression to zero still fails — and then widens it for the loop. *How many round trips fit in a second* is a fact about the machine, not about the cache.

## 5. `-v` was ~89% of the CI log — `.github/workflows/test.yml`

Both full-suite jobs ran `pytest tests/ -v`. Measured on a 513-test sample, `-v` costs **2.00 extra lines per test**; at 11,383 collected tests that is **~22,800 lines** around ~2,900 lines of everything else, which is past what the Actions UI renders in one piece. Every other pytest job in the file already used `-q`.

Nothing is lost: pytest's "short test summary info" names every failure by nodeid without `-v`, and under `-n auto` the per-test lines interleave across four workers, so they never read as a run order.

**Not fixed here:** the remaining ~1,750 log lines are the warnings summary, ~1,580 of which are `tests/x.py: N warnings` attribution for the **19 `@app.on_event` deprecations** in `server/app.py`. Migrating those to a lifespan handler is the fix, and it changes how the server starts — it belongs in its own PR rather than riding along here.

---

## Verification

- **CI:** all 13 checks green on `9965ba2`, including `test-python-windows` (9m59s) and Cursor Bugbot.
- **Full suite, before and after the production change:** byte-identical failure lists. The 32 local failures are all pre-existing — 24 are missing `rasterio`/`shapely` in my container; the other 8 reproduce on a clean `origin/main` checkout.
- **Collected node ids diffed** across the change: exactly one new test, none dropped.
- The Windows path reasoning was driven on Linux with Windows-shaped strings before pushing, rather than guessed: the old hand-rolled URL reproduces the CI failure byte for byte, and `_remap` returns `None` for the drive-relative root and the right path for the absolute one.

## Reviewers

Commits 1–4 are test-only. **`9965ba2` is the one production change** and deserves its own look rather than riding along with the test fixes — happy to split it into a separate PR if that reads better.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -201,9 +201,19 @@ jobs:
           sudo chmod +x /usr/local/bin/rclone
           rclone version
 
+      # `-q`, not `-v`: `-v` prints one line per test, and there are >11,000 of
+      # them, so it buried this job's actual output ~5:1 and pushed the log past
+      # what the Actions UI will render in one piece. It bought nothing that is
+      # not still here — pytest's own "short test summary info" names every
+      # failure by nodeid either way, and under `-n auto` the per-test lines
+      # arrive interleaved across workers, so they never read as a run order.
+      # The rest of the volume is the warnings summary, which is ~1,600 lines of
+      # `tests/x.py: N warnings` attribution for the 19 `@app.on_event`
+      # deprecations in server/app.py — fixed by migrating those to a lifespan
+      # handler, not by a pytest flag.
       - name: Run tests
         run: >
-          python -m pytest tests/ -v
+          python -m pytest tests/ -q
           --cov=fused_render --cov-report=term-missing --cov-report=xml --cov-report=html
 
       # One representative version avoids 4 duplicate reports/artifacts for
@@ -312,8 +322,10 @@ jobs:
           Add-Content -Path $env:GITHUB_PATH -Value $rcloneExe.DirectoryName
           & $rcloneExe.FullName version
 
+      # `-q` for the reason the Linux job above spells out: this job collects
+      # ~11,200 tests, and one `-v` line each is most of a 10k+ line log.
       - name: Run tests
-        run: python -m pytest tests/ -v
+        run: python -m pytest tests/ -q
 
   # The optional fused execution engine (D69) — the engine the migration plan
   # promotes to default, after which the built-in executor goes away. Until this
--- a/fused_render/server/routers/index.py
+++ b/fused_render/server/routers/index.py
@@ -786,16 +786,24 @@ def cancel_all_scans() -> list:
     The one caller is `shell/prefs.py`'s `PUT /api/prefs`, when the indexing
     toggle flips to off: the contract is that a scan running at the moment of
     toggle-off is cancelled outright, not merely refused going forward. Lives
-    here rather than in the index layer because it needs `_live_runs`'
-    request-scoped cache and the same live-run enumeration `_scan_in_flight`
-    uses, and the index package cannot import the server (this module
-    already owns that seam for scan/cancel/status).
+    here rather than in the index layer because the index package cannot import
+    the server (this module already owns that seam for scan/cancel/status).
+
+    Listed through `runner.list_runs` directly rather than through
+    `_live_runs`. That helper is NOT request-scoped, whatever an earlier
+    version of this docstring said -- it is a 1.0s process-wide cache, held so
+    that ranking a keystroke does not re-fold every run directory. Reading a
+    toggle-off through it meant any ranked request in the preceding second --
+    and `/api/index/rank` fires on every keystroke in two search boxes --
+    answered from a listing taken before the running scan existed, so the scan
+    this exists to stop was never seen and never cancelled. Toggling indexing
+    off is a rare deliberate act; it can afford the fold a keystroke cannot.
 
     Returns the run ids actually cancelled — a run already told to stop is
     skipped, same as `active_run`'s own liveness rule."""
     cfg = load_config()
     cancelled = []
-    for r in _live_runs(cfg):
+    for r in runner.list_runs(cfg, limit=KEEP_RUNS)["runs"]:
         if not r.get("running"):
             continue
         rid = r.get("run_id")
--- a/tests/_big_files.py
+++ b/tests/_big_files.py
@@ -1,0 +1,68 @@
+"""Files that REPORT a size without spending it on the disk.
+
+A handful of tests here are about a number the app derives from how big a
+model file is — a budget-aware download picking a smaller quantisation, a
+cached repo's `size_gb`. What those tests need is a file whose `st_size` is
+multi-gigabyte; the bytes inside it are never read. Writing them for real
+costs both wall clock and disk (one 5GB `write_bytes` was 67s of the suite's
+295s, and repeated local runs filled the disk), so set the length instead of
+writing data: `ftruncate` is constant time and allocates nothing.
+
+That is also what a real download looks like on this disk. `worker_base`'s
+own docstring for `folder_bytes` spells it out: our segments write out of
+order, so a `.fusedpart` is created at its final size with `ftruncate` and
+filled sparsely.
+
+THE ONE PLACE THIS IS WRONG is a partial file — `.fusedpart` or hf's
+`.incomplete`. Download PROGRESS is deliberately measured by allocated
+BLOCKS rather than by length, for exactly the reason above: a sparse file's
+length is its eventual size, not what has arrived. A sparse `.fusedpart`
+would report zero bytes downloaded, so those fixtures have to write real
+bytes — which is fine, they are sized in megabytes rather than gigabytes.
+"""
+
+import ctypes
+import os
+import sys
+
+#: FSCTL_SET_SPARSE — see `_mark_sparse`.
+_FSCTL_SET_SPARSE = 0x000900C4
+
+
+def _mark_sparse(fd: int) -> None:
+    """Ask NTFS not to reserve the clusters. Best effort, never fatal.
+
+    A POSIX filesystem makes this sparse implicitly: extending the length
+    without writing allocates nothing at all. NTFS does not — it sets EOF
+    and reserves the space, so a 5GB fixture would still cost 5GB of a
+    Windows runner's rather small disk. It stays fast there either way (no
+    data is written), which is why failing to set the flag is not an error:
+    the worst case is the disk usage we already had before this helper.
+    """
+    if sys.platform != "win32":
+        return
+    try:
+        import msvcrt
+
+        returned = ctypes.c_ulong(0)
+        ctypes.windll.kernel32.DeviceIoControl(
+            ctypes.c_void_p(msvcrt.get_osfhandle(fd)), _FSCTL_SET_SPARSE,
+            None, 0, None, 0, ctypes.byref(returned), None)
+    except Exception:
+        pass
+
+
+def sparse_file(path, size: int) -> None:
+    """Create `path` as a `size`-byte file holding no data.
+
+    `os.stat(path).st_size == size`, so every reader that sums or compares
+    lengths — `os.path.getsize`, `hub_cache`'s repo walk — sees `size`. Reads
+    return NUL bytes. Read the module docstring before reaching for this in a
+    partial-download fixture.
+    """
+    parent = os.path.dirname(os.fspath(path))
+    if parent:
+        os.makedirs(parent, exist_ok=True)
+    with open(path, "wb") as f:
+        _mark_sparse(f.fileno())
+        f.truncate(size)
--- a/tests/test_ai_llamacpp_worker.py
+++ b/tests/test_ai_llamacpp_worker.py
@@ -25,6 +25,8 @@ from pathlib import Path
 
 import pytest
 
+from _big_files import sparse_file
+
 WORKER_PATH = str(
     Path(__file__).resolve().parents[1]
     / "fused_render" / "ai" / "runners" / "llama_text.py"
@@ -338,7 +340,9 @@ def test_a_local_cache_hit_is_preferred_over_a_remote_listing(
     snapshot = folder / "snapshots" / "abc123"
     snapshot.mkdir(parents=True)
     curated_file = snapshot / "gemma-4-E4B-it-Q4_K_M.gguf"
-    curated_file.write_bytes(b"x" * (5 * 1024 ** 3))
+    # Sparse: only its `st_size` is ever read (`os.path.getsize`, below),
+    # and 5GB of real bytes was 67s of the suite. See `_big_files`.
+    sparse_file(curated_file, 5 * 1024 ** 3)
     local_file = snapshot / "gemma-4-E4B-it-Q2_K.gguf"
     local_file.write_bytes(b"x" * 1024)
     worker.worker_base.repo_folder = lambda model_id, repo_type="model": (
--- a/tests/test_ai_runtime.py
+++ b/tests/test_ai_runtime.py
@@ -37,6 +37,7 @@ from fused_render.server.routers import ai_runtime
 # "every test stubs the Hub" is not the same claim as "no test reaches the
 # network".
 from test_ai_hub_fetch import no_egress  # noqa: F401
+from _big_files import sparse_file
 
 # os.geteuid is POSIX-only; a bare call below would crash collection of this
 # whole module on Windows, before any skipif could act on it.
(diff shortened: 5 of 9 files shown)


isaacbrodsky:
**The fix works — all three target failures are gone from Windows.** `tests/test_app_fused_dir.py` does not appear in the failure list on `0da9ed2` at all, where it had three failures on `main` and one on the first revision here.

Progression, same job:

| head | job time | failures |
|---|---|---|
| `main` @ `2c0073b` | 9:18 | **3** × `test_app_fused_dir` |
| `12cd106` (rev 1) | 8:47 | **1** × `test_app_fused_dir` |
| `0da9ed2` (rev 2) | **14:03** | **0** × `test_app_fused_dir` — 2 unrelated |

The counts reconcile exactly: 10591 passed / 1 failed → 10590 passed / 2 failed, same 11227 total. The one `test_app_fused_dir` failure became a pass, and two different tests failed.

## The two new failures are not this PR's

```
FAILED tests/test_community_marketplace_fixes.py::test_refresh_full_clones_into_workspace_showcase
  subprocess.TimeoutExpired: Command '['git', …, 'push', '-q', '-u', 'origin', 'HEAD']' timed out after 30 seconds
FAILED tests/test_calls.py::test_viewing_the_log_is_recorded_like_any_other_call
  AssertionError: the read of the log should itself be recorded
  assert 1 == 2  +  where 1 = len(['seed'])
```

Neither file is touched by this diff — it changes `tests/test_app_fused_dir.py` and nothing else. Both are timing-shaped, and the run they appeared on took **14:03 against 8:47 for the previous head of the same job**, a 60% slowdown with no change in test count. A `git push` to a local remote exceeding a 30-second timeout is resource starvation, not logic; a call-log assertion finding one recorded call where it wants two is an ordering race under the same pressure. Neither appeared on `main`'s runs or on rev 1.

**Re-running once to confirm**, per the usual rule that "flake" is not a root cause. I tried immediately and got `403 This workflow is already running` — the other jobs in run `33438431320` are still in flight — so I will re-run the failed jobs as soon as the workflow completes, and report the result here either way. If they reproduce, they are real and I will dig in rather than hand them off.

## Side observation on the flakiness

While looking into why this suite is slow I measured something relevant: the 68,936 warnings are not a cost (pytest aggregates them; disabling capture entirely is within noise), but the suite has **216 subprocess-spawn sites across 90 test files** and **233 `time.sleep` call sites**. Combined with a 30-second hard timeout on a `git push`, that is a suite whose pass/fail depends on runner load — which is consistent with the same job varying from 8:47 to 14:03. Not this PR's business to fix, but it is probably why Windows in particular has been an unreliable signal.

---

isaacbrodsky:
Re-run result (the one I promised): **neither of the two failures reproduced — a third, different test failed instead.**

```
FAILED tests/test_index_search.py::test_the_run_listing_is_not_re_read_on_every_keystroke
  assert 2 == 1   +  where 2 = len([1, 1])
== 1 failed, 10591 passed, 636 skipped in 631.82s (0:10:31) ===
```

Three Windows runs, three **disjoint** failure sets, all on essentially the same code:

| run | duration | failures |
|---|---|---|
| `0da9ed2` first | 14:03 | `test_community_marketplace_fixes` (git push `TimeoutExpired` after 30s), `test_calls` (`assert 1 == 2`) |
| `0da9ed2` re-run | 10:31 | `test_index_search` (`assert 2 == 1`) |
| `12cd106` | 8:47 | `test_app_fused_dir` (fixed by this PR's second commit) |

Zero overlap between the two `0da9ed2` runs, and `test_app_fused_dir` — the file this PR exists to fix — appears in **none** of them.

## What that establishes

Not "two specific broken tests", but a **pool of load-sensitive tests** any of which can fail on a slow run. Every failure seen so far is of one shape — an exact-count or a deadline assertion:

- `test_calls`: `assert 1 == 2` — a call not yet recorded
- `test_index_search`: `assert 2 == 1` — a listing re-read that should have been cached once
- `test_community_marketplace_fixes`: a local `git push` exceeding a 30-second timeout
- (`test_apps_api` on Linux 3.13 earlier in this PR: a `session` file read back empty after `done`)

And the same job varies **8:47 → 14:03 → 10:31** with no change in test count, so which ones fire is a function of runner load.

I have measurements on where that load comes from, from profiling the suite on Linux (`--durations=25`, 11,055 tests, 295s wall):

- `test_ai_llamacpp_worker.py:341` does `write_bytes(b"x" * (5 * 1024 ** 3))` — allocates 5 GiB of RAM and writes a real 5 GiB file so a size check has something to measure. At **67.66s it is 23% of the suite's wall clock**, and under `-n auto` a single test is a floor no parallelism can lower. `truncate` gives an identical `st_size` in 0.00s with 0 bytes on disk (measured: 8.25s → 0.00s per GiB). `test_ai_models_api.py:2250` writes another 96 MiB the same way.
- The suite has **216 subprocess-spawn sites across 90 files** and **233 `time.sleep` call sites**.

That is a plausible contributor to the variance, and worth its own PR — but it would not make exact-count assertions deterministic, so I am not claiming it as the fix.

## Where this PR stands

Its stated purpose is met and the evidence is in the runs above: 3 → 1 → 0 `test_app_fused_dir` failures, with the pass/fail counts reconciling exactly each time. Nothing in the remaining failures touches any file in this diff.

The one re-run is spent, and re-running again would just draw a different name from the same pool — so I am not going to keep spinning CI. What is left is a judgement call for a human: this is the same situation #937 was in when it was merged past a red rollup, and `main` itself cannot produce a green `Test Status` today. I will keep watching the PR and will act on any review comment or state change.

---

isaacbrodsky:
## Windows CI: all three target failures are gone

`test-python-windows` on `0da9ed2` ([job](https://github.com/fusedio/fused-render/actions/runs/33438431320/job/99649898651)) — `1 failed, 10591 passed, 636 skipped in 631.82s`. The three `test_app_fused_dir` failures this PR exists to fix no longer appear:

```
test_a_moved_app_carries_its_claude_sessions_and_repoints_meta   ✓
test_a_move_re_roots_bookmarks_and_app_stores                    ✓
test_a_second_move_before_the_live_session_ends_loses_nothing    ✓
```

## The one remaining failure is not this diff's

```
FAILED tests/test_index_search.py::test_the_run_listing_is_not_re_read_on_every_keystroke - assert 2 == 1
 +  where 2 = len([1, 1])
```

This PR changes one file, `tests/test_app_fused_dir.py`. It cannot reach `test_index_search`. **Not a flake either** — this is the second run it has appeared on, and it is a wall-clock deadline the test never meant to impose.

`_live_runs` (`fused_render/server/routers/index.py:301`) caches the folded run listing for `RUNS_CACHE_S = 1.0`, comparing `time.monotonic()` against that window on every ranked request. The test fires five ranked requests and asserts `list_runs` ran once — which holds only if all five HTTP round trips finish inside one second. On a loaded runner they do not, the entry expires mid-loop, and a second listing is taken. `len([1, 1])` is exactly two calls: one boundary crossed, not a cache that stopped working.

Driving `_live_runs` directly on a real clock reproduces the CI message byte for byte:

| five calls | `list_runs` calls | test's assertion |
|---|---|---|
| all inside the window | `[1]` | passes |
| one 1.05s gap before the last | `[1, 1]` | `assert 2 == 1` |

**Fixed on a separate branch rather than here**, so this PR stays the one-file Windows path change it is reviewed as: [`1dd034a`](https://github.com/fusedio/fused-render/commit/1dd034a) on `claude/sine-generator-permission-error-2e8ozs`. The test now asserts the shipped `RUNS_CACHE_S` is a real window (so a regression to zero still fails) and then widens it for the loop, because *how many round trips fit in a second* is a fact about the machine, not about the cache. `test_the_cached_run_listing_expires` still owns expiry, and already forces the clock rather than waiting on it for the same reason.

Freezing `time.monotonic` for the loop was the first thing tried and is not viable — patching it module-wide stalls the TestClient's event loop, which reads the same clock for its own timeouts.

---

isaacbrodsky:
## Two more test-only commits, folded in

Following the comment above: rather than leave the keystroke-cache race for a separate branch, both changes now live here. Still **no production file is touched** — six test files, `0` under `fused_render/`.

### [`5d7c01d`](https://github.com/fusedio/fused-render/commit/5d7c01d) — the remaining `test-python-windows` failure

`test_index_search::test_the_run_listing_is_not_re_read_on_every_keystroke`, diagnosed in the previous comment: a 1.0s `RUNS_CACHE_S` window that five HTTP round trips have to fit inside, which they do not on a loaded runner. The test now asserts the shipped constant is a real window — so a regression to zero still fails — and then widens it for the loop, because *how many round trips fit in a second* is a fact about the machine rather than about the cache.

### [`a494127`](https://github.com/fusedio/fused-render/commit/a494127) — three fixtures that were writing real gigabytes

Three tests build a model file purely so something can read its length back — a budget-aware resolve picking a smaller quantisation, a cached repo's `size_gb`. The largest was `b"x" * (5 * 1024 ** 3)`: 5GB of RAM to build the string, 5GB written to disk, for a file nothing ever reads. `tests/_big_files.sparse_file` sets the length with `ftruncate` instead — identical `st_size`, which is what `os.path.getsize` and `hub_cache`'s repo walk both sum, and nothing allocated.

| test | before | after |
|---|---|---|
| `test_a_local_cache_hit_is_preferred_over_a_remote_listing` | 67.66s | 0.00s |
| `test_a_cached_entrys_size_is_its_real_measured_footprint` | 35.55s | 0.07s |
| `test_a_second_revision_landing_in_an_EXISTING_repo_updates_its_size` | 14.05s | 0.08s |

Measured on the full suite, two runs a side, identical collection (same 32 environment failures both sides — missing `rasterio`/`shapely` locally):

| | run 1 | run 2 | mean |
|---|---|---|---|
| `main` @ `d6bc270` | 298.4s | 285.3s | **291.9s** |
| with this commit | 258.2s | 276.0s | **267.1s** |

~25s of wall clock, not the ~117s the per-test numbers add up to — `-n auto` means three other workers absorb most of it. The disk saving is the part that may matter more here: **~11GB less written per run**, and `windows-latest` has the least headroom of the three runners.

Deliberately **not** applied to `test_fetched_bytes_reads_the_sidecar_not_the_part_files_length`, the one large fixture left. Download progress is measured by allocated blocks precisely because a preallocated `.fusedpart`'s length is its eventual size rather than what has arrived, so a sparse one would describe a download that had not started. It is 96MB. `tests/_big_files`'s module docstring records that as the one place this helper is wrong.

`sparse_file` also asks NTFS for a sparse file via `FSCTL_SET_SPARSE`, best effort — `truncate` on Windows is fast but reserves the clusters. That branch cannot run on Linux, so it is guarded and non-fatal: if it fails, the cost is the disk usage we already had.

Locally: the six files this PR touches are **659 passed, 8 skipped**. Windows CI is the verdict for the two fixes aimed at it.

---

isaacbrodsky:
## ⚠️ This PR now contains a production change — a real bug, found by CI going red

`test-python (3.13)` failed on `24c0677`:

```
FAILED tests/test_shell_prefs.py::test_turning_indexing_off_cancels_a_live_scan - AssertionError: assert False
 +  where False = exists('.../runs/20260831-225712-aad5fc/cancel')
```

I chased it expecting a fourth flaky test. It is not — it is a genuine defect in `fused_render/server/routers/index.py`, and the test was passing by luck.

### What is broken

Turning indexing off is documented to cancel a scan already in flight, not merely refuse the next one. `cancel_all_scans` enumerated live runs through `_live_runs`, and its own docstring justified that by calling it a *"request-scoped cache."* **It is not.** It is a 1.0s process-wide cache, held so that ranking a keystroke does not re-fold every run directory.

So any ranked request in the preceding second answered the toggle-off from a listing taken **before the running scan existed**. The scan was not in it, `cancel` was never written, and the scan kept running.

`/api/index/rank` fires on every keystroke in two search boxes, so in a running app that cache is essentially always warm. **This was the common case, not the rare one.**

### Verified, not inferred

Priming the cache the way any earlier ranked request would makes it deterministic:

| `_runs_cache` state when the toggle arrives | `cancel` sentinel written? |
|---|---|
| cold | ✅ yes |
| primed < 1s earlier | ❌ **no** — the failing assertion |

The existing test only ever passed because nothing happened to prime the cache in the second before it ran. Under `-n auto` that is a coin flip on what the worker did previously — hence intermittent, and hence it finally came up tails here.

### The fix — [`9965ba2`](https://github.com/fusedio/fused-render/commit/9965ba2)

List through `runner.list_runs` directly: the same call `_live_runs` wraps, so the entry shape is unchanged, and the ranking path keeps its cache untouched. Toggling indexing off is a rare deliberate act and can afford the fold a keystroke cannot.

The new test primes the cache first, turning "depends on which test ran a moment earlier in this worker" into a deterministic assertion. **Mutation checked**: it fails on the old enumeration, passes on the new one. Full suite before and after the change: **byte-identical failure lists** (32, all pre-existing — 24 are missing `rasterio`/`shapely` in my container, 8 reproduce on a clean `origin/main` checkout).

### Reviewers: this changes the PR's shape

Every earlier commit here was test-only. This one touches server code, so it deserves its own look rather than riding along with the test fixes. Happy to split it into a separate PR if that reads better — say the word and I will move it.

---

PR #940, opened 2026-08-31
Add heartbeat to remote Claude calls to prevent stale reporting

isaacbrodsky:
## Summary
Fixes a bug where Claude API calls longer than 30 seconds were incorrectly reported as "No longer reporting" in the activity row, even though they were running successfully. The fix adds a periodic heartbeat that keeps the row's `updated_at` timestamp fresh while the call is in flight.

## Key Changes

- **Added heartbeat mechanism to `_ai_relay`**: Introduced `_start_remote_beat()` and `_stop_remote_beat()` functions that periodically update the job row's timestamp without modifying any fields, keeping it from appearing stale.

- **New module constants**:
  - `_REMOTE_TICK_S`: Heartbeat interval (10 seconds), well under the 30-second stale window
  - `_REMOTE_ROW_DETAIL`: Standardized detail text for active remote calls ("Claude — remote")
  - `_REMOTE_QUEUED_DETAIL`: Detail text for calls waiting behind another call in the queue

- **Queue state awareness**: Calls now detect when they're waiting for `_AI_SESSION.lock` and display `_REMOTE_QUEUED_DETAIL` instead of appearing to be actively running, then switch back to `_REMOTE_ROW_DETAIL` once execution begins.

- **Heartbeat lifecycle management**: The beat task is properly cancelled and awaited in `finally` blocks for both streaming and non-streaming paths to prevent task leaks.

- **Test infrastructure**: Added `_tight_timers()` helper and `_watch_row()` async sampler to test row state during call execution (not just after completion), plus four new tests covering:
  - Slow calls don't report as stalled
  - Slow streaming calls don't report as stalled
  - Heartbeat stops when call completes (no task leaks)
  - Queued calls display correct state and don't report as stalled

- **Fake process enhancement**: Updated `_FakeProc` test helper to support `turn_delay` parameter for simulating slow model responses.

## Implementation Details

The heartbeat uses a fieldless `_report_remote()` call, which leverages `jobs.upsert`'s behavior of only applying keys that are present. This means a heartbeat tick updates only the `updated_at` timestamp and cannot accidentally reopen dismissed rows or overwrite the detail line set by `run_once` during queue transitions.

The heartbeat is started after `_open_remote_job()` and stopped before the row is closed, ensuring the row never appears stale during execution while preventing ticks from landing after the row is dismissed.

https://claude.ai/code/session_012WsZQf6mqc8buXUhAFZTsq

---

> [!NOTE]
> **Low Risk**
> Changes are limited to activity-row reporting around the existing Claude relay; inference and API contracts are unchanged aside from more accurate UI state.
> 
> **Overview**
> Fixes **false "No longer reporting"** on remote `fused.ai()` / `/api/ai` calls that run longer than the jobs registry’s 30s stale window (while timeouts allow up to 10 minutes).
> 
> The relay now runs a **10s display heartbeat** on both streaming and non-streaming paths: fieldless `jobs.upsert` ticks refresh `updated_at` without changing title, detail, or progress. Heartbeat tasks are **stopped in `finally`** so they don’t leak or tick after the row closes. **`_report_remote`** now swallows all exceptions so a registry glitch can’t kill the beat loop and recreate the stall.
> 
> **Queued calls** behind `_AI_SESSION.lock` briefly show **"Queued — another Claude call is in flight"** instead of looking actively busy, then revert to **"Claude — remote"** when the turn starts.
> 
> Docs (`SPEC.md`, jobs skill) describe the same pattern for page-owned jobs. Tests add slow-turn fakes and assert in-flight rows never go `stalled`, streaming parity, beat shutdown, resilient reporting, and queue detail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 692c297c89877a3aac7883ac1ab2b0cdd52a280f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


isaacbrodsky:
### `test-python-windows` is red, and it is not this PR's

`test-python-windows` failed on `e7f6776` with 4 tests. Three of them are red on this PR's own base commit (`d6bc270`, main) too — [run 33435076780](https://github.com/fusedio/fused-render/actions/runs/33435076780/job/99629850817), same three nodeids, same assertions:

```
tests/test_app_fused_dir.py::test_a_moved_app_carries_its_claude_sessions_and_repoints_meta
tests/test_app_fused_dir.py::test_a_move_re_roots_bookmarks_and_app_stores
tests/test_app_fused_dir.py::test_a_second_move_before_the_live_session_ends_loses_nothing
```

Base: `4 failed, 10587 passed`. Here: `4 failed, 10591 passed` — the four extra passes are this PR's new tests, which passed on Windows.

**Ported the existing fix.** #939 fixes exactly these three (drive-relative `\somewhere\else\demo` never matching what production abspaths, and a hand-rolled POSIX URL encoder). Its two `tests/test_app_fused_dir.py` commits are cherry-picked here as `ecef6e8` and `daba96d` — that file is disjoint from this PR's diff, and this port no-ops once main carries #939. I deliberately left #939's other four commits alone: the sparse-file fixtures, the `-v`→`-q` CI change and the `index` scan-cancel commit are not what is failing here, and carrying them would widen this PR.

**The fourth failure rotates and is a runner flake.** Base ran red on `test_ai_runtime.py::test_loading_the_same_model_twice_joins_rather_than_restarting` (readiness timeout); this run on `test_canvases.py::test_opening_a_canvas_retires_the_legacy_fused_plugin` (`PermissionError` on a temp `settings.json` under an xdist worker). Different test each time, neither in code this diff touches. The push above re-runs the job, which is the one re-run I'd have spent confirming it.

Locally on the pushed head: `tests/test_app_fused_dir.py` + `tests/test_server_ai.py` → 106 passed, 1 skipped.

---

PR #942, opened 2026-09-01
Move the 19 on_event hooks to a lifespan handler (73,154 warnings → 4)

isaacbrodsky:
FastAPI deprecated `@app.on_event`, and `create_app` had **19** of them — 12 startup, 7 shutdown. Each warns on every app built, and pytest attributes every warning to every test file that triggered it.

This is the other half of the CI-log problem #939 started on: that PR removed `-v` (~22,800 lines); this removes the warnings summary.

| | before | after |
|---|---|---|
| warnings recorded | **73,154** | **4** |
| warnings-summary lines | 1,749 | **9** |
| full `-q` log | 2,883 lines | **1,160** |

Combined with #939's `-v` removal, the Windows CI log goes from **~25,000 lines to ~1,200**.

The four survivors are raised by the vendored `fused` package's own `_udf/decorators.py`, not by anything in this repo.

## What actually changed

**The handlers are untouched** — each keeps its body, its comment, and its position beside the routes it belongs with. Only the decorator changes, to one of two collectors:

```python
@on_startup                 # was: @app.on_event("startup")
async def _startup_prewarm_ai():
    prewarm_ai()
```

`_lifespan` reads those lists when the app **starts**, not when it is built. That is what lets a handler 450 lines below `FastAPI(...)` register with a lifespan that had to be passed at construction — and it is asserted, since it is the non-obvious part.

It is a module-level factory rather than a closure so its contract can be driven with fake handlers instead of the real 19, which start schedulers and warm engines.

## The two details worth reviewing

Both are taken from what `on_event` **did**, not what it looks like it did — `fastapi/routing.py`, `_startup`/`_shutdown` and `_DefaultLifespan`. (Worth noting: Starlette 1.6 removed `on_startup`/`on_shutdown` outright; FastAPI now carries its own copy, so that file is the only authority.)

**1. Shutdown runs in registration order, not reversed.**

```python
async def _shutdown(self) -> None:
    for handler in self.on_shutdown:      # forwards
```

Teardown conventionally unwinds in reverse, and this one never did. Reversing it here would change the order 7 real handlers close things in, while reading as a cleanup — so there is a test named for exactly that, with the reason attached.

**2. Shutdown runs even when the served application raised.**

`__aexit__` ran regardless once `__aenter__` had returned. A bare `yield` with no `finally` would quietly stop doing that, and nothing else in the suite would notice — the server would simply stop cleaning up after a crash. A startup that *raises* still skips shutdown entirely, because `__aenter__` never returned, which is why the `try` opens **after** the startup loop rather than around it.

## Verification

`tests/test_app_lifespan.py`, 6 tests, **mutation checked** — each failure mode is caught by exactly one test, so none of them is decorative:

| mutation | fails |
|---|---|
| `reversed(shutdown_handlers)` | `..._NOT_reversed` |
| drop the `try`/`finally` | `..._still_runs_when_the_application_raised` |
| reinstate one `@app.on_event` | `..._registers_nothing_on_the_deprecated_path` |

Full suite: **byte-identical failure list** to `main` (32, all pre-existing — 24 are missing `rasterio`/`shapely` in my container, 8 reproduce on a clean checkout). Wall clock 249s vs 267s, which is within run-to-run noise; the warnings were never the slow part.

Driven through `asyncio.run` rather than async test functions: nothing else in this suite needs `pytest-asyncio`, and one file is a poor reason to add a plugin every run then loads.

https://claude.ai/code/session_01HXDxeKbDsWkANaiDrzre3Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01HXDxeKbDsWkANaiDrzre3Y)_

---

> [!NOTE]
> **Medium Risk**
> Touches process-wide startup/shutdown ordering for pooled HTTP, AI workers, engines, captures, and background threads; behavior is intended to be identical but mistakes in lifespan semantics would surface as subtle teardown or ordering bugs.
> 
> **Overview**
> Replaces deprecated `@app.on_event` with a custom **`lifespan`** on `create_app`, cutting CI deprecation noise (on the order of **73k warnings → 4** per the PR description) without changing what each hook does.
> 
> Handlers keep the same bodies and placement; only decorators switch to **`@on_startup`** / **`@on_shutdown`**, which append to lists read when the app **starts** (not at `FastAPI(...)` construction). **`_lifespan`** mirrors old FastAPI behavior: startup and shutdown run in **registration order** (shutdown is **not** reversed), shutdown runs in a **`finally`** after a successful startup even if the app raises, and a failing startup skips shutdown.
> 
> **`app.state.startup_handlers`** / **`shutdown_handlers`** expose the registry for tests. New **`tests/test_app_lifespan.py`** locks the contract and asserts all 12 startup + 7 shutdown hooks match the pre-migration order, with guards against reintroducing **`on_event`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd79b232ee2d0d96234d6583674a193254abc0f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/fused_render/server/app.py
+++ b/fused_render/server/app.py
@@ -376,6 +376,13 @@ def create_app(start_dir: str) -> FastAPI:
 
     app = FastAPI(title="fused-render",
                   lifespan=_lifespan(startup_handlers, shutdown_handlers))
+    # Exposed so the registry is inspectable — `on_event` kept its own on
+    # `app.router.on_startup`/`on_shutdown`, and losing that would mean the
+    # ORDER these run in, which is the contract, could only be checked by
+    # reading 450 lines of decorators. `tests/test_app_lifespan.py` asserts
+    # the exact sequence against what `on_event` produced.
+    app.state.startup_handlers = startup_handlers
+    app.state.shutdown_handlers = shutdown_handlers
     app.state.start_dir = start_dir
 
     # Shared keep-alive HTTP pool for the opt-in pooled /api/fs/raw proxy
--- a/fused_render/server/app.py
+++ b/fused_render/server/app.py
@@ -20,6 +20,7 @@ import json
 import os
 import threading
 import time
+from contextlib import asynccontextmanager
 
 from fastapi import FastAPI
 from fastapi.staticfiles import StaticFiles
@@ -302,6 +303,42 @@ def _export_bundled_uv_path() -> None:
     os.environ["PATH"] = (uv_dir + os.pathsep + path) if path else uv_dir
 
 
+def _lifespan(startup_handlers: list, shutdown_handlers: list):
+    """The startup/shutdown contract `on_event` used to provide.
+
+    Lifted out of `create_app` so it can be driven with fake handlers rather
+    than with the 19 real ones, which start schedulers and warm engines. The
+    lists are read when the app STARTS, not when this is called, so a handler
+    registered further down `create_app` is still picked up.
+
+    Two details are the whole reason this is not a one-liner, both taken from
+    what `on_event` actually did (fastapi/routing.py, `_startup`/`_shutdown`
+    and `_DefaultLifespan`):
+
+    * Both lists run in REGISTRATION order. Shutdown is **not** reversed — it
+      reads like it ought to be, and it never was, so reversing it here would
+      be a silent behaviour change wearing the clothes of a cleanup.
+    * Shutdown runs even when the served application raised, because
+      `__aexit__` ran regardless once `__aenter__` had returned. A bare
+      `yield` with no `finally` would quietly stop doing that. A startup that
+      raises still skips shutdown entirely, because `__aenter__` never
+      returned — which is why the `try` opens after the startup loop rather
+      than around it.
+    """
+
+    @asynccontextmanager
+    async def lifespan(app: FastAPI):
+        for handler in startup_handlers:
+            await handler()
+        try:
+            yield
+        finally:
+            for handler in shutdown_handlers:
+                await handler()
+
+    return lifespan
+
+
 def create_app(start_dir: str) -> FastAPI:
     # Engine (D69/D70 + SPEC §20): validate any FUSED_RENDER_ENGINE override
     # ONCE at startup — this raises on a bad value and fails loudly for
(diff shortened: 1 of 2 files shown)


isaacbrodsky:
## Follow-up: the original tests proved the mechanism, not the wiring

Pushed [`dd79b23`](https://github.com/fusedio/fused-render/commit/dd79b23) after a fair challenge to the verification.

The tests I opened this PR with drive `_lifespan` with **fake** handlers. That pins the contract — order, `finally`, a startup that raises — but says nothing about whether the real 19 hooks reach it. **A decorator pointed at the wrong collector would have passed every one of them**, and most of the suite builds apps without running lifespan at all, so it would not have been caught there either.

### Now asserted against ground truth

`EXPECTED_STARTUP` / `EXPECTED_SHUTDOWN` were read off `app.router.on_startup` and `.on_shutdown` **on the commit before the migration** — not transcribed from the decorators. Both match exactly, 12 and 7.

Checking that needed the registries reachable, so `create_app` puts them on `app.state`. That restores something the move had quietly dropped: `on_event` kept its own registry on `app.router`, and without one the order these run in — the contract — could only be verified by reading 450 lines of decorators.

Two further mutations now fail, each on exactly the new test:

| mutation | |
|---|---|
| `@on_startup` → `@on_shutdown` on one handler | ✅ caught |
| swap two adjacent registrations | ✅ caught |

### The app was also actually started

Separately, not as a committed test: the real lifespan was entered and left on **both** revisions, with every handler wrapped to record what fired.

```
main   ran 19 handlers, /api/config → 200
branch ran 19 handlers, /api/config → 200
IDENTICAL SEQUENCE: True
```

Nothing had started the app until then — worth saying plainly, since the full-suite pass reads like stronger evidence than it was.

### One transient, chased down rather than waved off

A full run turned up `test_apps_api.py::test_poll_serves_a_run_started_by_the_server`, which had not failed in 7 earlier full runs (two of them on `main`). It passes 10/10 in isolation, and it is a 30-second deadline around a real subprocess spawn — the same load-sensitive shape as the tests fixed in #939. A confirming re-run came back **32 failures, byte-identical to the pre-existing set**. Not this PR's, and not suppressed — recorded here in case it resurfaces.

---

isaacbrodsky:
## `test-python-windows` on `7cd35d7` — neither failure is this PR's

Recording the diagnosis. **That head is superseded** by `dd79b23`, and a fresh run is in flight; this is not a request for anyone to act.

First, the result this PR is actually about — from that same failing run:

```
2 failed, 10601 passed, 638 skipped, 4 warnings in 538.31s
```

**`4 warnings`**, on Windows, where it was ~68,936. The migration does what it claims on that platform too.

### 1. `test_canvases::test_opening_a_canvas_retires_the_legacy_fused_plugin`

```
PermissionError: [Errno 13] Permission denied:
  'C:\...\claude-home-guard625\settings.json'
```

A genuine Windows-only race **in the test**, not in the app. `tests/test_canvases.py:1748` polls the file 20×/second, unguarded, while a background thread rewrites it:

```python
while time.time() < deadline:
    with open(lib.SETTINGS_PATH, encoding="utf-8") as f:   # ← unguarded
```

The writer is `claude_config.lib.write_json`, which is correct — `_atomic_write` writes a sibling temp and `os.replace`s it over the target. On POSIX that is atomic: a concurrent `open` always gets either the old file or the new one, never an error. On Windows, replacing a file has a window in which opening the target raises `ERROR_ACCESS_DENIED`, surfacing as exactly this `PermissionError`.

So the production write is right and the *reader* is missing the platform's retry. The loop already retries — it just does not tolerate the one error the retry exists to ride out:

```python
    try:
        with open(lib.SETTINGS_PATH, encoding="utf-8") as f:
            enabled = (json.load(f).get("enabledPlugins") or {})
    except (PermissionError, FileNotFoundError):
        pass          # mid-`os.replace` on Windows; the next poll sees it
```

Same family as the three fixtures fixed in #939 — a test that assumed POSIX file semantics. **Not fixed here**, because it is unrelated to the lifespan change and this PR should stay reviewable as one thing. Happy to send it as its own PR.

### 2. `test_ai_runtime::test_loading_a_second_text_model_evicts_the_first`

```
AssertionError: org/second never became ready: ... 'state': 'starting'
```

A readiness deadline that expired with the model still `starting` — the load-sensitive shape, on a job that varies 8:47–14:03 with no change in test count. Nothing in this diff touches the AI supervisor.

### Why neither can be this change

The handler set and order are provably identical to `main`: read off `app.router.on_startup`/`.on_shutdown` on the pre-migration commit and asserted in `tests/test_app_lifespan.py`, plus both revisions were started end-to-end and ran the same 19 handlers in the same sequence. Locally, two consecutive full runs on this branch produce a failure list byte-identical to `main`'s.

Watching `dd79b23` for whether either recurs.

---

PR #947, opened 2026-09-01
Sweep orphan part files so a recipe-scoped repo stops reading partial

iamsdas:
## Summary

- An AI model card could read **"partial" forever**, and pressing "Continue downloading" jumped straight to 100% without fixing it. Nothing in the codebase ever deleted the part files responsible.
- The cause is recipe-scoped fetches. `_GGUF_RECIPES` limits `black-forest-labs/FLUX.2-klein-4B` to a `keep` allow-list, so most of the base repo is deliberately never fetched. An earlier, wider fetch of that same repo leaves `.fusedpart`/`.incomplete` files for names the recipe now permanently excludes — and `hub_cache._unfinished_fetch` calls a repo partial on the mere presence of one. The only cleanup that existed, `_clear_parts`, fires from a single call site inside `_fallback()`, reached only when a fetch *raises*. The success path cleaned up nothing.
- Fixed on the downloader side, because it has to be: the checker runs in the server process, which structurally cannot import the runner venv (`formats.py` documents this precedent for `COMPONENT_REPOS`), so the server can never know a recipe's scope.
- New `_sweep_orphan_parts()` runs after any scope's download **succeeds** — the `_cached_path` fast path included, which is precisely the no-op retry case — and removes part files older than a five-minute grace window.

The insight that makes this cheap: **after a download for scope S succeeds, every file S covers is a complete blob by definition, so any part file still in `blobs/` is out of scope.** No etag-to-filename mapping is needed to identify orphans — which matters, because none is possible offline. Our sidecar records only `version/etag/size/segments`, and hf's `.incomplete` has no sidecar at all.

## Visuals

```mermaid
flowchart TD
    A[download_snapshot for scope S] --> B{Scope already on disk?}
    B -->|yes, fast path| C[sweep orphan parts]
    B -->|no| D[fetch]
    D -->|success| E[record fetch]
    E --> C
    D -->|Cancelled| F[raise, keep resume state]
    C --> G{Part file past grace window?}
    G -->|yes| H[remove: out of scope]
    G -->|no| I[keep: presumed in flight]
```

## Usage

Not directly user-invocable — it runs inside the model downloader. The user-visible effect is on the AI Models page: a card stuck on "partial" because of excluded-file leftovers now flips to "downloaded" the next time any download over that repo succeeds, including a "Continue downloading" press that opens no connection at all.

Note the two conditions, both deliberate: the repo reads partial until the *next* successful download over it, and a part file written moments before that retry is left alone as presumed in-flight.

## Test Plan

- [x] `tests/test_ai_worker_base.py` — unit coverage for `_sweep_orphan_parts`: all four part-file shapes removed once aged, fresh files preserved, no-op when `blobs/` is absent, failed unlink logged rather than swallowed. Plus end-to-end: `hub_cache._unfinished_fetch` flips `True` → `False` after the `download_snapshot` fast path runs. **127 passed**
- [x] `tests/test_ai_hub_fetch.py` — a real segmented fetch proving an out-of-scope orphan is swept once a recipe-scoped download completes, and a cancel test proving in-scope resume state survives even past the grace window. **131 passed**
- [x] `tests/test_ai_models_api.py` — `test_the_part_suffix_is_the_fetchers_own` strengthened to pin `hub_cache._PART_SUFFIXES` against both `worker_base.PART_SUFFIX` and `worker_base._HF_PART_SUFFIX`, so the two constants cannot drift. **169 passed**
- [x] `tests/test_ai_diffusers_worker.py` — recipe allow/ignore filtering unaffected. **17 passed**
- [ ] Full suite + CI — running now; PR stays draft until both are green.

### Reviewer note

One consequence worth a second opinion: because the sweep keys off "some scope succeeded", a **deliberately paused** wider download's resume state is removed if a narrower scope later succeeds over the same repo and the part file has aged past the grace window. That is the same mechanism that fixes the bug rather than a separate defect, and a cancel on its own still preserves resume state — but it is a real behaviour change for pause-then-fetch-something-else.

### DIFF
--- a/fused_render/ai/hub_cache.py
+++ b/fused_render/ai/hub_cache.py
@@ -1088,12 +1088,35 @@ def _unfinished_fetch(repo_dir: str) -> bool:
     So two facts, both of which are something that IS there:
 
     1. **A part file in `blobs/`.** Only an interrupted (or in-flight) fetch
-       leaves one: our own `finish()` renames the part over the blob and drops
-       its sidecar, hf does the same with `.incomplete`, and the one path that
-       abandons ours (`worker_base._clear_parts`, taken when a segmented fetch
-       falls back to hf) deletes them before handing the repo over. A cancel
-       does NOT go that way — `except Cancelled: raise` — which is exactly why
-       the bytes, and therefore this evidence, are still here.
+       leaves one — our own `finish()` renames the part over the blob and drops
+       its sidecar, hf does the same with `.incomplete` — and a live fetch keeps
+       rewriting it, so a part file's continued presence is either an in-flight
+       download or one nothing has finished cleaning up yet.
+
+       Two different things clean a part file up, on two different schedules,
+       and the distinction matters to why this predicate eventually stops
+       seeing a repo as partial rather than staying wrong forever.
+       `worker_base._clear_parts`, taken only when a segmented fetch gives up
+       and falls back to hf, drops OUR part files unconditionally and
+       immediately — hf is about to fetch those same names itself. Everything
+       else — a recipe whose `allow_patterns` permanently excludes a name an
+       earlier, wider fetch already left a part file for — is swept by
+       `worker_base._sweep_orphan_parts`, which runs after ANY scope's download
+       succeeds (the `_cached_path` fast path included, so a "Continue
+       downloading" retry that opens no connection still triggers it) and
+       removes any part file — ours or hf's — old enough (past
+       `_PART_GRACE_SECONDS`) that it cannot belong to the scope that just
+       finished. So a repo this predicate calls partial reads that way until
+       the NEXT download attempt of any scope over it succeeds and the
+       leftover part file has aged past the grace window — not forever, and not
+       on the first retry either, since a part file written moments before a
+       retry is presumed to be someone else's in-flight fetch rather than swept
+       out from under it.
+
+       A cancel does NOT go through either cleanup path —
+       `except Cancelled: raise` skips the fallback, and no download succeeded
+       to trigger a sweep — which is exactly why the bytes, and therefore this
+       evidence, are still here for a paused download to resume into.
     2. **No snapshot directory at all.** A folder with blobs and nothing to open,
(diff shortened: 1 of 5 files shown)


iamsdas:
### CI is red from failures inherited from `main`, not from this branch

Every failing job here fails on a test that is already failing on `main` at [`e42b5e83`](https://github.com/fusedio/fused-render/actions/runs/33488986604), with identical assertions. This branch introduces no CI failures.

| Job | Failing test | Assertion | Same on `main` |
|---|---|---|---|
| `test-python` 3.11 / 3.12 / 3.13 | `test_dev_sh_process_cleanup.py::test_shutdown_is_a_single_handler_reaping_every_background_pid` | `CORE_WATCH_PID is not reaped by dev_shutdown` | yes |
| `fused-engine` | same test | same assertion | yes |
| `test-python-windows` | `test_app_git.py::test_legacy_per_app_repo_keeps_committing_into_itself` and `::test_local_monorepo_migration_adopts_per_app_repos` | path-separator mismatch (`C:/...` vs `C:\\...`) | yes, byte-for-byte |

Runs compared: this PR [`33492003152`](https://github.com/fusedio/fused-render/actions/runs/33492003152) against `main` [`33488986604`](https://github.com/fusedio/fused-render/actions/runs/33488986604).

Not one AI test failed remotely. Locally, every test file touching this code passes:

| File | Result |
|---|---|
| `tests/test_ai_worker_base.py` | 130 passed |
| `tests/test_ai_hub_fetch.py` | 132 passed |
| `tests/test_ai_models_api.py` | 169 passed |
| `tests/test_ai_diffusers_worker.py` | 17 passed |
| `tests/test_ai_runtime.py`, `test_hub_models.py`, `test_ai_disk_precheck.py`, `test_ai_benchmark_api.py` | 643 passed, 1 skipped |

Both inherited failures are unrelated to part-file handling — `dev.sh` shutdown signal handling, and Windows path normalisation in the app-git work from #933. Fixing them belongs on its own branch.

PR #948, opened 2026-09-01
Project venvs live at <project>/.venv

iamsdas:
## Summary

- **A project's venv now lives at `<project>/.venv`** instead of centrally at `<home_dir()>/venvs/<key>`. That is the layout every other Python tool already expects — `uv run`, VS Code, and this app's own notebook kernel picker (`templates/notebook/kernel.py`'s `.venv` walk, which has been looking for exactly this file and never finding ours).
- **The home store survives as the fallback, not the rule**, for the folders that genuinely cannot hold a `.venv`: one that ships inside the installed `fused_render` package (the AI runner folders — read-only on the AppImage's squashfs mount and under a Windows `Program Files` install, D376), one that fails a writability probe for any other reason, and any folder at all when `FUSED_RENDER_VENV_IN_TREE=0`. D376's manifest mirror is untouched; its scope just narrows to the case it was written for.
- **Core templates go in-tree too.** A release-time re-stage wipes a staged template's `.venv`, and that is accepted: the uv cache lives outside the staged tree either way and on the same filesystem, so the rebuild is a hardlink relink rather than a re-download.
- **`gc()` gained a relocation arm**, without which every existing home-store venv would leak forever — its project folder still exists, so `_source_is_deleted` correctly refuses to collect it, and nothing would ever read it again. Some of those are multi-gigabyte.
- **`venv_key_for` is unchanged and still load-bearing.** It stopped being the storage path but remains the one per-folder identifier for the install-progress directory, the `/api/env/progress?key=` and `/api/env/cancel` parameters, and the install-dedup lock.

## Visuals

The delta is the placement predicate in `projectenv.venv_dir_for` — previously an unconditional home-store join, now three cheap-to-expensive guards ahead of the in-tree default.

```mermaid
flowchart TD
    A[venv_dir_for] --> B{IN_TREE=0?}
    B -->|yes| H[home store]
    B -->|no| C{inside the package?}
    C -->|yes| H
    C -->|no| D{write probe passes?}
    D -->|no| H
    D -->|yes| E[project/.venv]
    H --> F[keeps the .src mirror]
```

## Usage

Not user-invocable — the placement is automatic and nothing in the run path changes shape. Two things a user can now do that they could not before:

```bash
cd ~/my-app          # a fused-render project folder with a pyproject.toml
uv run python -c "import pandas"   # the app's own environment, not a second one
```

And the notebook template's kernel picker now offers `.venv — my-app` for a project the app has installed.

To opt out — worth it for a project folder under cloud sync or on a network mount, which would otherwise sync a multi-gigabyte venv:

```bash
FUSED_RENDER_VENV_IN_TREE=0
```

## Test Plan

- [x] `tests/test_projectenv.py` — new coverage for the predicate: a writable project resolves in-tree; a folder inside the package resolves to the home store; `FUSED_RENDER_VENV_IN_TREE=0` forces the home store; an unwritable folder falls back; the writability answer is memoised once per project per process and the reset seam clears it.
- [x] `tests/test_projectenv.py` — `gc()` reclaims a home-store venv whose project now resolves in-tree, **and keeps** one whose source folder is merely unreachable (parent directory missing). The second is the dangerous case: getting the ordering wrong would delete a multi-gigabyte venv the moment an external drive is unplugged.
- [x] `tests/test_env_install.py` — placement assertions updated to go through `venv_dir_for` rather than reconstructing a home-store path; renaming a project keeps its in-tree venv with no rebuild.
- [x] `tests/test_engine.py` — a path-shape assertion (`"venvs" in prefix`) replaced with one that asks `venv_dir_for`.
- [x] `tests/test_core_templates.py` — staging never copies a `.venv` out of the package tree (`copytree(ignore=...)`), so a dev checkout's hand-made venv cannot be duplicated into the staging dir on every release-digest change.
- [x] `pytest tests/test_projectenv.py tests/test_env_install.py tests/test_engine.py tests/test_ai_runtime.py tests/test_bundle_contents.py tests/test_template_locks.py -q` → 881 passed, 2 skipped.
- [ ] Manual, on a packaged build: an AI runner's first model download still resolves through the home store and its `<key>.src` mirror (this is the one path with no local coverage — the runner folders are writable in a dev checkout, so the read-only branch is only exercised synthetically).

Docs moved with the code: SPEC.md PY-16/PY-18, a new DECISIONS row (D630) with a pointer amendment on D376, and the module docstrings in `projectenv.py`, `envinstall.py`, `_env_install_worker.py` and `app_git.py` that stated the old rule as a guarantee.


iamsdas:
### CI red is inherited from `main`, not introduced here

`test-python` (3.11/3.12/3.13), `fused-engine` and `test-python-windows` are red on this branch. The failure set is identical to the one on `main`'s own HEAD run ([7dd5efea](https://github.com/fusedio/fused-render/actions/runs/33497611011)), and to the two `main` runs before it:

| Job | Failing test |
|---|---|
| `test-python` ×3, `fused-engine` | `test_dev_sh_process_cleanup.py::test_shutdown_is_a_single_handler_reaping_every_background_pid` |
| `test-python-windows` | `test_app_git.py::test_legacy_per_app_repo_keeps_committing_into_itself` (path separator) |
| `test-python-windows` | `test_app_git.py::test_local_monorepo_migration_adopts_per_app_repos` |

No test outside that set fails. The two Windows `app_git` failures are worth calling out explicitly because this branch does touch `app_git.py` (`_GITIGNORE` gains the write-probe pattern) and `test_app_git.py` — they fail identically on `main` with the same assertions, so they are not caused here.

Locally, the venv-adjacent suite is green: `tests/test_projectenv.py test_env_install.py test_engine.py test_core_templates.py test_templates_api.py test_export.py test_app_git.py test_ai_runtime.py test_server_env_install.py test_bundle_contents.py test_template_locks.py` → **1159 passed, 2 skipped**.

Notably `test-python-windows` passes everything else, which covers the parts of this change most at risk on Windows: the writability probe and the `.venv/Scripts/python.exe` interpreter path.

iamsdas:
Absorbed into #935 and closed in favour of it.

This branch was merged into `install-consent` with no textual conflicts, and all 14 commits are preserved there — nothing is lost by closing this PR rather than merging it.

The two branches overlapped on five files (`envinstall.py`, `_env_install_worker.py`, `projectenv.py`, `tests/test_env_install.py`, `tests/test_projectenv.py`), which are the core of both. Reviewing them separately meant neither PR showed the combination that would actually land, so they are one review now.

The overlap was safe because of this branch's own design: `venv_dir_for` moves the storage path while `venv_key_for` stays the venv's name, and the key is what #935 builds its job ids and progress rows on. Verified after merging — the 8-slot worker argv contract stays coherent (`venv_dir` at 3, `allow_build` at 7), and `test_env_install.py test_projectenv.py test_server_env_install.py test_jobs_api.py` run serially give 391 passed, 1 skipped.

`worktree-in-tree-venv` is not deleted, in case you want the standalone history.

PR #954, opened 2026-09-01
Daemon unification: fused.engine collapses into fused.daemon

iamsdas:
## Summary

- **`fused.engine` is gone.** The warm worker was never a second mechanism — `engine_worker.py` is an HTTP daemon answering `POST /call`, i.e. the daemon you get when you didn't write one. It is now literally that, and `engine_host` drops from three `Child.kind`s to two.
- **A folder declares a daemon by protocol and lifetime, not by kind.** `[tool.fused-render.app]` takes exactly one of `daemon = "x.py"` (your own HTTP surface) or `main = "x.py"` (the shipped worker exposing your `main()`), plus an optional `idle_timeout_s`. The now-meaningless `kind = "background"` key is dropped.
- **Idle retirement is per-child policy instead of a hardcoded kind.** `Child.idle_timeout_s` replaces `APP_IDLE_RETIRE_S` + the `kind == "app"` reaper gate; `0` means resident. Defaults reproduce the previous behaviour — 900s for `main`, resident for `daemon`.
- **One venv rule across both paths.** The venv-unbuilt `409` is gone: a daemon whose project venv isn't built falls back to `sys.executable`, with a shared `unbuilt_deps_reason()` so manual start and autostart describe that condition identically. This reverses half of D500 deliberately.

## Visuals

How a folder's declaration resolves — one manifest key drives protocol, lifetime, and how the page reaches it:

```mermaid
flowchart TD
    A["fused.daemon.*"] --> B{"[tool.fused-render.app]"}
    B -->|"main = x.py"| C["Shipped worker<br/>exposes main()"]
    B -->|"daemon = x.py"| D["Author's own<br/>HTTP surface"]
    C --> E["Reaped when idle,<br/>re-warmed on next call"]
    D --> F["Resident"]
```

## Usage

A folder that just wants its `main()` kept warm writes no daemon at all:

```toml
[tool.fused-render.app]
main = "compute.py"
```

```js
const res = await fused.daemon.run({ action: "list", bucket });
```

`fused.daemon.run(params)` is the shipped-worker convenience over `call` — it unwraps the same `/api/run` envelope the deleted `fused.engine(py).call(params)` returned. A folder that writes its own `daemon.py` is unchanged and still uses `fused.daemon.call(path, body)`.

**Breaking:** `fused.engine`, `POST /api/engine`, and `POST /api/engine/forget` are removed with no alias — same posture as D506's `fused.app` → `fused.daemon` rename. There were no in-tree callers.

## Test Plan

- [x] `tests/test_background_apps.py` — manifest cases for `main =`, per-protocol `idle_timeout_s` defaults, and rejection when both/neither key is present
- [x] `tests/test_daemon_lifetime.py` (renamed from `test_engine_app.py`) — idle-reap coverage ported off the deleted `ensure_app` onto a policy-driven child
- [x] `tests/test_server_ai.py` — per-app-build session isolation, and that an app's shutdown closes its own session
- [x] `bun run test --filter DownloadManager`; `bun run build` succeeds
- [x] **Full local suite** — 11255 passed. The 17 failures are environmental (no `claude` binary, index-scan state) and reproduce identically on unmodified `origin/main`.
- [x] **CI matches `main`'s baseline exactly** — see below
- [ ] Reviewer check: `grep -rn "fused\.engine\|/api/engine\b"` should return hits only in `DECISIONS.md`'s historical entries

### CI status

Every remaining CI failure is pre-existing on `main` and unrelated to this branch:

| Job | This branch | `main` baseline |
|---|---|---|
| `test-python` 3.11 / 3.12 / 3.13 | 1 failed, 11111 passed | 1 (`test_dev_sh_process_cleanup`) |
| `fused-engine` | 1 failed, 11264 passed | 1–2 (same `dev_sh`, plus a flaky slot) |
| `test-python-windows` | 2 failed, 10704 passed | 2 (`test_app_git`, Windows path separators) |
| frontend, bundle-contents, linux-desktop, detect-changes | pass | pass |

## Review

Code review raised six findings; all six are fixed (D632–D638), plus one follow-up found while verifying them (D639):

- **`main =` apps never re-warmed** after the idle reaper retired them — the protocol's core promise. `reap_idle_children` now leaves the Child record so the proxy can heal it, and `daemon.run()` starts the app itself when it isn't known-running.
- **Call-timeout budget** was keyed on reap-eligibility, silently truncating calls for a `daemon =` author who opted into an idle timeout. Now keyed on `child.module`.
- **`_validate_background`** checked the daemon path but not `module`, the thing that actually executes.
- **`version_for`** no longer identified `engine_worker.py` itself, dropping what `APP_WORKER_VERSION` protected.
- **`last_used`** wasn't stamped after spawn, so a short `idle_timeout_s` could expire during bootstrap.
- **AI session ownership** — a module-level `_AiSession` holding an `asyncio.Lock` was shared across app builds, which bound the lock to one event loop and broke 7 Windows tests. Each app build now owns its session, and shutdown closes that session rather than whatever the global currently points at.

## Scope

Stage 1 of 2. Folding `kind="template"` in and generalising `reinit` replay is deliberately **not** here — it rewrites the map viewer's tile path and ships separately. `reinit` replay remains template-only.


iamsdas:
Superseded by #960, which carries stages 1-3 of the daemon unification as one reviewable PR against main. Branch `worktree-daemon-unification` is untouched if this needs reopening.

PR #958, opened 2026-09-02
Delete Child.kind: retry-safety is declared, not inferred from a bring-up path

iamsdas:
## Summary

- **`Child.kind` is deleted.** Stage 1 collapsed background apps and the shipped warm worker into one concept but left the field behind, still naming a distinction the runtime no longer draws. Templates already matched a `daemon =` app on every dimension that mattered — always resident, no `module`, unbounded call timeout, kind-agnostic routing.
- **The one decision `kind` still made is re-keyed onto the property it was actually about.** `at_most_once` in the proxy was `child.kind != "template"`; it is now `not child.retry_post`. Retry-safety is a property of a daemon's HTTP surface, not of which bring-up path produced it.
- **Templates gain a manifest for declaration only — the trust boundary does not move.** The map template's `pyproject.toml` declares `daemon = "daemon.py"` and `retry_post = true`. `engine_host._validate` still gates templates against the hardcoded blessed roots; the manifest is read *after* validation, never as part of it. `~/templates` stays exactly as trusted as it was.
- **Pure re-keying, identical truth table.** Before: true for every background app, false for every template. After: false only for the one template that opts in, true for everything else. No behavior change for any existing child.

## Visuals

```mermaid
flowchart TD
    A["Proxied POST"] --> B{"child.retry_post"}
    B -->|"declared true"| C["Pooled, retry-friendly"]
    B -->|"default false"| D["At most once"]
```

The default inverted in the safe direction: a future non-idempotent template now reads as at-most-once instead of inheriting a blanket "templates are exempt" carve-out unrelated to its actual behavior.

## Usage

A daemon whose POSTs are safely re-runnable opts into pooled retries:

```toml
[tool.fused-render.app]
daemon = "daemon.py"
retry_post = true
```

Omit the key and you get at-most-once, which is what every background app already had. The capability is now reachable by a `daemon =` author, not just by templates.

**Wire change:** `GET /api/engines/running` no longer returns `kind`.

## Test Plan

- [x] `tests/test_background_apps.py` — `retry_post` parsing, its `False` default, and three end-to-end tests (one per `ensure_background` call site) driving real `load_manifest` → real `ensure_background` → real `Child`
- [x] `fused_render/templates/map/tests/test_engines_router.py` — map's POSTs stay retry-friendly
- [x] `tests/test_daemon_lifetime.py`, `tests/test_engines_api.py` — `Child` construction without `kind`
- [x] `bun run test --filter DownloadManager`; `bun run build` typechecks the dropped wire field
- [x] **Full local suite** — 11394 passed. The 39 failures + 1 collection error are environmental; the template block (24 failed / 133 passed / 8 skipped / 1 error) is byte-identical on the base branch, and the rest is the same `claude_health` / `index_scan` / `dev_sh` set stage 1 carried. Template deps (`rasterio`, `joblib`) live in each template's own project venv per D276, not the repo `.venv`.
- [x] **CI matches `main`'s baseline on every job**
- [ ] Reviewer check: `grep -rn "\.kind\b" fused_render/server/engine_host.py` returns nothing

### CI status

Baseline verified against `main` run 33532852829 — the green `main` runs above it skipped every test job via path filtering and are not baselines.

| Job | This branch | `main` baseline |
|---|---|---|
| `test-python` 3.11 / 3.12 / 3.13 | 1 failed, 11116 passed | 1 (`test_dev_sh_process_cleanup`) |
| `fused-engine` | 2 failed, 11268 passed | 1–2 (same `dev_sh`, plus a flaky timing slot) |
| `test-python-windows` | 2 failed, 10709 passed | 2 (`test_app_git`, Windows path separators) |
| frontend, bundle-contents, linux-desktop, detect-changes | pass | pass |

A third Windows failure (`test_canvases`, `PermissionError` on a temp file in an xdist worker) appeared on the first run and did not recur on a re-run of the same job — worker co-location flake, not an interaction with this change.

## Review

Two rounds, both finding defects the scoped tests could not see:

- **The `kind` wire field had a live frontend reader.** `running_engines()` echoed `kind` into `GET /api/engines/running`; deleting it server-side left `RunningEngine.kind` in the TS interface and `{engine.kind}` rendering as a chip on every engines-panel row. The frontend suite missed it because the fixture's `as RunningEngine` cast suppresses TypeScript's excess-property check — that cast is now gone and a test pins the fixture's key set.
- **`retry_post` never reached any background app.** `ensure_background` accepted and stored the flag, but all three production call sites stopped one argument short, so every background app silently got the default regardless of its manifest. Only the map template — which goes through `ensure()` — was ever wired. The existing test missed it by constructing `Child(retry_post=True)` directly and bypassing the manifest path. D640's claim about this plumbing was wrong and is corrected in D645.

Sibling fields `idle_timeout_s` and `module` were audited for the same defect at the same call sites and are correctly threaded.

Also corrected during the build: the handoff claimed `engines.py:145` was the only behavioral branch on `kind`. It was not — `_spawn_env`, `ensure_background`'s reuse-check, `background_running_folders`, `running_engines`, and `restart` all branched on it (D641). Each was redundant with an existing `folder`/`_matches` check and was removed rather than re-keyed.

## Scope

Stacked on #954 — targets `worktree-daemon-unification`, review that first.

Deliberately **not** here, and not deferred either — the investigation found both to be non-work:
- **`reinit` replay was never template-only.** `restart()` calls `_replay()` for every child and the registry is keyed on opaque ids. Handing `daemon =` authors a replay-on-restart capability is new product surface, not unification, so it is not in this PR.
- **No tile-path rewrite.** That cost only materialises if templates adopt `engine_id_for()`'s `bg_<sha1>` hash. Template engine_ids stay readable literals, so `map_render.py`, `template.html`, and the viewer's URLs are untouched.

Template interpreter resolution (`sys.executable` handed over by the caller) is also unchanged — reconciling it with `interpreter_for` is PY-16 territory and buys nothing here.


iamsdas:
Superseded by #960, which carries stages 1-3 of the daemon unification as one reviewable PR against main. Branch `worktree-daemon-stage2` is untouched if this needs reopening.

PR #1003, opened 2026-09-04
App doctor: review one app before it is shared

iamsdas:
## Summary

- **`fused-render doctor` reviews one app folder before it goes anywhere.** Five checks, all mechanical: leaked credentials, absolute paths that only resolve on the author's machine, generated caches or logs loose in the tree instead of `.fused/`, missing structure (entry page, README, `preview.png`, parseable `pyproject.toml`), and a declared `fused-api-version` behind current.
- **The engine forms no opinion about the app's code.** Judging whether a real `fused.*` call is used correctly belongs to the nine skills that already own those surfaces, each maintained against the runtime. `SKILL.md` routes there instead of re-deriving that judgment: `fused.ai` to `fused-render-ai`, `runPython`/`params` to `fused-render-authoring`, `trackJob` to `fused-render-jobs`, and so on. That mapping is read from each skill's own `description:` line so it can be re-checked when one changes.
- **Severity lives in one table, not per caller.** `secrets` and `device-path` are HIGH and fail a run under `--check`; `structure`, `api-version` and `generated` are LOW and never fail one. A single `_SEVERITY` map means the skill and the CI workflow cannot drift into disagreeing about what counts as broken.
- **Findings are masked at the source.** No finding's `excerpt` ever carries a whole matched secret, because reports land in CI logs and chat transcripts. `tests/test_app_doctor.py` asserts this rather than leaving it to convention.
- **CI covers the mechanical half only, because a GitHub runner has no reader in it.** `ci/app-check.yml` loops the community-apps repo's app folders and calls the doctor once per app. The routing pass above is deliberately not wired into it.

`fused_render/app_doctor.py` is stdlib-only and imports nothing from the rest of the package, so the few rules it shares with `app_listing`, `current_apps` and `fused_api_version` are duplicated deliberately.

## Visuals

The command handles what wants determinism. The skill handles what needs a reader, by handing off rather than answering.

```mermaid
flowchart TD
    SKILL[Skill: review before sharing] --> CLI
    CI[Community-apps CI] --> CLI
    CLI["fused-render doctor"] --> ENGINE["app_doctor.check()"]
    ENGINE --> HIGH["HIGH: secrets, device-path"]
    ENGINE --> LOW["LOW: structure, api-version, generated"]
    HIGH --> FAIL["--check exits non-zero"]
    LOW --> PASS[Reported, never fails]
    SKILL --> ROUTE{"Which fused.* surfaces?"}
    ROUTE --> OWNER["Load the skill that owns each one"]
```

## Usage

Review the app in the current directory, or one named by path:

```
fused-render doctor
fused-render doctor --check path/to/app
fused-render doctor --json path/to/app
```

Human output groups findings by family and closes with a count. On a folder with a planted key, a home-folder path, and a loose cache file:

```
device-path (HIGH):
  index.html:6  /home/<user>/secrets/pantry.csv
  shopping.py:2  /home/<user>/Documents/pantry.csv

secrets (HIGH):
  shopping.py:1  OPENAI_API_KEY = sk************************************56

generated (LOW):
  shopping-cache.db  looks generated (belongs in .fused/, or shouldn't be committed)

structure (LOW):
  .  no README in the app folder
  .  no preview.png thumbnail (or it is empty)

6 finding(s), FAILING under --check
```

`ci/app-check.yml` is written for `fusedio/fused-render-community-apps` and takes effect only once someone copies it into that repo's `.github/workflows/`. Nothing in it needs filling in, and that copy is a manual step outside this PR.

## Test Plan

- [x] `tests/test_app_doctor.py` — one case per secret and device-path rule, plus the masking guarantee asserted directly.
- [x] `tests/test_app_doctor_housekeeping.py` — a positive and a negative case per structure, version and stray-file check.
- [x] `tests/test_app_doctor_cli.py` — grouped human output, `--json`, `--check` exiting non-zero only when a HIGH family fired, and a nonexistent path failing loudly rather than reporting phantom findings.
- [x] 106 passing across those three files plus `tests/test_skill_plugin.py`.
- [x] `ci/app-check.yml`'s shell loop run locally against fixture app folders: an app with declared dependencies installs before its tests, an app without a `pyproject.toml` tests anyway, an app with no test files is reviewed but not tested, and pytest's exit 5 is not mistaken for a failure.
- [ ] Repo CI — see the note below; not yet green.

Three things a reviewer should weigh, none of them settled:

- **The secret and device-path regexes need a tuning round, and they are now the whole HIGH tier.** A review found `/media/` inside a CDN URL and a README mentioning `/tmp/scratch` both reported HIGH, which would fail CI for entirely portable apps; and in the other direction, the placeholder filter drops any value merely starting with `my`/`test`/`sample`, so `DB_PASSWORD = "mysql-prod-9f3k2xyz"` scans clean, while the assignment pattern requires quotes, so a `.env` holding `DB_PASSWORD=supersecretvalue123` matches nothing. Both directions are bad: false HIGHs teach people to delete the workflow, false negatives are what the feature exists to prevent.
- **`structure` pools a missing entry page (the folder is not an app at all) with a missing README (a nicety),** both at LOW.
- **`APP_DOCTOR_DECISIONS.md` is this build's working notes, now 503 lines — longer than the engine.** It should not land on `main`.


iamsdas:
Superseded by #1005, which carries the same work on a branch whose history matches its contents.

PR #1075, opened 2026-09-08
Native Claude chat (PR4/4): scheduled runs, live watch, external-run follow, artifact strip

AkshilVT:
Stacked on #1074 (PR3) → #1064 (PR2) → #1061 (PR1). Merge in order; bases retarget as each lands.

## What

Completes the port of the claude template into the native chat: scheduled runs, a standing live watch, external-run follow, and the artifact strip. Flag-gated like PR1–PR3; flag off unchanged.

- **Scheduled messages** — the composer block (banner with the template's row, `state · when` text, "N more after it.", disabled textarea + calendar pill carrying the reason), a 15 s schedule poll that baselines fired entries at load and attaches this session's runs ("◷ Your scheduled message is running now."), foreign fires noted once, two-press stop arming for repeats with outside-`pointerdown` / Escape disarm, the refusal note, and row click through to the Tasks calendar.
- **Standing live watch** — 5 s tick gated on visibility plus `storage`/`focus` pokes; adopts a run started in another tab and follows a transcript moved by a run outside this app ("Running outside this app", no stop button), refreshing history without a skeleton and keeping scroll position.
- **Artifact strip** — chips above the footnote during a chat, fed by every eighth poll and by run end; the snapshot list is invalidated on run end.
- **Stop always works** during a live turn even while a schedule blocks the composer; schedule reset happens only on a transcript replacement.
- Behaviour note carried from the review: a `done` probe that carries no `message` now repairs nothing on the non-`matches` path (matches the template, T:17836-17849) — the two run-controller fixtures gained a `message` for that reason, not to fit the code.

## Verification

- typecheck · boundaries (643 files) · `bun test` 4211/0 · build.
- Browser QA (cmux, `.claude-design/qa/pr4-round1.md`) on the split layout: composer banner copy/geometry/disabled states, cancel round-trip (banner clears, composer re-enabled, draft preserved), same-session scheduled auto-fire (note → stream → banner clear, no reload), cross-tab live watch (new turn in ~3 s, no skeleton, no scroll jump), adopt-without-`run` and bogus-`run` recovery card, an external `claude --resume` run detected with the transcript auto-refreshed, snapshot invalidation after an edit; legacy sanity on two URLs. Zero console/page errors in every state tested, native and legacy.
- Code review (`.claude-design/qa/pr4-review1.md`): 1 blocker (send/stop button disabled under a composer block, so a user could not stop their own streaming turn), 2 majors (frozen `due` in the schedule absorb dedupe; the schedule reset keyed on session id instead of a transcript replacement) and the minors all fixed, with tests added for the gaps that let them through.

---

> [!NOTE]
> **Medium Risk**
> Touches run adoption, schedule polling, transcript refresh, and composer blocking in the main chat path; behavior is heavily tested but coordinates multiple timers and shared controller state.
> 
> **Overview**
> **PR4 wires scheduled runs, a standing live watch, artifact strip, and related run-loop behavior into native `ClaudeChat`.**
> 
> The chat now runs **`useSchedule`** with a **`SchedBlock`** above the composer: pending scheduled messages block the composer (and screenshot/comment/mic seats, but not stop mid-recording), reset the schedule baseline only on **`transcriptGen`** (transcript replacement) or explicit Back—not on mount or first session id—and use **`history: replace`** when clearing fired `run` params or spent `msg` anchors.
> 
> A new **`createLiveWatch`** polls every 5s (plus storage/focus/visibility/`TASKS_CHANGED_EVENT`) to adopt external runs, refresh history without skeleton, and drive an “running outside this app” line via **`setExternalWorking`**, with pure **`followDecision`** logic and guards against own-echo and duplicate adoption (**`hasShownRun`**, **`ownRunEndedAt`**).
> 
> **`ClaudeChat`** also delays **`onReady`** until recent sessions load (or chat/no-agent paths), sets **`leftLive`** on Back for extra session list reads after mid-turn exits, hooks artifact polling and snapshot invalidation on run end, **`useRepairScroll`** on **`repaired`**, and refocuses the composer after re-attach. Session list titles no longer leak truncated `<annotations>` tags.
> 
> Extensive tests cover boot schedule reads, landing ready, live watch, run-controller PR4 paths, **`RecControls`** blocking, and params **`replace`** for anchors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d28b91136448dda6b4b43635d9f9ed63f3573055. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AkshilVT:
bugbot run

PR #1086, opened 2026-09-09
Mode dropdown: "Render" becomes "Preview" with the app star icon

vasu2001:
Opening a fused HTML file, the mode dropdown labelled the file's own view "Render" with a play-in-a-box glyph. It now reads **Preview** and wears the brand four-point star (`AppStar`), the same mark an app without `icon.svg` gets on the Projects row, the app cards and the app page.

- `mode-name.ts`: `_render` → "Preview" (+ its bun test)
- `ModeSwitcher.tsx`: `RENDER_SENTINEL_ICON` is `<AppStar>`; imports the shared component instead of a local SVG
- `templates/claude/template.html`: `paneModeLabel` mirror follows (+ `tests/test_claude_pane_mode_label.py` pin)
- `SideChrome.tsx`: comment only — the folder pane's Preview side keeps its play glyph, since it means "this row's default view", not always this mode

tsc clean. Visual smoke pending.

---

> [!NOTE]
> **Low Risk**
> Display-string and icon alignment across pickers with test updates; no auth, routing, or data-path changes.
> 
> **Overview**
> The **`_render`** sentinel (the file’s own HTML in an iframe) is renamed from **Render** to **Preview** everywhere users see it: `mode-name.ts`, the Claude template’s mirrored `paneModeLabel`, and matching unit/integration tests.
> 
> Its icon is unified with the explorer mode menu: **`AppStar`** replaces the old play-in-a-box SVG in `ModeSwitcher.tsx`, and the chat **left-mode picker** renders the same star for `_render` (with new **`c-mode-ic-star`** CSS so the mask-based icon styles don’t show a square behind the SVG). The folder listing pane’s **Preview** side tab still uses the play-in-frame glyph—comment-only clarification in `SideChrome.tsx`.
> 
> **`mode-labels.test.ts`** now derives `isDir` from registry keys (`/` suffix) instead of testing both `true`/`false` for every key, so file-only labels like `_render` → “Preview” don’t false-fail collision guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26abab78acb59f42574729a47a83377f80a63b6e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vasu2001:
Bugbot's find was right about the symptom, wrong about the fix: `_render` is only ever offered on file keys (`_match_registry` / `_key_segments` refuse a bare key for a directory), so the guard's `isDir=true` pass over `.html` was a phantom list. Guard now derives `isDir` from the key's trailing `/` instead of permuting it. Both mode tests green locally.

vasu2001:
CI state after the fixes:

- **frontend, Bugbot, linux-desktop, bundle-contents**: green.
- **fused-engine**: red once on `tests/test_env_install.py::test_a_retry_inside_the_poll_window_leaves_one_live_mirror_thread` (thread-count race, file last touched by #1084); green on re-run.
- **test-python-windows**: two failures, neither from this branch:
  - `test_git_status_listing.py::test_parse_porcelain_decodes_a_non_utf8_name_like_scandir_would` — `UnicodeDecodeError` on Windows. Fails on main's last two runs too, since #1072.
  - `test_claude_config_api.py::test_config_lock_held_commit_does_not_deadlock` — 15s worker timeout on the Windows runner; not seen on main's recent runs, one-off.

Nothing here touches Python beyond a string in `tests/test_claude_pane_mode_label.py`, which passes.

PR #1109, opened 2026-09-11
Native chat: "Recap:" line after an absence (Claude Code session recap parity)

AkshilVT:
## What

The native Claude chat draws Claude Code's own session-recap line when you come back after being away ≥60s: the last row of the log reads `※ Recap: <1–2 plain sentences>` — where the conversation stands and the one next action. Native chat only; legacy template untouched.

## Why not read Claude Code's recap

It does not exist on disk. Claude Code (2.1.268) generates its "Session recap" in memory on terminal blur by forking its live cache-safe params, shows it as a `recap_fold` row, and never writes it to the session JSONL or the sessions registry. Forking the session ourselves (`--resume --fork-session --no-session-persistence`) works but is a full cache miss (~170k tokens, ~$3 on a long session; haiku refuses "Prompt is too long"). So we summarize a transcript tail instead.

## Server

`GET /api/claude-sessions/recap?file&session_id&for_uuid` → `{text, for_uuid, at}`; `text: ""` means "nothing to show" and is never an error.

- Tail = last 8 turns from `_history()`, user + assistant prose only, ~12k chars, wrapped in `<transcript>` tags with the instruction after it — a raw tail made haiku *answer* the last user line ("ok", "Gibberish. What's the task?").
- `claude -p --no-session-persistence --max-turns 1 --tools "" --model haiku --output-format json`, ~10s.
- Output stripped of markdown, one-word answers refused, 400-char cap (Claude Code's own cap).
- Cached per (session, turn), single-flight per key, 25s timeout, failures cached 60s.

## Frontend

- `useAwayRecap`: fetch on **return** (we cannot pre-warm on blur like Claude Code can). Gates: 60s away, pref on, no turn running, empty composer draft, turn not already answered, <2 failures, root visible, one fetch per page per return.
- **Opt-in per mount** (`recap` prop on `ChatMount`, default off). Only the three primary chat sites pass it — explorer content pane, `?_side=claude` sidebar, and the folder listing's companion pane (where `/explorer/view/<dir>?session_id=` lands). Without this the tasks wall's per-card chats fired seven model calls on one focus.
- Anchor skips Claude Code's `[Request interrupted by user]` user row (has a uuid, renders without one).
- Row is a `note` turn inside the log: `※` glyph, bold `Recap:`, italic body at reading size. Not clickable, no dismiss; goes away on the next send.
- Pref **Session recap** (`chat.recap`, default on) under Native chat (beta) in Preferences.

## Testing

- bun: 5232 pass; pytest: 108 pass on the touched suites (6 new server tests: cap + argv, `is_error`, cache, single-flight, user-last skip, 400s; sanitizer + prompt wrapping).
- Browser (cmux): real OS-level hide/show ≥65s fires exactly one request and one row; 20s away fires nothing; Tasks and Home pages fire zero; row absent for a session whose last turn is the user's; legacy chat unaffected.

Design notes: `.claude-design/session-recap.md` (local, not committed).

---

> [!NOTE]
> **Medium Risk**
> Triggers paid CLI/model work on focus return and touches prefs plus a new API endpoint, but heavy gating (opt-in mounts, visibility, dedupe) limits blast radius; server failures are silent empty text.
> 
> **Overview**
> Adds **session recap** ("While you were away") to the native Claude chat: after ≥60s away, an opt-in primary mount can show a `※ Recap:` line at the bottom of the transcript via `useAwayRecap` and `GET /api/claude-sessions/recap`.
> 
> **Cost control** is central: `ChatMount` gains an **opt-in `recap` prop** (default off — prop omitted, not just falsy) so tasks-wall card chats do not each fire a ~12s model call on window `focus`. Only explorer content pane, `?_side=claude` sidebar, and folder listing companion pass `recap`. The hook also dedupes one fetch per page return, requires on-screen root, empty composer, no running turn, and reads `prefs.chat.recap` (default on) from the **same `/api/prefs` read** as native chat in `feature-flag.ts`.
> 
> **Server** builds a prose tail from `_history`, runs a one-shot `claude -p` (no session persistence, no tools), returns `text: ""` on failure (200), with cache + single-flight per `(file, session_id, for_uuid)`.
> 
> **Anchor fix**: `INTERRUPT_MARK` moves to `protocol/wire.ts`; `recapAnchor` skips interrupt user rows so recap scroll targets match `data-msg` turns. **Preferences** adds Session recap toggle; **Transcript** renders optional recap and turns off tail-follow when `?msg=` anchor scroll runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce16d69992c9516f493b062bf2d3a879e29825ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/src/apps/claude/ChatMount.render.test.tsx
+++ b/frontend/src/apps/claude/ChatMount.render.test.tsx
@@ -14,7 +14,7 @@ const { createMemoryParamsStore } = await import("./params/store");
 // The native branch is a `lazy` chunk, so an `act` that does not AWAIT pins the
 // Suspense fallback and nothing else. Resolving the module once, here, makes
 // every mount below able to reach the chat itself inside one async `act`.
-await import("./ClaudeChat");
+const { ClaudeChat } = await import("./ClaudeChat");
 const { publishNativeChatEnabled, resetNativeChatFlagForTests } = await import("./feature-flag");
 
 // NO PREFS GET FROM THIS FILE. Every test publishes the flag directly, but the
@@ -262,4 +262,24 @@ test("a host id that arrives later pushes only its own key", async () => {
   // A run id of its own still lands.
   act(() => r.update(<Probe sessionId="s3" runId="r2" />));
   expect(wrote[wrote.length - 1]).toEqual({ run: "r2" });
+});
+
+test("the recap is OPT-IN: absent unless the host asked for it", async () => {
+  // The bug this pins: "While you were away" is a ~12s model call fired by
+  // window `focus`, which EVERY mounted chat hears. A tasks wall spent seven of
+  // them on one return. Default-off is the guarantee — a new embed site cannot
+  // inherit the cost by not thinking about it — so what is asserted is the
+  // absence of the prop, not merely a falsy one.
+  publishNativeChatEnabled(true);
+  const off = await mountAsync(
+    <ChatMount file="/w/p" chatOnly legacySrc={SRC} paramsSource="url" />,
+  );
+  const propsOf = (r: ReturnType<typeof create>) =>
+    r.root.findByType(ClaudeChat).props as Record<string, unknown>;
+  expect("recap" in propsOf(off)).toBe(false);
+
+  const on = await mountAsync(
+    <ChatMount file="/w/p" chatOnly legacySrc={SRC} paramsSource="url" recap />,
+  );
+  expect(propsOf(on).recap).toBe(true);
 });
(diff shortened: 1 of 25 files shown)


AkshilVT:
Both Bugbot findings fixed in 9f058e67f.

**Interrupt blocks session recap** — right, and it was the worst possible case to lose: hitting stop and walking away is the usual way to be away. `_recap_tail` now drops the `[Request interrupted by user]` marker entirely (a new `_INTERRUPT_MARK`, the server-side twin of `protocol/wire.ts`'s), so it is neither summarized nor counted as the user having spoken last.

**Recap cache ignores file path** — right on both halves. The cache and single-flight ledger are keyed on `(canonical file, session_id, for_uuid)` now. Successes also expire after 15 minutes instead of never: the key names the last *user* turn, and the assistant's answer to it can still be growing when a run is driven from a terminal outside the app, so a recap taken mid-reply is no longer pinned forever.

Regression tests for both in `tests/test_claude_session_recap.py` (10 passed; 222 across the touched server suites).

PR #1123, opened 2026-09-11
Native chat: the reload and adopted roads no longer flash the old reply into the new bubble

isaacbrodsky:
Split out of #1119, which is now just the legacy-chat fix for the reported duplication. **This PR is a different, narrower bug** and is not what that report was about — see "Scope" below.

## What this fixes

`pollLoop` opens on a window that can still hold the PREVIOUS reply: `_read_current_turn`'s cursor only advances past a turn boundary it has proven (a `_starts_new_turn` echo with a `result` before it), so a send into a still-live host reopens the window on the reply already on screen and keeps it there for several polls. Two roads reached that window with no base to hide it behind:

- **The reload road.** `landedWindow` only ever covers a reply *this* controller watched land, so a transcript restored from `_history` — a reload, or a second tab opened on the same conversation — had nothing there even though the reply was rendered. `priorReply`/`settledReply` is the base for those.
- **The adopted road** (`resumeRun`/`adoptLiveRun`) started its loop with no base at all. `sendMessage` had always seeded one; the difference was never the window, only which road opened it.

Plus, in `agent.py`, `_absorbed_turn_breaks` skipped the window's own opening echo *by index* (`rows[0]`). A window reopened on a live host doesn't start with that echo — the CLI writes `system/init` and `system/status` first — so the window's own turn was counted as a follow-up folded into a reply that hadn't begun, its closing `result` reported a fold-in seam for a payload holding one turn, and the genuinely new echo later in the window had nothing left to report a boundary against.

## Scope — please read

This does **not** fix the duplication reported against the default chat. That is #1119, in `templates/claude/template.html`. I spent a long time fixing the wrong implementation before noticing that `native_chat_enabled` defaults to off and that the legacy template contains no reference to `turn_breaks` at all.

Measured on `126b06c` (pre-any-fix), same scenario driven through the real app:

| chat | peak copies of the previous reply |
|---|---|
| native | 1 — clean |
| legacy (default) | 2 — duplicates |

So the native chat never had the reported bug; it was born with `landedWindow`. What it *did* have are the narrower gaps above, which is what this PR closes. They are worth having, but they are not urgent, and nothing here is load-bearing for #1119.

The `agent.py` change is likewise invisible to the default chat, which never reads `turn_breaks`.

## Verification

Every fix here has a regression test confirmed failing against the commit before it. The later commits in this branch are corrections to earlier ones, each prompted by a Cursor Bugbot finding on #1119 that turned out to be correct — the history is kept rather than squashed because those exchanges are the reasoning.

`bun test src/apps/claude` 1562 pass · `tsc --noEmit` clean · `tests/test_claude_followup_seam.py` + `test_claude_agent_segments.py` 82 pass.

Rebased on `main` (`0f6f36ddd`).

https://claude.ai/code/session_01SUo4BiYdwk9GDnW7qT4AeL

---
_Generated by [Claude Code](https://claude.ai/code/session_01SUo4BiYdwk9GDnW7qT4AeL)_

---

> [!NOTE]
> **Medium Risk**
> Changes live poll rendering and Python turn-break/segmentation logic for multi-tab and post-reload sends; wrong seam handling could drop or duplicate turns, but behavior is heavily regression-tested.
> 
> **Overview**
> Fixes a **one-frame flash** where the previous assistant reply could appear inside the new bubble when sending after a reload, opening a second tab on the same live conversation, or when the backend withholds `turn_breaks` (e.g. rows between an old `result` and the new echo).
> 
> **Native chat (`run-controller.ts`)** seeds `pollLoop` with the **already-rendered prior reply** via `settledReply()` / `priorReply` on `sendMessage` and `resumeRun`/`adoptLiveRun`, skipping the optimistic user bubble so the composer path is covered. Segment slicing stays **untrusted** until a reported seam confirms the count or the window no longer opens on that text—avoiding history/live segment mismatches, stuck `sending`, and one-poll “grace” mistakes on busy hosts. Seam adoption uses the **first** boundary past the base (not the last) and a narrower **confirm-base** path for adopted loops.
> 
> **Backend (`agent.py`)** breaks segments at genuine new user turns when a `result` closed since the last echo (so wake continuations don’t merge into the next reply), reports turn seams **without** requiring `result` adjacency to the echo, and treats the window’s **opening** echo correctly when `system/init` rows precede it.
> 
> Regression coverage is expanded heavily in `run-controller.test.ts` and `test_claude_followup_seam.py`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63c014c2c6a45daf77061e15ecb6e3acae77a06b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/src/apps/claude/protocol/run-controller.test.ts
+++ b/frontend/src/apps/claude/protocol/run-controller.test.ts
@@ -1381,17 +1381,23 @@ describe("follow-ups (T:16024, D687)", () => {
             await controller.sendFollowUp("and one more thing");
             return poll({ segments: [A], text: A.text });
           }
-          // Now a seam arrives — but this one closes the reply the follow-up
-          // was folded into, so it is OURS, not the base.
+          // Now the echo lands, and the payload carries TWO seams: the
+          // pre-send boundary (end of the reply already on screen) and the
+          // `result` closing the reply the follow-up was folded into. The
+          // first is the base; the second is ours.
+          const twoSeams = [
+            { segments: 1, text: A.text.length },
+            { segments: 2, text: A.text.length + B.text.length },
+          ];
           if (n === 2) {
             return poll({
               segments: [A, B, C],
               text: A.text + B.text + C.text,
-              turn_breaks: [{ segments: 2, text: A.text.length + B.text.length }],
+              turn_breaks: twoSeams,
             });
           }
           return poll({ done: true, segments: [A, B, C], text: A.text + B.text + C.text,
-                        turn_breaks: [{ segments: 2, text: A.text.length + B.text.length }] });
+                        turn_breaks: twoSeams });
         },
       },
       params,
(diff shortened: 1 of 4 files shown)


isaacbrodsky:
### `fused-engine`: `test_desktop_probe` again — third occurrence, and it is not this PR's

**What failed:** one test — `tests/test_desktop_probe.py::test_matching_server_true_for_our_token`, `assert False is True`. `1 failed, 12310 passed` in 5m10s.

**This is the third time the identical test has failed** across this work, and the second diff it has blocked:

| head | PR | diff | result |
|---|---|---|---|
| `379a29c` | #1119 (pre-split) | native + agent.py | failed |
| `379a29c` re-run | — | same | passed |
| `d29f7d2` | #1119 | + template.html | passed |
| `0ae04ef` | #1119 | + template.html | failed |
| `63c014c2c` | **#1123** | native + agent.py, rebased on main | **failed** |

Two different diffs, one of which (#1119 at `0ae04ef`) does not contain this PR's changes at all. That rules out a common cause in the code.

**Why it is not this PR's:**

1. `test-python` 3.11/3.12/3.13 run the same file and pass on these same commits — `test_desktop_probe.py` is not gated on the fused engine (its own docstring: *"Pure stdlib + a real local uvicorn instance; no pywin32, so this runs on every platform"*).
2. `fused-engine` is green on `main`.
3. No mechanism. This PR touches `run-controller.ts`, its test file, `agent.py`'s `_absorbed_turn_breaks` (a pure function over parsed rows), and one test module. None of it is reachable from an HTTP probe against a local uvicorn.

**Cause** — two defects in the file's own `_serve` fixture, neither in product code:

- It picks a port with `_free_port()` (bind `:0`, read it, **close**), then hands that port to uvicorn — unowned in between, with `pytest-xdist` running workers in parallel. If the bind then loses, the wait loop simply times out and `yield port` proceeds anyway, testing nothing.
- `matching_server` treats *any* transport failure as a plain `False` by design, and this is the only test in the file that both asserts `True` and gives the server a single `urlopen(timeout=0.5)` shot. Its sibling `test_wait_until_ready_true_when_server_answers` polls with a 10 s budget against the same fixture and has never failed; the two that assert `False` pass whether or not the request lands. `fused-engine` is the heaviest job in the matrix, which is where a 0.5 s budget is missed.

**Patch** (not applied — it belongs to that file's own area, and neither this PR nor #1119 otherwise touches it):

```python
@contextlib.contextmanager
def _serve(app):
    # Port 0: uvicorn binds first and we read back what it got, so nothing can
    # take the port in the gap between probing for a free one and binding it.
    server = uvicorn.Server(uvicorn.Config(app, host="127.0.0.1", port=0, log_level="warning"))
    thread = threading.Thread(target=server.run, daemon=True)
    thread.start()
    try:
        deadline = time.monotonic() + 10
        while not server.started and time.monotonic() < deadline:
            time.sleep(0.02)
        # Never yield an unbound port: the probe would answer False and the
        # test would read that as a contract failure.
        assert server.started, "uvicorn never bound"
        yield server.servers[0].sockets[0].getsockname()[1]
    finally:
        server.should_exit = True
        thread.join(timeout=5)
```

…and giving the failing assertion the same polling budget its sibling already uses, rather than one 0.5 s shot.

**Re-run:** this PR's one re-run is unspent. The run is still finishing (`test-python` ×3 and `test-python-windows` in progress), and GitHub refuses `rerun-failed-jobs` while a run is live, so it will be spent as soon as the run completes.

Everything else on this head is green: `frontend`, `bundle-contents`, `windows-desktop`, `linux-desktop`, `detect-changes`, and **Cursor Bugbot with no findings**.

---

PR #1128, opened 2026-09-13
.fused app file: opaque container instead of a renamed zip

vasu2001:
## Why

Gmail flags an exported `.fused` as suspicious/dangerous. Mail scanners classify attachments by sniffed bytes, never by extension, so the "zip renamed to `.fused`" was never a disguise: Gmail saw `PK`, walked the members, and judged the `.html` + `.py` inside a script-shaped archive. A real export (`Random.fused`) had only `json/png/svg/html/md` members, nothing on Google's published block list, so the trigger is the archive-with-scripts heuristic, not a banned extension. Hiding member names inside a zip would still be a zip; the owner chose a format nothing classifies as an archive.

## What

- **`fused_render/appfile_container.py`** (new): the v2 format. `FUSEDAPP` magic, u16 version, deflated JSON index (`fused_app_file: 2`, `name`, `entry`, and every member's `path/offset/size/csize/sha256`), then one deflate stream per file. Offsets are relative to the data section so the writer streams the payload in one pass.
- **Hardened reader**, since a mailed `.fused` is as untrusted as an uploaded template pack: index inflate bounded on both declared and yielded size; every path relative, forward-slash, no `..`/`\`/NUL/empty segment, unique, never both a file and another file's directory; entry/per-file/total caps and in-file offsets checked before any byte lands; `extract` verifies each body's written length and sha256 against the index and removes the stage on any disagreement.
- **`appfile.py`**: `export_app_file` writes v2 only. `read_manifest`, `read_preview`, `open_app_file` branch on the magic, so every v1 zip already exported and indexed (D396) keeps opening through `zip_import` unchanged. Extension, UTI (`io.fused.render.app` → `public.data`, now true of the bytes too), routes, template binding and the `application/octet-stream` download are untouched.
- SPEC §43 intro + AF-6 amended; DECISIONS D870 records the diagnosis, the option not taken, and the accepted cost (a second hardened extractor, which D386 avoided).

## Tests

`tests/test_appfile_container.py` pins each guard with a crafted file (bad magic, future version, index bomb, traversal paths, file/dir conflicts, caps, out-of-bounds offsets, body/index disagreement, bounded member read). Existing appfile/clone/exported-apps tests updated from zip-member assertions to index assertions. 72 pass locally.

## Not verified

Whether Gmail passes a v2 file is a hypothesis until one is mailed. Export any app on this branch and attach it to a Gmail message; if it still flags, the format is also what the next experiment needs.

Lost on purpose: `unzip -l` inspection of a `.fused`. A `fused appfile ls` is the follow-up if anyone misses it.

---

> [!NOTE]
> **High Risk**
> Adds a second untrusted-input extractor and changes the on-disk export format while keeping v1 read paths; mistakes in validation or branching could affect open/clone of mailed or legacy `.fused` files.
> 
> **Overview**
> Exports now produce **v2 `.fused` files** as an opaque `FUSEDAPP` container (versioned header, zlib-compressed JSON index with `fused_app_file: 2`, per-file deflate streams) instead of a zip renamed to `.fused`, so mail scanners no longer see `PK` and script-shaped archive members.
> 
> **`fused_render/appfile_container.py`** implements write/read/extract with the same hardened posture as `zip_import`: bounded index inflate, path validation, caps, offset checks, and sha256/size verification on extract (stage removed on failure).
> 
> **`appfile.py`** routes export through `container.write`; `read_manifest`, `read_preview`, and `open_app_file` branch on magic—**v1 zips remain readable** via existing `zip_import` paths. SPEC §43 and DECISIONS D870 document the change; appfile/clone tests assert index members instead of zip listings; **`tests/test_appfile_container.py`** pins each guard.
> 
> User-visible routes, extension, and download MIME type are unchanged; **`unzip -l` no longer applies** to new exports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9886d34fe6f85143a46d68c26bc369153f2e0dc7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


isaacbrodsky:
>     fused_render/appfile_container.py (new): the v2 format. FUSEDAPP magic, u16 version, deflated JSON index (fused_app_file: 2, name, entry, and every member's path/offset/size/csize/sha256), then one deflate stream per file. Offsets are relative to the data section so the writer streams the payload in one pass.

Did you try rewriting the magic bytes instead? That would leave us still able to use standard zip libraries after resetting the magic.

PR #1181, opened 2026-09-16
Surface the index-rank timing breakdown in the browser console on slow searches

iamsdas:
## Summary

- A customer reporting 6-7s home searches cannot be asked to open `$TMPDIR/fused-render-<pid>.log`, but they do have the browser console open with verbose enabled. This puts the timing breakdown there instead.
- `api_index_rank` already computes `total_ms`, `lane_wait_ms` and `worker_ms` and throws them away into a log line. This returns them on the wire as an optional `timing` object, and the frontend emits a single `console.warn` when a search takes **2s or more**.
- **The point of the change is the gap.** The UI's number is measured client-side (`issuedAt.current` to the response landing); the server's `total_ms` starts only at handler entry. Everything before that — browser connection queueing, ASGI accept backlog, transit — is invisible to `total_ms`, `lane_wait_ms` *and* `worker_ms` alike. Reporting both makes that unaccounted time visible for the first time.
- That distinction is load-bearing for triage: a large `lane_wait` means lane contention, a large `worker` means the user's own filesystem I/O is slow, and a large *unaccounted* gap means the handler was fast and the problem is not on the server side at all — three different fixes.

## Visuals

```mermaid
flowchart TD
    A[Keystroke issues indexRank] --> B[Client starts the clock]
    B --> C[Gap: queueing, backlog, transit]
    C --> D[Handler entry: server clock starts]
    D --> E[lane_wait, then worker]
    E --> F[Response carries timing]
    F --> G{elapsed >= 2s}
    G -->|yes| H[console.warn with breakdown]
    G -->|no| I[Silent]
```

The server clock covers only `D` to `F`. `C` is what nothing measured before this change, and it is reported as `unaccounted/outside-handler`.

## Usage

Not user-invocable and not visible in the UI — nothing is rendered, and normal searches print nothing. During a slow episode the browser console shows one line per affected search:

```
[explorer] slow home search: query="readme" elapsed=2100ms | server: total=1800ms lane_wait=50ms worker=1750ms | unaccounted/outside-handler=300ms
```

Against a server that predates the field, it degrades honestly rather than printing `NaN`:

```
[explorer] slow home search: query="readme" elapsed=2500ms — server timing unavailable (older server, or the response omitted it)
```

Aborted/superseded keystrokes never warn, so fast typing does not spray the console.

## Test Plan

- [x] `tests/test_index_search.py` — `test_rank_route_answers_a_timing_breakdown`: the rank response carries `timing` with all three keys.
- [x] `frontend/src/apps/explorer/FilesHome.render.test.tsx` — a fast response does not warn; a >=2s response warns with the breakdown; a >=2s response with `timing` absent warns without `NaN`; an aborted request never warns.
- [x] Scoped runs: `tests/test_index_search.py tests/test_index_api.py` → 203 passed. Frontend `FilesHome.render.test.tsx` + `home-search.test.ts` + `useListingSearch.render.test.ts` → 171 passed. `tsc --noEmit` clean.
- [ ] Full suite and CI in flight — draft until both are green.

**Manual verification:** open the Files home search with the browser console on, type against a slow mount, and confirm a single warn per slow search with the three server numbers plus the gap; confirm fast searches print nothing and rapid typing does not warn for superseded keystrokes.

**Scope note:** `rank` only. `stats` and `search` share the same lane and `_bounded_index_read` path and would take the same treatment, but are deliberately left out of a diagnostic-only change. The existing Python log line and its 750ms WARNING threshold are byte-for-byte unchanged.


iamsdas:
**CI status note — two lanes are red for failures that reproduce identically on `main`.**

Green here: `test-python` 3.11 / 3.12 / 3.13, `frontend`, `bundle-contents`, `linux-desktop`, `detect-changes`.

Red, and why each is not this diff:

| Lane | Failing test | Same failure on `main`? |
|---|---|---|
| `fused-engine` | `tests/test_env_install.py::test_the_mirrored_row_accepts_a_pathlib_project_dir` — `AssertionError: assert None == 'C:/Users/runner/app'` | Yes — identical assertion on `main` @ `b75a1b825` |
| `test-python-windows` | `tests/test_tasks_sent_mark.py::test_the_row_lands_in_the_project_the_send_named` — `'C:/Users/...' == 'C:\\Users\\...'` | Yes — identical assertion on `main` @ `71a45f94f` |

The first CI attempt at this sha also failed `tests/test_index_api.py::test_a_slow_query_does_not_block_the_interactive_lane` (`assert elapsed < 0.5`). That one was investigated and is a flake:

- The branch's only server change is inside `api_index_rank`, after `out` is computed — three `round()` calls plus a dict literal on the return value. It does not touch `api_index_stats`, `api_index_query`, or the lane semaphore, so it is not on the code path that test measures.
- The same test passed on all three `test-python` lanes in that very run.
- 60 local runs, 0 failures: 10x on branch, 10x on a fresh `origin/main` worktree, 5x under 16 CPU spinners.
- It passed on the CI re-run of the same sha.

PR #1221, opened 2026-09-18
Home search: detect changes on page focus so a new download is findable

iamsdas:
## Summary

- **Download a file, search for it on home, and it isn't there.** Home search is index-backed and global, but every existing trigger that refreshes the index structurally misses this case: the startup scheduler runs once per boot, `index_touch` only fires for mutations this app itself makes, and the folder-open freshness check only ever sees the folder being *listed* — home search lists no folder.
- Even pointing that check at the scan root wouldn't help. It compares a directory's `mtime_ns`, and a directory's mtime moves only when its **direct** entries change (`index/freshness.py`'s own "KNOWN BOUND, by design"). Scan roots default to `~`, so a file landing in `~/Downloads` moves `~/Downloads`'s mtime and not `~`'s — invisible.
- This PR adds a **focus trigger**: when the home page regains focus after 30s+ away, the server starts an ordinary incremental rescan of any scan root that has gone stale. No new scanning machinery — the same `runner.start` every other trigger uses, behind its own pacing and staleness floors.
- `~/Library/Caches` joins `default_ignore()` so the rescan doesn't index cache churn.

## Visuals

```mermaid
flowchart TD
    A[Home regains focus, 30s+ away] --> B{hidden_s floor}
    B -->|below| Z[No-op]
    B -->|above| C{Per-root floors}
    C -->|checked recently, or fresh| Z
    C -->|stale| D[Start incremental scan]
```

Four floors sit on that path, and they answer different questions — the reason they are four constants and not one:

| Constant | Question | Value |
|---|---|---|
| `MIN_HIDDEN_S` | Was this a real absence, or two of our own windows trading focus? | 30s |
| `DETECT_INTERVAL_S` | Has this root been *checked* recently? (in-memory request collapsing) | 30s |
| `FOCUS_STALE_S` | Is this root stale enough to be worth **this** trigger acting on? | 300s |
| `freshness.MIN_INTERVAL_S` | Is this root due for a rescan at all — for any trigger? | 60s |

`FOCUS_STALE_S` is the one code review added. Reusing the shared 60s floor meant a user who kept tabbing away for just over 30s and back could restart a full scan roughly once a minute, indefinitely. A full incremental walk of a real home root measures **4.37s over 78,717 directories**, so at 300s the worst-case duty cycle is ~1.5% rather than ~7%.

## An earlier design was removed — worth knowing before reviewing

This PR originally ran `index/fsevents.py:hint()` **standalone** on focus and scanned only when the journal reported a real change, so the quiet path cost nothing. That inverts on a real home directory, measured on one:

| Time since last scan | `hint()` result |
|---|---|
| immediately after a scan | 123 events, **48ms** |
| 7.2 hours after | **199,943 events, 11.9s** — sitting on the 200,000 cap |
| same minute, first call | **`None`** ("cannot tell") at 5.8s |

`~` generates roughly **28,000 FSEvents/hour**, so `hint()` stops being able to answer after about **7 hours** — i.e. precisely when new files are most likely to be missing. Every test mocked `hint()`, so nothing caught it. The standalone replay is gone; `detect.py` no longer imports `fsevents` at all.

## Usage

Not directly user-invocable — it fires on its own:

1. Open the home page, then switch away to another app for more than 30 seconds.
2. While away, download a file or drop one into any folder under a scan root.
3. Return to the home page and search for it.

The endpoint behind it:

```
POST /api/index/note-home-focus   {"hidden_s": 45}
```

Fire-and-forget — returns `{"ok": true}` immediately whether or not a check runs. `hidden_s` is untrusted and the server cannot verify it; the `MIN_HIDDEN_S` gate rejects only *smaller* claims, so what actually bounds a modified client is the pacing and staleness floors above, not that constant.

## Design notes

| Decision | Why |
|---|---|
| Home page only, not app-wide focus | Every other index-backed surface already has a trigger. Home search is the one with none. |
| `visibilitychange` **and** window blur/focus | `visibilitychange` alone never fires when the user switches to a different *application* while the window stays visible — it misses the motivating scenario on the packaged desktop app. |
| Away-detection climbs to `window.top` | Explorer renders in an iframe beside sibling shell panes, so a frame-local `document.hasFocus()` counts clicking the sidebar as "went away". Falls back to the frame-local check when unframed or cross-origin. |
| The correctness of the trigger does not depend on the ignore list | `IndexConfig.ignore` is user-editable and, once saved, frozen — see the caveat below. A trigger cannot afford to lose correctness to an edited preference. |

**Two costs, stated plainly.**

Adding `~/Library/Caches` changes `IgnoreRules.sig()`, and by that function's documented behaviour **the first scan after upgrade is a full rescan** for every user who has not saved a custom ignore list. One-time, and arguably right — existing indexes hold cache rows that now need purging — but real.

And the mirror image: `default_ignore()` is a `default_factory` consulted **only** when the persisted config has no `ignore` key, and `save_config` writes the list verbatim. **Anyone who has ever pressed Save in the Indexing panel carries a frozen snapshot and will never receive `~/Library/Caches`.** That mechanism is pre-existing and out of scope here, but it means this PR's ignore addition reaches new installs and untouched configs only.

## Test Plan

- [x] `tests/test_index_detect.py` — the per-root `DETECT_INTERVAL_S` floor, `FOCUS_STALE_S`, the `freshness.MIN_INTERVAL_S` scan floor, indexing-pref-off, hidden duration below the floor, an active-run guard, a joined scan not being reported as started, a mount-backed root, and the "never raises" contract
- [x] `tests/test_index_api.py` — the endpoint: `_require_fused` guard, fire-and-forget contract, malformed `hidden_s`, the floor applied before the slot acquire, slot drop-on-overlap, disabled no-op. Background threads are joined before each test returns, so the module-global `_detect_slot` cannot leak into another test.
- [x] `frontend/src/apps/explorer/lib/focus-detect.test.ts` — the pure gate and `isAway`, including focus moving to a sibling shell pane (`bun test`)
- [x] Full suite clean vs `origin/main` — zero branch-only failure IDs
- [ ] **`test-python-windows` is red on `main` too** (starved runner; the failing file churns between runs). No regression from this PR and no fix for that lane here.
- [ ] **Not verified: the end-to-end browser wiring.** Worth a manual pass before merge — hide the home page 30s+, drop a file under a scan root, return and search.
- [ ] **Not verified: the `window.top` assumption against the packaged shell's real frame tree.** The same assumption is already made in `TaskCards.tsx`, but it was not independently checked this round. If the shell nests panes under an intermediate frame, this degrades silently to the old frame-local behaviour rather than failing loudly.
- [ ] **Not verified on Windows by any local run.** The `ntpath.expanduser` portability fix was reasoned from CPython behaviour and confirmed only by CI dropping that failure.


iamsdas:
CI note on the merge commit `81edd59a7`: the first run showed `test-python (3.11)` red on a single test, `tests/test_index_api.py::test_a_wedged_rank_request_does_not_permanently_hold_its_lane_slot`. It passed on re-run.

It is a flake, not a regression from this branch:
- The same test ID already failed on main's own run `35350883296`, before this branch merged.
- It monkeypatches `ABANDON_S` to `0.05`, so it asserts against a 50 ms window on a shared runner.
- This branch's addition to `server/routers/index.py` takes its own `_detect_slot` and never touches the interactive-lane permits that test exercises.

Current state: 3.11 / 3.12 / 3.13, frontend, fused-engine, linux-desktop, bundle-contents all green. `test-python-windows` is red, as it is on main (3 of its last 3 runs); this branch's failures there are a strict subset of main's.

iamsdas:
Parked as a draft (2026-09-21). The focus trigger is superseded by a live filesystem watcher being built on branch `index-live-watch`; see `SPEC-index-live-watch.md` there. The two reusable pieces here (the `~/Library/Caches` default ignore and the root-mtime evidence helper) are re-applied or made unnecessary by that build. Not closed yet in case the watcher approach is rejected.

PR #1276, opened 2026-09-21
App Doctor: cross-browser check runs as a task, cached on task + checksum

vasu2001:
## Summary
- The cross-browser row's Check button now **creates an App Doctor check task** (same `_create_app_task` seam as fix tasks) instead of blocking the request on a one-shot `claude -p`. It shows on the app's Tasks tab, raises the shell's finished-task notice and the sidebar unread dot, and the row reads `Checking…` off the task store on every GET (`check_task`), so a tab switch or reload never loses the in-flight state.
- **Cache keyed on task + content checksum**: server writes `.fused/cache/app-doctor/cross-browser.json` (`{checksum, task_id, started_at, ...}`), the session writes `cross-browser.verdict.json`. Row = join of the two; a checksum mismatch resets to "app changed", a record with no verdict is "checking" while the task is live or "ended without a verdict" once it is gone.
- Fix and check tasks share the one-live-task-per-app 409 gate (`is_doctor_prompt`), told apart via `is_doctor_check_prompt` so the panel draws "Checking…" vs "Fix in progress".
- Panel polls the report every 4s while a check task is live so the verdict lands in place.
- Skill doc updated; the `--json-schema` one-shot spawn is gone, so the verdict shape is prompt-enforced and validated leniently on read.

## Test plan
- [ ] Open an app's Doctor tab, press Check on the cross-browser row: row says Checking…, task appears under Tasks tab, notice fires on finish, verdict lands on the row without reopening
- [ ] Switch tabs / reload mid-check: row still shows Checking…
- [ ] Edit a view file after a verdict: row returns to "app changed"; Check again creates a new task
- [ ] Press Fix while a check runs: 409 / disabled

---

> [!NOTE]
> **Medium Risk**
> Changes how on-demand checks run (task lifecycle, cache format, and mutual exclusion with fixes) and removes the old one-file cache shape—worth exercising reload mid-check, re-check, and fix-during-check paths.
> 
> **Overview**
> The cross-browser **Check** action no longer blocks the HTTP request on a one-shot Claude CLI call. It **creates an App Doctor check task** (same `_create_app_task` path as fixes), returns immediately, and surfaces the work on the **Tasks tab** with normal finish notifications.
> 
> **Server:** Verdict caching is split into a server-written **run record** (checksum + `task_id`) and a session-written **verdict file** under `.fused/cache/app-doctor/`. GET attaches live check tasks as `check_task` (separate from fix `task`) via `is_doctor_check_prompt` and `_settle_check_row`. Fix and check tasks share the **one-live-task-per-app** gate (409). The blocking spawn, threading lock, and `--json-schema` path in `app_doctor_ai` are removed in favor of `begin` / `record_task` and `check_prompt`.
> 
> **Client:** In-flight state comes from **`check.check_task`** on the report (not local-only state), with a short **creating** phase and **4s polling** until the verdict lands; the status dot polls the same way. **Fix all** and row actions respect check-in-progress; API types add `AppDoctorRunResult` and `announceTasksChanged` on new tasks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4a2135fb5233780845806d372020bb01c91916a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vasu2001:
bugbot run

